PR #9, opened 2015-09-17
Fix #8

### LINKED ISSUE #8: pin property accessible?
```python
led = LED(2)
led.on()  # it works
led.pin = 3
led.off()  # error
```

Users can do this. Should I care? @waveform80 

Alternatively, the `pin` property could be prefixed with an underscore.

waveform80:
Make the pin property a read-only property


bennuttall:
Excellent - thanks

Issue #7, opened 2015-09-15
API design


bennuttall:
*This comment will be updated as API design is agreed.*

*`.xx()` denotes method, `.xx` denotes property.*

**LED**:

- `is_active`
- `on()`
- `off()`
- `blink()`
- `toggle()`

**Buzzer**:

- `is_active`
- `on()`
- `off()`
- `toggle()`
- `flash()`

**Motor**:

- `on()`
- `off()`

**Robot**

- `forwards()`
- `backwards()`
- `left()`
- `right()`
- `stop()`

**Button**:

- `is_pressed` (is it currently pressed?)
- `when_pressed` (callback)
- `when_released` (callback)
- `wait_for_press()` (wait for edge)

**MotionSensor**:

- `motion_detected`

**LightSensor**:

- `value`

**TemperatureSensor**:

- `value`

waveform80:
Might be worth taking a leaf out of the Explorer HAT API for the button and adding `when_released`, `when_changed` (pressed or released) and similar for the wait functions?

waveform80:
Oh, and `blink` for the LED class? (again, an idea shamelessly stolen from Explorer HAT :). I can come up with a PR for these if you want to have a play with them before deciding whether they're worth including?

bennuttall:
Yes please! As long as features are:

- useful
- common
- uncomplicated
- don't make other use more complicated

then I'm happy.

waveform80:
Okay, as of #14 we've got the following, starting with the generic stuff (I've left out derived stuff and just indicated the parent class):

#### GPIODevice

* pin (RO) - The GPIO pin the device is associated with
* is_active - True if the GPIO is active, False if it's not. Descendants automatically compensate for pull-up/down and yes, this applies to outputs as well (it works, so why not?)

#### InputDevice(GPIODevice)

* pull_up (RO) - True if the input's set to pull-up, `False` if it's pull-down. Read-only
* wait_for_input() - basically an alias for GPIO.wait_for_edge. I'm still not entirely sure about having this in the base class: it doesn't play nice with some current descendants and personally I'd rather remove it and replace it with class specific stuff like `wait_for_press` on Button and similar things on other class'.
* add_callback() & remove_callback() - similar to above - I'd rather have class specific stuff like `when_pressed`, etc. I'm also a bit wary of directly using GPIO's callback mechanism. It's not very friendly if you attempt to add multiple callbacks to a particular device (do we want to support this?)

[...]

PR #12, opened 2015-09-17
Add LED.blink method - ref #7

waveform80:
Adds a background threaded `blink` property to the LED API as suggested in comments to #7. Uses `Event` for termination to ensure quick reaction to requests to stop / change blink cycle (I should change the `MotionSensor` class to use this method too...).

This is a rather messy PR, sorry! I got tired of all the repeated definitions of the read-only `pin` property so this breaks out that into a `GPIODevice` base class which `InputDevice` and `OutputDevice` then inherit from. If you'd rather stick with the repeated definitions let me know and I'll redo this.

### DIFF
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -4,6 +4,11 @@ from RPi import GPIO
 GPIO.setmode(GPIO.BCM)
 GPIO.setwarnings(False)
 
+
+from .devices import (
+    GPIODevice,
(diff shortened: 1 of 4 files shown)


waveform80:
Actually, forget this - there's an issue with using `__del__` to shut down the background threads. Something a little more nuanced is called for (both for this, and `MotionSensor`). I'll bung another PR together tomorrow.

PR #13, opened 2015-09-18
Shut down GPIO threads nicely

waveform80:
The motion sensor queue doesn't shut down properly at script end at the
moment and prevents the interpreter shutting down. This is because it's
a non-daemon thread so `__del__` never gets run and so on.

This is a bit of a major PR - I can split it up if you want. Firstly it
makes a common base class called `GPIODevice` for both `InputDevice` and
`OutputDevice`. This just takes care of the read-only pin stuff. Next it
makes a `GPIOThread` class that ensures its a daemon thread, and which
also ensures proper cleanup on shutdown.

Finally, it fixes `MotionSensor` to use the new `GPIOThread` class
(tested this time! Works nicely) and adds the `blink` method to the
`LED` class (which also works nicely this time).

### DIFF
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -4,14 +4,11 @@ import atexit
 
 from RPi import GPIO
 
-
-def gpiozero_shutdown():
-    GPIO.cleanup()
-
-atexit.register(gpiozero_shutdown)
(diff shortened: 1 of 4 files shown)


waveform80:
Oops, forgot to remove the redundant pin definition from InputDevice....

PR #35, opened 2015-09-27
Fix #27

### LINKED ISSUE #27: RuntimeWarning: A physical pull up resistor is fitted on this channel!
A `RuntimeWarning` is raised when setting up GPIO pins 2 or 3 because they have physical pull-up resistors on them.

1. Is it possible to suppress this warning when it won't make a difference? e.g. an LED
1. Is it possible to not allow setup on these pins when it won't work? e.g. when `pull_up=False` is provided on a Button setup

waveform80:
This patch throws InputDeviceError if anything attempts to initialize GPIO pins 2 or 3 as an input with pull-down (because that can't work), and otherwise suppresses the `RuntimeWarning` displayed when an attempt is made to initialize pins 2 or 3 to anything (basically the RPi.GPIO library is a bit over-enthusiastic in its default warnings).

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -1,6 +1,7 @@
 from __future__ import division
 
 import inspect
+import warnings
 from functools import wraps
 from time import sleep, time
 from threading import Event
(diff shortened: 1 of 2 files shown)


bennuttall:
Looks good!

PR #39, opened 2015-09-28
Explicit cleanup

### LINKED ISSUE #38: Can't recreate a Button object
Running the following code in a shell:

```python
>>> button = Button(15)
>>> button = Button(15)
```

The second line causes a `RuntimeError` raised by `GPIO.add_event_detect()`

Oddly, trying a couple more times, it eventually lets you.

I also managed to get a Segfault doing this once too many times! (Probably a `RPi.GPIO` problem, not `gpizero`)

waveform80:
This is a rather complex PR which partially (but not completely) attempts to fix #38. There are several issues here which include:

* Tracking object lifespan in a garbage collected, dynamic language is *hard*. Hard enough that I'd never expect a newbie programmer to understand it (in the case of GPIO stuff it involves figuring out when objects are being kept alive because a callback they implement is referenced by a chunk of external C code).
* Even if you can track the lifespan you can't force the garbage collector to do what you want. You may think the object can be killed off, but unless the garbage collector agrees that every reference is gone, it's not going to do what you want.
* The RPi.GPIO library isn't really designed for interactive REPL use; it makes several tacit assumptions that GPIOs won't be re-used for things during a single session (or at least not without the requisite cleanup functions being called in the proper order).

Anyway, the upshot is that the following is *never* going to work:

```python
>>> from gpiozero import Button
>>> btn = Button(14)
>>> btn = Button(14)
```

Consider the sequence of events:

1. The gpiozero library is imported and we grab the Button class from it
2. We construct a Button instance bound to GPIO pin 14
3. We assigned that instance to the name `btn`
4. We construct another Button instance bound to GPIO pin 14; this will fail because there's already another instance bound to that pin
5. The re-assignment of the `btn` name never happens because the construction of the new Button instance has already failed

You might be tempted to think the following would work:

```python
>>> from gpiozero import Button
>>> btn = Button(14)
>>> del btn
>>> btn = Button(14)
```

But it won't either because of the effects of callbacks referencing their owning objects alluded to earlier. The only solution to permitting such behaviour is to have an explicit method used to release all acquired resources. In deference to the Python idiom of using `close` for such a method, that's what this PR adds. This means you *can* do the following:

```python
>>> from gpiozero import Button
>>> btn = Button(14)
>>> btn.close()
>>> btn = Button(14)
```

It also adds the ability to use GPIODevice instances as context managers to do this implicitly in a one-shot manner:

```python
with Button(14) as btn:
    # do something with btn
```

### DIFF
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -1,11 +1,6 @@
 from __future__ import absolute_import
 
-import atexit
-
-from RPi import GPIO
-
 from .devices import (
-    _gpio_threads_shutdown,
     GPIODeviceError,
     GPIODevice,
 )
@@ -35,11 +30,3 @@ from .boards import (
     TrafficHat,
 )
 
-
-def gpiozero_shutdown():
-    _gpio_threads_shutdown()
-    GPIO.cleanup()
-
-atexit.register(gpiozero_shutdown)
-GPIO.setmode(GPIO.BCM)
-GPIO.setwarnings(False)
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -1,11 +1,36 @@
+import atexit
 import weakref
-from threading import Thread, Event
+from threading import Thread, Event, RLock
 from collections import deque
 
 from RPi import GPIO
 
 
+_GPIO_THREADS = set()
+_GPIO_PINS = set()
+# Due to interactions between RPi.GPIO cleanup and the GPIODevice.close()
+# method the same thread may attempt to acquire this lock, leading to deadlock
+# unless the lock is re-entrant
+_GPIO_PINS_LOCK = RLock()
+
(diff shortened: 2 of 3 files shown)


waveform80:
Okay, I'm reasonably happy with how this works now. Interestingly I think I've come across a bug in RPi.GPIO though... try creating a `LightSensor` with this, use it for a bit then call `close()` on it. Then try re-creating it and you'll get an error that it can't add edge detection. Try again a bit later and it suddenly works though. I've dug into RPi.GPIO's code and there's something going on in `add_edge_detect` in `event_gpio.c` but I haven't narrowed down exactly what yet.

waveform80:
Anyway, feel free to merge this in the meantime (I don't seem to have the necessary permission to do so despite being in the RPi-Distro org now ... lacking some auth on the group perhaps?)

PR #42, opened 2015-10-03
Docs clean up part 1

waveform80:
Big push on getting the docs cleaned up before 1.0. Proper wrapping of
everything so it's decently viewable from the command line (or as
decently viewable as markdown can be - the tables will never look great
from the command line).

Only one code change in this PR: rename `bouncetime` to `bounce_time`
(everything else is PEP-8, so this probably should be too) and change
its units to seconds from milliseconds (again, all other durations in
the library are in seconds, so it feels inconsistent that this one
isn't; for the sake of those who won't read the docs - which is most
people - I figure consistency helps with guessing!).


bennuttall:
Great, thanks. What's part 2 entail?

waveform80:
Part 2's about finishing off as many examples as I can cram in :)

PR #59, opened 2015-10-12
Fix #57

### LINKED ISSUE #57: Blink does not work for PWMOutputDevice
When you try to blink a `PWMOutputDevice`, it flickers on and off quickly once, then stays off apart from random short flashes, even after turning off!

It may be caused by `blink` definition being inherited from `DigitalOutputDevice`, and it calling `super`'s `on` and `off`, but I'm not sure.

bennuttall:
I believe this fixes #57 but perhaps there's a better way? See explanation in comments on #57.

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -62,19 +62,25 @@ class DigitalOutputDevice(OutputDevice):
         self._blink_thread = None
         self._lock = Lock()
 
+    def _on(self):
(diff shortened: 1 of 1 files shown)


bennuttall:
Shall I close this one then? @waveform80

waveform80:
Yeah, let's close this one - I'll rebase #61 in a mo

PR #60, opened 2015-10-12
Fix #50

### LINKED ISSUE #50: Heisenbug in add_event_detect
The one issue I've come across in practice with RPi.GPIO (which annoyingly happened during the demo too the other day!) is that sometimes (not often - it's a Heisenbug) the library fails to add edge detection when requested. It happens particularly with Button and LightSensor which both use `add_event_detect`. I haven't yet tied down if there's any pre-requisites to re-produce but it's something I need to look into as when it manifests the only option is to shut down the interpreter and try again.

So, first step: mitigation. Ensure all resources allocated in all constructors are cleaned up in the event the constructor fails (in particular cleaning up the allocation of the GPIO pin and pull-up). Second step: dig into the C source code and find out what on earth is causing `add_event_detect` to actually fail on rare occasions. I'm leaving this as the second step as it's going to be Hard (tm); probably some subtle race condition or something like that.

waveform80:
This isn't a full fix but I can't seem to reproduce the issue in #50 at
the moment. So for now this just ensures that exceptions in the
constructors get cleaned up properly (so they don't block future
construction attempts against the same pin)

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -63,18 +63,22 @@ class InputDevice(GPIODevice):
         self._inactive_state = GPIO.HIGH if pull_up else GPIO.LOW
         pull = GPIO.PUD_UP if pull_up else GPIO.PUD_DOWN
 
-        # NOTE: catch_warnings isn't thread-safe but hopefully no-one's messing
-        # around with GPIO init within background threads...
-        with warnings.catch_warnings(record=True) as w:
-            GPIO.setup(pin, GPIO.IN, pull)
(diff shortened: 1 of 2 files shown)


bennuttall:
Is this ready to merge?

waveform80:
Yup, should be - but it might break the merging of #61 - if it does, no worries - I'll just rebase the latter

waveform80:
Sorry - did you want me to merge it? No worries if so, I was just wondering if you wanted to sign off on stuff first?

PR #61, opened 2015-10-12
Fix #57

### LINKED ISSUE #57: Blink does not work for PWMOutputDevice

waveform80:
This refactors the GPIOOutputDevice to have an internal `_write` method
similar to the `_read` method used to grab the value of a GPIO device.
This is used for simple 1s and 0s until we get to PWMOutputDevice which
replaces the value with the PWM device's duty cycle level. As a result,
all the DigitalOutputDevice base methods (including blink) "just work".

This commit also, for the moment, removes min_pwm and max_pwm because
I'm not 100% certain how they should interact with the notion of
is_active at the moment. They'll probably get added back in at some
point but I need a bit more time to think about it!

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -39,17 +39,20 @@ class OutputDevice(GPIODevice):
             self.close()
             raise
 
+    def _write(self, value):
+        GPIO.output(self.pin, bool(value))
+
     def on(self):
         """
         Turns the device on.
         """
-        GPIO.output(self.pin, True)
+        self._write(1)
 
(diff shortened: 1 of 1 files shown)


bennuttall:
Can you resolve and merge this please @waveform80?

PR #62, opened 2015-10-12
Attempt to fix example syntax sections

waveform80:
Something fishy going on with the Markdown for the code examples in the
Notes page. Might be down to the indentation of the list item's content?
The vim markdown parser seems to think so anyway; let's see what
pythonhosted.org says ...

### DIFF
--- a/docs/notes.md
+++ b/docs/notes.md
@@ -2,88 +2,88 @@
 
 1. **BCM pin numbering**
 
-    This library uses Broadcom (BCM) pin numbering for the GPIO pins, as
-    opposed to BOARD. Unlike the `RPi.GPIO` library, this is not configurable.
+   This library uses Broadcom (BCM) pin numbering for the GPIO pins, as
+   opposed to BOARD. Unlike the `RPi.GPIO` library, this is not configurable.
 
-    Any pin marked `GPIO` can be used for generic components.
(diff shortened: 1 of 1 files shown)


bennuttall:
Hmm. It's still broken. The resource generator I use tends to be ok at generating code blocks nested within list items as long as everything's indented, ensuring the list numbering doesn't end.

However, mkdocs is breaking the code blocks here. I've tried with two, three and four space indentations and they're all broken.

I did manage to fix it by replacing fenced code blocks with just double indented code, like so:

```
1. **Keep your program alive with `signal.pause`**

    The following program looks like it should turn an LED on:

       from gpiozero import led

       led = LED(2)

       led.on()
```

waveform80:
Hmm, the GFM-based syntax highlighter I use in vim (https://github.com/jtratner/vim-flavored-markdown) thinks that the 4-char indented content of the lists is all a huge code block (because it's 4-chars indented) and corrects this when I dropped it to 3, which is why I thought it might work on pythonhosted. Unfortunately this is one of several areas where MD isn't well specified so implementations tend to vary substantially in their interpretations.

I'm vaguely tempted to convert that chapter to reST so I can at least link to things like signal.pause properly and highlight lines in code blocks, but I know you'd hate me for it ;) I'll tweak it to use indented blocks for now!

PR #77, opened 2015-10-17
Fix #74

### LINKED ISSUE #74: led.is_on not led.is_active
Someone suggested on the forums that the `LED` class's `is_active` property should be named `is_on`. I think it's a good idea and aligns with other aliases like `is_pressed` in `Button`.

@waveform80 can we move `_alias` from `input_devices.py` to `devices.py` and use it in both input and output devices?

waveform80:
Add more property aliases and do them properly (no more lambdas and
string lookups) which means we can remove `_alias`. This commit also
defines `__slots__` for all classes which should prevent assignation of
invalid attributes with an AttributeError (more friendly than silently
doing the wrong thing). Finally, it cleans up all the property defs to
use Ben's preferred decorator style.

### DIFF
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -47,7 +47,11 @@ class GPIODevice(object):
         The GPIO pin (in BCM numbering) that the device is connected to. If
         this is `None` a `GPIODeviceError` will be raised.
     """
+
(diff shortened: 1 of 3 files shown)


bennuttall:
Awesome - thanks

PR #88, opened 2015-10-24
Fix incorrect variable name

tjguk:


tjguk:
Sorry; forgot to put a description. Using the Robot object. There was an incorrect var name ("back" instead of "backwards") in the Motor.__init__

waveform80:
Doh - that's my fault! Thanks for catching it

PR #95, opened 2015-10-26
Fix #90

### LINKED ISSUE #90: Optional PWM interface to boards?
I can create an instance of a `TrafficHat`, but then I have access only to `LED` objects, meaning I can't dim or pulse the LEDs.

How about adding a PWM option on init?

```python
th = TrafficHat(pwm=True)
```

Then the LEDs would be created as `PWMLED` objects.

waveform80:
This is a combination of Ben's original commit to fix #90 with a couple of tweaks to ensure Python 2 compatibility and get the super-call in the value setter to work properly. Tested with both Py 2 and 3, seems to work nicely.

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -21,8 +21,9 @@ class LEDBoard(SourceMixin, CompositeDevice):
     """
     A Generic LED Board or collection of LEDs.
     """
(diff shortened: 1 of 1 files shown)


bennuttall:
Neat

PR #102, opened 2015-11-04
Fix #100

### LINKED ISSUE #100: Music Box example doesn't work
When I first saw the http://pythonhosted.org/gpiozero/recipes/#gpio-music-box example, I *suspected* that something didn't look right - AFAICT the locally-defined play function would never get called.

So I finally got round to breaking out my breadboard and installing gpiozero, and giving it a whirl. Upon running the code, it actually prints out an error!
`sudo python musicbox0.py` prints:
```
  File "musicbox0.py", line 21, in <module>
    button.when_pressed = sound.play
  File "/usr/local/lib/python2.7/dist-packages/gpiozero/devices.py", line 75, in __setattr__
    return super(GPIOBase, self).__setattr__(name, value)
  File "/usr/local/lib/python2.7/dist-packages/gpiozero/input_devices.py", line 163, in when_activated
    self._when_activated = self._wrap_callback(value)
  File "/usr/local/lib/python2.7/dist-packages/gpiozero/input_devices.py", line 210, in _wrap_callback
    'value must be a callable which accepts up to one '
gpiozero.input_devices.InputDeviceError: value must be a callable which accepts up to one mandatory parameter
```

and `sudo python3 musicbox0.py` prints:
```
  File "/usr/local/lib/python3.2/dist-packages/gpiozero/input_devices.py", line 197, in _wrap_callback
    inspect.getcallargs(fn)
  File "/usr/lib/python3.2/inspect.py", line 940, in getcallargs
    spec = getfullargspec(func)
  File "/usr/lib/python3.2/inspect.py", line 834, in getfullargspec
    raise TypeError('{!r} is not a Python function'.format(func))
TypeError: <built-in method play of Sound object at 0x76959308> is not a Python function

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.2/dist-packages/gpiozero/input_devices.py", line 203, in _wrap_callback
    inspect.getcallargs(fn, self)
  File "/usr/lib/python3.2/inspect.py", line 940, in getcallargs
    spec = getfullargspec(func)
  File "/usr/lib/python3.2/inspect.py", line 834, in getfullargspec
    raise TypeError('{!r} is not a Python function'.format(func))
TypeError: <built-in method play of Sound object at 0x76959308> is not a Python function

During handling of the above exception, another exception occurred:

  File "musicbox0.py", line 21, in <module>
    button.when_pressed = sound.play
  File "/usr/local/lib/python3.2/dist-packages/gpiozero/devices.py", line 75, in __setattr__
    return super(GPIOBase, self).__setattr__(name, value)
  File "/usr/local/lib/python3.2/dist-packages/gpiozero/input_devices.py", line 163, in when_activated
    self._when_activated = self._wrap_callback(value)
  File "/usr/local/lib/python3.2/dist-packages/gpiozero/input_devices.py", line 210, in _wrap_callback
    'value must be a callable which accepts up to one '
gpiozero.input_devices.InputDeviceError: value must be a callable which accepts up to one mandatory parameter
```

(I'm testing on Wheezy, hence the requirement for `sudo`).
Perhaps to a beginner, it might not be clear that the "up to one mandatory parameter" error message also includes the option of zero parameters?

It looks like the error is coming from https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/input_devices.py#L187 - observe this output from an interactive python prompt:
```python
>>> import pygame.mixer
>>> pygame.mixer.init()
>>> drum = pygame.mixer.Sound("samples/drum_tom_mid_hard.wav")
>>> type(drum.play)
<type 'builtin_function_or_method'>
>>> callable(drum.play)
True
>>> import inspect
>>> inspect.getcallargs(drum.play)
  File "/usr/lib/python2.7/inspect.py", line 900, in getcallargs
    args, varargs, varkw, defaults = getargspec(func)
  File "/usr/lib/python2.7/inspect.py", line 815, in getargspec
    raise TypeError('{!r} is not a Python function'.format(func))
TypeError: <built-in method play of Sound object at 0x76d5e120> is not a Python function
```

I **was** able to get the GPIO Music Box example working by modifying it to:
```python
from gpiozero import Button
import pygame.mixer
from pygame.mixer import Sound
from signal import pause

pygame.mixer.init()

def play(button):
    pin = button.pin
    sound = sound_pins[pin]
    print("playing note from pin %s" % pin)
    sound.play()

sound_pins = {
    2: Sound("samples/drum_tom_mid_hard.wav"),
    3: Sound("samples/drum_cymbal_open.wav"),
}

buttons = [Button(pin) for pin in sound_pins]
for button in buttons:
    button.when_pressed = play

print("ready")
pause()
```

This is only the first time I used gpiozero, but it seems counter-intuitive that the `play` callback is actually slightly more complex than the [RPi.GPIO equivalent](https://www.raspberrypi.org/learning/gpio-music-box/worksheet/), due to having to unwrap the pin number from the button object? But perhaps this allows additional flexibility.

waveform80:
Permit built-in (C implemented) methods and functions to be used as
event callbacks. Given that prototype introspection is not available for
such routines, assume that they take no (mandatory) parameters.

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -147,10 +147,10 @@ class WaitableInputDevice(InputDevice):
         active.
 
         This can be set to a function which accepts no (mandatory) parameters,
-        or a function which accepts a single mandatory parameter (with as many
-        optional parameters as you like). If the function accepts a single
-        mandatory parameter, the device that activates will be passed as that
-        parameter.
+        or a Python function which accepts a single mandatory parameter (with
+        as many optional parameters as you like). If the function accepts a
+        single mandatory parameter, the device that activates will be passed as
+        that parameter.
 
         Set this property to `None` (the default) to disable the event.
 
@@ -169,10 +169,10 @@ class WaitableInputDevice(InputDevice):
         inactive.
 
         This can be set to a function which accepts no (mandatory) parameters,
-        or a function which accepts a single mandatory parameter (which as
-        many optional parameters as you like). If the function accepts a single
-        mandatory parameter, the device the deactives will be passed as that
-        parameter.
+        or a Python function which accepts a single mandatory parameter (which
+        as many optional parameters as you like). If the function accepts a
+        single mandatory parameter, the device the deactives will be passed as
+        that parameter.
 
         Set this property to `None` (the default) to disable the event.
 
@@ -189,6 +189,13 @@ class WaitableInputDevice(InputDevice):
             return None
         elif not callable(fn):
             raise InputDeviceError('value must be None or a callable')
+        elif inspect.isbuiltin(fn):
+            # We can't introspect the prototype of builtins. In this case we
+            # assume that the builtin has no (mandatory) parameters; this is
+            # the most reasonable assumption on the basis that pre-existing
+            # builtins have no knowledge of gpiozero, and the sole parameter
+            # we would pass is a gpiozero object
+            return fn
         else:
             # Try binding ourselves to the argspec of the provided callable.
             # If this works, assume the function is capable of accepting no


bennuttall:
Why are built-in functions not seen to be callable?

waveform80:
@lurch D'oh - absolutely correct - I'll fix that in a mo

waveform80:
Built-ins do register as callable, but you can't inspect their parameter
prototype which is why the existing code failed. The new code assumes that
built in callables have no mandatory parameters

lurch:
Just out of curiosity - is that a "known issue" and something that might potentially get fixed in a future Python3 version?

waveform80:
Not going to happen as far as I can see. In order to introspect the prototype of the function, the inspect module needs to be able to query the metadata of the function. For Python functions, that's easy enough - it just looks up co_argcount et al in the function's object. But these things don't exist for builtin/external C functions (in either Python 2 or 3).

When builtin/external modules register themselves with Python they just pass in a load of [PyModuleDef](https://docs.python.org/3/c-api/structures.html#c.PyMethodDef) structures (one for each module-level function). Note that there's nowhere in that structure for specifying the names (or even number) of parameters of the function call. Instead, the external implementation is called with a tuple and optional dict of args and is expected to unpack them in its implementation (presumably using `PyArg_ParseTuple` or `PyArg_ParseTupleAndKeywords` although there's no absolute requirement for it to do so).

So, the argspec of the function is effectively within the implementation of the builtin/external function. There's no way to get at it without disassembling the function's implementation (and even if that were feasible there's no guarantee you'd even find a simple call to `PyArg_ParseTuple` or `PyArg_ParseTupleAndKeywords`).

Basically, it's by design (presumably for speed and efficiency which is largely the point of builtin/external functions) so I doubt there's an impetus to fix it.

As it is, I don't think there's any great problem with the kludge I've used here (the assumption that any external function used as a callback takes no mandatory parameters). The only parameter we would pass is a gpiozero object; given gpiozero is relatively new we can reasonably assume no builtin/external implementations currently use or rely upon being passed a gpiozero object (in any circumstance). We can also reasonably assume there are plenty of builtin/external functions which take no (mandatory) parameters (like `pygame.mixer.Sound.play`) which would be of use as callbacks.

This does impose the burden that any external implementations wishing to work with gpiozero in future need to ensure their callbacks take no mandatory parameters, but given it's trivial to wrap an external implementation in a Python function in order to pass a parameter I don't think that's an excessive cost.

PR #120, opened 2015-11-19
Update recipes

lurch:
Remove superfluous function from Music Box example, and add pause() to all the recipes that need it


lurch:
I haven't tested all the recipes, but I *think* I've added `pause()` in all the right places.

lurch:
Any chance you could do whatever magic needs to be done to get http://pythonhosted.org/gpiozero/recipes/ updated?

PR #122, opened 2015-11-24
Added LED bargraph

martinohanlon:
I have added an LED bargraph to the output devices. 

You can deal with it as a single component rather than as separate led objects.

### DIFF
--- a/docs/outputs.md
+++ b/docs/outputs.md
@@ -307,3 +307,52 @@ Motor(forward=None, backward=None)
 | `value`     | The current speed and direction of the motor. `-1.0` if full speed backward, `0.0` if still, `1.0` if full speed forward. | Float               |
 | `values`    | A generator continuously yielding the motor's current value.                                                              | Generator           |
 | `source`    | A generator which can be used to continuously set the motor's value.                                                      | `None` or Generator |
+
+## LEDBargraph
+
+A strip of seperate LEDs in a single component.
+
+### Wiring
+
+The LED bargraph component has 2 rows of pins, a row of cathodes and a row of
+anodes, one pair for each LED in the bargraph.
+
+Connect the cathode of each LED to a ground pin, and connect the
+anode (the longer leg) to any GPIO pin, with a current limiting resistor in
+between:
+
+### Code
+
+Ensure the `LEDBargraph` class is imported at the top of the file:
+
+```python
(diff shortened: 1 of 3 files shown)


waveform80:
Lovely idea - couple of questions:

1. Would LEDMeter be a better name? Simpler, but I'm not sure it's as descriptive.
2. Rather than `on_up_to` and `on_down_to`, I'd much rather the `value` property were adjusted for this class so that value 0 means all LEDs off, value 1 is all LEDs on, and values in between light up a linearly proportional number of LEDs along the array (and if down-to support is really needed, which I'm not entirely convinced about given the array could just be reversed, values below zero could light LEDs in the opposite direction)
3. Once the point above is addressed, `SourceMixin` could be added to the front of the parent class specification so that you could do things like `meter.source = light_sensor.values` and instantly you've got a digital light meter.

martinohanlon:
I went with LEDBargraph as that is that general name of the component.

I like the idea of having a value property, but why not have both? A value property and descriptive methods which implement the action. I've got a requirement for on_down_to which is interestingly to display negative numbers, so the suggestion for negative values works quite well.

martinohanlon:
I'll submit a new PR with .value property attached to LEDBargraph.

waveform80:
Actually, no need for a new PR - you can just force push to your existing fork and it'll update this PR. On the subject of `on_up_to` and `on_down_to` - okay, you've convinced me there's a use-case for down-to (a bar graph that can display positive and negative values). I'm still not entirely convinced of the need for separate methods though. My comparison test in this case is `PWMLED` (or more generically `PWMOutputDevice` above it). This has the traditional `on` and `off` methods which just set it to 1 and 0 respectively, and a `value` property for more refined control. I suppose I could be convinced we ought to extend the `on` method to accept an optional parameter which specified the value to switch on to (which defaulted to 1) but I'm not sure I see the need immediately.

For that matter I'd also remove the `*args` parameter from the current PR's `on` and `off` methods. The class is for a bar graph and these args permit arbitrary combinations of LEDs within the array to be switched on/off. If I switch on every other LED then read the `value` property, what should it return? If users need to switch on arbitrary combinations of LEDs, they've already got the `LEDBoard` class for that; the use case for this class is rather more narrow.

martinohanlon:
I did not know that, I'll make some changes..

PR #126, opened 2015-11-29
Created LEDBargraph class

martinohanlon:


martinohanlon:
@bennuttall Any feedback on this PR?

lurch:
Just some comments from somebody who hasn't used GPIOZero very heavily yet...

I wonder if `on_up_to` and `on_down_to` should error-check for out-of-range values?
Perhaps `_on_up_to` and `_on_down_to` could just be replaced with a single function with `start` and `end` arguments?

I wonder if it'd be worth having a `on_only` method which could turn on just a single LED, and turn off all others?

Perhaps the `__init__` could have a parameter allowing you to specify whether the LEDs are configured in a common-cathode or common-anode arrangement?

I guess a value of `-1.0` would look visually identical to `1.0` (with all the LEDs being on in both cases) ? Are there use-cases in which you'd expect to send both negative and positive values to the same Bargraph object? (or would it make more sense if the bargraph only supported +ve values?)

waveform80:
I'll have a go at rebasing this for 1.1 - there's a few changes I'd like to make anyway (like getting rid of on_up_to and off_up_to; they're effectively redundant with .value and it'd be nice to have a mechanism of supporting -ve values in a consistent manner, if I can think of one). The common anode/cathode stuff should be handleable fairly easily once I've got the PWM-active-high stuff merged in a mo.

lurch:
Yeah I can see that only supporting a `value` set between 0.0 and 1.0 fits in with the rest of GPIOZero, but if you've got an 8-segment Bargraph, and you want to illuminate the first 3 LEDs, it seems a little user-unfriendly to require the value to be set to 0.375 ? \*shrug\*
I guess it depends how you envision the Bargraph object being used.

waveform80:
If you've got an 8-segment and you want the first three lit, just set value to 3/8. Yeah, okay, that's 0.375 if you calculate it, but I'd never write 0.375 in the code (and I probably wouldn't even bother calculating it); I'd write 3/8 because I want 3 of my 8 LEDs lit. Maybe that's worth including in an example in the docs?

lurch:
>  I'd write 3/8 because I want 3 of my 8 LEDs lit. Maybe that's worth including in an example in the docs?

Agreed.

martinohanlon:
I have a use case for displaying negative values on an led bargraph which is why this function is really useful for me. Hence the PR including it.

Removing it makes this class redundant to me, if you feel its the right way to go thats fine, but why throwaway functionality for a desire to be consistent?

lurch:
I've just noticed that the Motor class supports values between -1 and 1, so there is a precedent for supporting -ve values...

lurch:
I was wondering if this should inherit from LEDBoard, but then I found the second paragraph at https://github.com/RPi-Distro/python-gpiozero/pull/122#issuecomment-160323533 which explains why not :)

lurch:
...and in terms of supporting -ve values, I guess you could do something like the following (untested, as I don't have an MCP3008, so can't be sure I've got my generator-handling code correct) to scale the 0 -> 1 values from an ADC into the range -1 -> 1 :

```python
from gpiozero import MCP3008, LEDBargraph
from signal import pause

def scale_source(gen):
    for value in iter(gen):
        yield (value * 2) - 1

pot = MCP3008(channel=0)
bar = LEDBargraph((24, 25, 10, 9, 11, 8, 7, 5, 6, 12))
bar.source = scale_source(pot.values)

pause()
```

And if that works, you could also do something similar to connect a potentiometer to a motor, so that if the pot is fully-left the motor would be going full-speed backwards, if the pot is in the middle the motor would be stopped, and if the pot is fully-right the motor would be going full-speed forwards? Might be something nice to add to the Recipes page :-)

waveform80:
Sorry for the silence the couple of days - off ill! @martinohanlon on negative values it was my intention to do as @lurch suggests and use the -1 to +1 model from Motor (as you say, there's use cases for it, and it doesn't detract from any other functionality so it's a good addition).

@lurch on your code example above: that was exactly the sort of thing I had in mind (including the motor example!), and yes I was intending to make LEDBargraph's constructor similar to LEDBoard's (I'm also pondering renaming it to LEDGraph just for the sake of brevity).

lurch:
> (I'm also pondering renaming it to LEDGraph just for the sake of brevity).

IMHO it should stay as LEDBargraph - people might see 'LEDGraph' and automatically think of an X-Y graph?

Funny that we both thought of exactly the same motor example - all the more reason to add it as a Recipe :-)  (when you get time!)

pcopa:
Just to chime in on this. I have a pH and ORP probe to control my pool. Both can have negative values.  I'm intending to use the differential inputs for this, but I also want to use single-ended for temperature.  Anyway the +1, -1 would work fine for me.

waveform80:
Okay, I'm going to cherry-pick this into a branch on my repo and get it merge-ready for 1.1. Some notes on changes I need to make (for my own reference):

* Rename to LEDBarGraph (as @lurch points out LEDGraph is potentially ambiguous, but PEP8 and all that - bar graph is two words)
* Convert docs to RST
* Implement `value` as -1 to +1, remove `on_up_to`, `on_down_to`
* Convert `led` method to `leds` property (same as LEDBoard for consistency)
* Add recipe demonstrating simple technique of turning on specific number of LEDs (3/8, -3/8, etc.)
* PWM? If there's time (`value` permits partially lit LEDs in this case)
* `active_high`? Hmm ... LEDBoard doesn't have this ... I wonder if it should? Leave it for now; it can wait for 1.2.

lurch:
> Some notes on changes I need to make

* Change the constructor to better-match LEDBoard

> * PWM? If there's time (value permits partially lit LEDs in this case)

Yeah, I'd considered that option too but thought it might be too much effort (and might not be noticeable?). But given that you've already added RGBLED colour-fading... ;-)

> `active_high`? Hmm ... LEDBoard doesn't have this ... I wonder if it should?

Oooh, I never noticed that! IMHO it definitely should (I'll open up a ticket for that), as when I have a bit more time I was planning to implement a SevenSegmentLED class, and those definitely come in both common-anode and common-cathode varieties.

waveform80:
Yeah, there's so much overlap between LEDBoard and LEDBarGraph I've wound up factoring out the common stuff into an abstract base. Just gonna run a few tests then I'll push the new branch in a mo. Don't think I'll have time for PWM right now, but that's fine - it can go in for 1.2.

bennuttall:
I was going to ask - what's the difference between LEDBoard and LEDBarGraph? The PiLiter should have the features of BarGraph, for instance. Why wouldn't another LED board have these features? Why can't we just add the BarGraph methods to LEDBoard?

waveform80:
LEDBoard represents a loose collection of LEDs any of which can be on or off at any given time. As such its value is a straight tuple of LED states and it enforces nothing about those states. LEDBarGraph is a refinement of this in that only certain LED states are valid (i.e. contiguous from start to finish for +ve and contiguous from finish to start for -ve) hence its value is a scalar ranging from -1 to +1 which is sufficient to describe the states of all LEDs within the array (this is also why you can't have both sets of functionality within a single class; but there's enough overlap that a common base makes sense).

bennuttall:
Ok sure. So PiLiter should extend LEDBarGraph, not LEDBoard?

http://cdn2.bigcommerce.com/server3700/8d73a/products/369/images/1498/B045_Pi_Liter_New__20992.1424968431.1280.1280.jpg?c=2

waveform80:
Only if you want to always treat it as a graph and not as a loose collection of LEDs. Personally I'd be tempted to have PiLiter extend LEDBoard and add a PiLiterBarGraph which extends LEDBarGraph (both as trivial classes which simply fix the *pins in the constructor).

bennuttall:
Oh I see. Having a bar graph stops you controlling individual LEDs, right? Makes sense

waveform80:
Yup. I can add PiLiterBarGraph while I'm at this if you want; it's trivial enough

waveform80:
Implemented with #176

PR #127, opened 2015-12-01
Fix upgrade instructions

lurch:
'apt-get upgrade' always upgrades all packages, to upgrade a single package we need to use 'apt-get install'


lurch:
There's many references about this online, e.g. http://askubuntu.com/questions/44122/how-to-upgrade-a-single-package-using-apt-get

bennuttall:
Thanks!

PR #131, opened 2015-12-03
Add SnowPi - wip, close #130

### LINKED ISSUE #130: Add SnowPi
I'm currently working on adding in Ryan's [SnowPi](https://ryanteck.uk/raspberry-pi/114-snowpi-the-gpio-snowman-for-raspberry-pi-0635648608303.html) add-on board.

SnowPi is simply a collection of 9 LEDs, each of which can be named according to its position (nose, left, eye, right, eye, etc.)

A simple implementation could be:

```python
class SnowPi(SourceMixin, CompositeDevice):
    """
    Ryanteck SnowPi add-on board.
    """
    def __init__(self, pwm=False):
        self.nose = LED(25)

        self.left_eye = LED(23)
        self.right_eye = LED(24)

        self.left_arm_top = LED(17)
        self.left_arm_middle = LED(18)
        self.left_arm_bottom = LED(22)

        self.right_arm_top = LED(7)
        self.right_arm_middle = LED(8)
        self.right_arm_bottom = LED(9)
```

or even:

```python
class SnowPi(LEDBoard):
    """
    Ryanteck SnowPi add-on board.
    """
    def __init__(self, pwm=False):
        super(SnowPi, self).__init__(7, 8, 9, 17, 18, 22, 23, 24, 25, pwm=pwm)
```

However, I wonder if it would be possible to create groups of LEDs so that they could be accessed like so:

```python
snow_pi.eyes # both eyes
snow_pi.eyes.left # left eye
snow_pi.arms # both arms
snow_pi.arms.right # right arm
snow_pi.arms.right.top # top LED of right arm
```

Where these are akin to LEDBoard or TrafficLights, i.e. `snow_pi.arms.on()` turns all LEDs in both arms on, `snow_pi.arms.left.on()` turns the left arm on, and `snow_pi.arms.left.middle.on()` turns just one LED on.

Would that be possible? Or is it overkill?

bennuttall:
Untested WIP implementation of the Ryanteck SnowPi interface

May also require `close()` method, `closed` property, `source` and `values`?


bennuttall:
@waveform80 I've added a complete interface allowing control of the SnowPi body parts. Works nicely!

Just missing source/values and throws errors upon closing the shell. I'm not sure how source/values would be used for this accessory.

@ryanteck Check this out! See #130 and try out the code in this branch.

lurch:
Given the overlap between methods in LEDBoard and the various SnowPi classes (leds/lights, on, off, toggle, blink), I wonder if that suggests there could/should be e.g. a CompositeLEDDevice Mixin that would automatically provide these methods?
(but I'm not very familiar with mixins, so maybe that's not a sensible suggestion, and a re-arrangement of the inheritance hierarchy would be better?)

waveform80:
`source` and `values` will be implemented implicitly; SourceMixin supplies the first (which you've got), ValuesMixin (which CompositeDevice inherits from) supplies the second. However, both expect the implementing class to provide `value` property for them to use - which I think is missing at the moment (couldn't see it in any of the commits there). I can add that easily enough though.

waveform80:
Hmm ... I like the idea of a CompositeLEDDevice class. The only question is how to handle the tuple class that'll be required for the `value` property; oddly enough I've been working on something similar to this for another of Ben's suggestions (linking gpiozero to other libraries like picamera via source/values). I'll see if I can find some time to pull some code from that (won't be this week - I've just seen the list of gpiozero tickets in my watchlist - wow!)

lurch:
Why would SnowPi need to use the ValuesMixin - isn't it effectively an output-only device? (and thus only needs SourceMixin?)
And on that note, why does CompositeDevice inherit from ValuesMixin?

IMHO the `lights` methods that @bennuttall 's used here should also be renamed to `leds`.

> The only question is how to handle the tuple class that'll be required for the value property;

Do we necessarily need a `value` property for a CompositeLEDDevice? (showing my gpiozero-newbness here!)  Especially for something like SnowPi where we have CompositeLEDDevices nested inside other CompositeLEDDevices - a single 'value' seems kinda meaningless?

Apologies for going OT on an unrelated topic, but looking at e.g. https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/boards.py#L244 which has both input devices (a button) and output devices (LEDs and a buzzer) what's the "logic" (or "rule") that dictates that getting the `value` returns a 4-tuple of the LED and buzzer states, and not the button's state?

> I've just seen the list of gpiozero tickets in my watchlist

Sorry about that... ;-)

waveform80:
All devices provide `values`, but only output devices (including some composite devices) provide `source`. This is because it's perfectly valid to read the state of an output device ("is the LED lit or not") and allows "device chaining", e.g.:

```python
from gpiozero import LED, Button
from signal import pause

btn = Button(2)
red = LED(17)
blue = LED(18)

blue.source = red.values
red.source = btn.values
pause()
```

Now, you may (rightly) argue "but you could achieve the same simply by setting both sources to `btn.values`". However, chaining becomes useful once you start introducing transforms, e.g. a delay:

```python
from gpiozero import Button, LED
from itertools import tee
from time import sleep
from signal import pause

def slow(it, delay=0.1)
    for item in it:
        yield item
        sleep(delay)

def delayed(it, n=10):
    queue = []
    for item in it:
        queue.append(item)
        while len(queue) > n:
            yield queue.pop(0)

def pairwise(it):
    # From itertools docs
    a, b = tee(it)
    next(b, None)
    return zip(a, b)

btn = Button(2)
leds = [LED(pin) for pin in (17, 18, 27, 22, 23)]
for led, source_led in pairwise(leds):
    led.source = delayed(slow(source_led.values))
leds[-1].source = btn.values
pause()
```

Now, once again you could achieve the same just with `btn.values`, but in that case you'd need to wrap it repeatedly or vary its wrapper parameters; this way seems cleaner. Which is why basically everything in gpiozero has `ValuesMixin` somewhere in its ancestry.

As for why `CompositeLEDDevice` requires `value`; because everything descends from `ValuesMixin`, everything must provide `value` for the `values` property to work. All composite devices provide a `value` which is a tuple of the values of the devices they compose. Why a tuple? Firstly this allows decomposition (if you only care about a subset of the values, i.e. the state of a single device, then just pick it out; we always provide namedtuples wherever possible to make this more friendly). Secondly because it allows users to compose inputs to composite output devices just with Python's `zip` built-in (especially in Python 3; in Python 2 you'd usually need `izip`). Just zip together the sources (buttons, pots, sensors, transformation functions, constants, whatever) and you've got your composite source.

Originally I considered using individual `value` properties in composite devices for each subordinate device but this would multiply all the `source` and `values` properties (which in turn necessarily multiplies the number of threads in the system, not to mention the API docs!). This implementation is simpler, and more efficient. It also means we can make composite devices of composite devices (which can have tuples of tuples for `value`).

So, the general rule for `value` is: Individual devices have a scalar `value` which takes a number between 0 (off) and 1 (on) and in certain special cases -1 (backwards/negative something) to +1 (forwards/positive something). Composite devices have a tuple value comprising the values of each subordinate device (whether that device is composite or individual), using a namedtuple class wherever possible to ease access. Why the limit on 0 to 1? Firstly because False and True cast to 0 and 1 in Python, which makes the transition between digital and analog-ish devices (like PWM) smooth. Secondly because any other scale implies an arbitrary number of steps is supported by the device (if we used 0 to 100 does a PWMLED really support 100 steps? Same argument for 0-255 and so on). With 0 to 1 we don't imply anything (and we support lots more precision than 0-100 and 0-255 for those devices which actually support such precision without having to vary the scale across devices).

I should probably add some of this to a "design" section at the back of the docs (for any devs that wander through), but to be honest they're all just things that popped into my head while mulling the source/values API. And do keep asking these questions! It definitely keeps fresh in my head why I did certain things (and occasionally prompts me to write them down in a form which is a bit more permanent than my deteriorating neurons!).

P.S. Apologies if any of the above code doesn't work; it's untested and off the top of my head :)

lurch:
Right, that all makes sense thanks (I'll read through it all carefully in more detail later) but I guess I was confused by the fact that https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/boards.py#L284 only allows you to 'get' the values of the output devices, and not the button as well (although perhaps this is related to #164 too?)

For CompositeDevices that have both input and output devices, will the values 'set tuple' always be the same size as the values 'get tuple'? What's the expected behaviour of trying to 'set' the value of a button? Does it simply get ignored?
Or will the 'get tuple' always include all devices (both input and output), and the 'set tuple' only include output devices?

> And do keep asking these questions!

Cool... :-)

I guess if we had a ButtonBoard class, then you could theoretically do something like:
```python
from gpiozero import LEDBoard, ButtonBoard
from signal import pause

buttons = ButtonBoard(5,6,7,8)
leds = LEDBoard(11,12,13,14)
leds.source = buttons.values

pause()
```
and it'd work, as long as the number of buttons matched the number of LEDs? Would there be errors if there was a mismatch in the number of each, or would any 'extra' components on either side simply be 'unused' ?

waveform80:
Very good point on the composite device not including the button. Oddly enough I was pondering this earlier (without realizing we'd already committed such an error). In my opinion a composite device composed of both input and output devices should return the state of *all* subordinate devices in its `value` property (in other words, the implementation of `TrafficLightsBuzzer.value` is currently wrong for excluding the button value from its result tuple).

However, on the question of what to do when *setting* the `value` of a composite device composed of both input and output devices, I'm not so sure (haven't thought about it enough yet). It seems to me there's three options:

* Have a different tuple type that doesn't include the input devices. Personally I find this option distasteful as it effectively means that the type of the `value` property is different between the "get" and "set" operations which rather breaks the calculated property contract. It's a bit like a property which returns an integer but only accepts a string.

* Accept the same tuple type, but ignore the values for any input device elements. This is the way I'm currently leaning simply because it makes chaining composite devices nice and easy. It's a little concerning to ignore values, but I can live with it.

* Accept the same tuple type, but if any value other than `None` is specified for input device elements, throw an exception. This seems the "strictly clean" solution, but I don't particularly like it because it requires a transform in order to chain composite devices.

Opinions welcome!

On whether the number of elements must match: yes they must match as the number of elements is intrinsic to the type of the tuple. That said, just because we're using namedtuples for the result of `value` we accept non-named tuples on setting (provided they have the same number of elements) as namedtuples are just a specialization of tuple.

If the number of elements doesn't match, one can use slicing (or concatenation) in a generator transform function to compensate. Assuming a (currently) fictional `ButtonBoard`:

```python
from gpiozero import LEDBoard, ButtonBoard

def first(it, n=1):
    for value in it:
        yield value[:n]

def last(it, n=1):
    for value in it:
        yield value[-n:]

def append_ones(it, n=1):
    for value in it:
        yield value + ((1,) * n)

buttons = ButtonBoard(5, 6, 7, 8)
leds = LEDBoard(11, 12, 13, 14)

# Map all buttons to LEDs
leds.source = buttons.values

# Map the first two buttons to LEDs and set the rest always on
leds.source = append_ones(first(buttons.values, 2), 2)

# And so on...
```

On whether there should be a variety of generator transforms built into the library: I'm quite torn on this question. I'd love to do it but there's two things holding me back: firstly I'm not convinced gpiozero is actually the right place for them (I'm currently working on something which might be: a library for linking arbitrary stuff like gpiozero to any other library based on a suggestion from Ben that it'd be nice to link it to picamera so you could control, e.g. the camera brightness with a pot). Secondly, it almost feels like I'd be removing a really great learning opportunity. Not that that's actually a good reason to exclude it from the library (one can always ignore a library when teaching and then reveal that what the students have written is already implemented for them).

On composites of composites: I would nest the tuples instead of flattening (arbitrary depth flattening is generally a bad idea for reasons too long to go into here). This should lead to a natural progression of names in the namedtuple classes. Consider the SnowPi example. I wouldn't bother with Ben's `SnowPiArms` class but I would keep `SnowPiArm` and simply instantiate it twice in `SnowPi`: once for left, once for right. Then `SnowPi`'s namedtuple would look something like: `(eyes, nose, left, right)`, while `SnowPiArm`'s namedtuple would be `(top, middle, bottom)` and `SnowPiEyes` would be `(left, right)`.

So, you'd access the value of the left eye as `my_snow_pi.value.eyes.left`, or the value of the middle LED on the right arm as `my_snow_pi.value.right.middle`, or get a tuple of the left arms LED values as `my_snow_pi.value.left`, and so on.

Finally, would the code work identically if switched? Basically yes; it's just a question of which background reading thread gets set up first. So there's a tiny (microsecond level) difference in the start time of one thread from the other, but otherwise no difference.

lurch:
I'm glad my "awkward questions" are giving you plenty of food for thought! :-)

On the topic of TrafficLightsBuzzer not including the button in its `value` - I agree that it should (I'll add that to #164), and perhaps the button was overlooked because it's not featured in the classname? Although TrafficLightsBuzzerButton is a bit of a mouthful ;)

I agree with your middle option of accepting but ignoring any output devices in the `value` tuple - because it means that if you had two TrafficLightsBuzzer boards, you could then just do `board2.source = board1.values`.

I guess your `yield value + ((1,) * n)` example could instead be written as `yield chain(value, repeat(1, n))` ? :)

If you're going to move each SnowPiArm directly inside SnowPi (so that it'd be `my_snow_pi.left` instead of `my_snow_pi.arms.left`) would it be better to rename them to `leftarm` and `rightarm`, as otherwise it's not immediately clear what `my_snow_pi.left` is referring to?  Although personally I can't see the harm of leaving SnowPiArms in there. \*shrug\*

And on the topic of not flattening nested CompositeDevices, why does TrafficLightsBuzzer's `value` return all the lights individually, and not as a TrafficLightTuple ?   Or is this one of those "use your best judgement" things when deciding how the `value` for each CompositeDevice should work?
I guess if TrafficLightsBuzzer's `value` *had* included a TrafficLightTuple, then you could have hooked up separate TrafficLightsBuzzer and TrafficLights boards and then do "TrafficLights.source = TrafficLightsBuzzer.value.lights" ?

lurch:
...and I guess this is a prime candidate for the new LEDCollection class? ;-)

I don't have a SnowPi myself, but I guess it could be fun to use the `blink()` method to e.g. make the left arm and right arm flash at different rates :)
(and this would also be a nice example of the sort of thing you can easily do using the generators built into GpioZero, but would be much harder just using RPi.GPIO and a simple while-loop!)

bennuttall:
Ok I think I've done it successfully now. I've done the following:

- Implement blink and pulse on SnowPiArms and SnowPi
- Add close() method
- Ensure self.leds is always a tuple of LEDs/PWMLEDs

I'm no longer getting the error `AttributeError: 'SnowPi' object has no attribute '_source_thread'` on exiting the shell, and everything seems to work as expected (besides me occasionally forgetting where to tell it to stop blinking - would be awesome to be able to close all blink threads from the root device).

However, I feel I've used much duplication. I've copy-pasted the blink and pulse method multiple times now and it's giving me anxiety. I wanted to commit the working version now but it'd be good if this could be cleaned up later.

@waveform80 are you happy with this implementation?

waveform80:
It looks good - however, at the moment I'm re-working the CompositeDevice hierarchy based on @lurch's suggestions of common functionality - so a lot of the copy-pasted stuff can probably disappear once that gets merged. I've got a CompositeDevice that deals with all the value/tuple stuff now, and a descendent CompositeOutputDevice that implements on/off/toggle/blink for all subordinate output devices (with a fix for #189) - but I've still got to write docs and tests for it all. Any chance you can wait until the weekend when I'll have a chance to finish all this stuff? (too busy with picademy right now!)

bennuttall:
That sounds sensible. Yes - it can wait :)

ryanteck:
As long as its done by the 1st of November 2016 there's no rush at all on a
Snowman Now :)

*Thanks -*

*Ryan Walmsley*
*Ryanteck LTD.*
*Email: *Ryanteck@gmail.com
*Skype:* Ryanteck
Website:* Http://ryanteck.org.uk <Http://ryanteck.org.uk>*

lurch:
Sorry to drag things off-topic again, but going back to the CompositeDevice stuff that I was discussing earlier...
I noticed recently that the [TrafficLightsBuzzer](http://gpiozero.readthedocs.org/en/latest/api_boards.html#trafficlightsbuzzer) init method takes a `TrafficLights` object, a `Buzzer` object, and a `Button` object; whereas the [Robot](http://gpiozero.readthedocs.org/en/latest/api_boards.html#robot) init method takes two tuples (each of two integers).
i.e. we have `TrafficLightsBuzzer(TrafficLights, Buzzer, Button)` and `Robot((int, int), (int, int))`

I guess if we were aiming for "interface consistency" then either TrafficLightsBuzzer's init should be `TrafficLightsBuzzer((int, int, int), int, int)`, or Robot's init should be `Robot(Motor, Motor)` (personally I'd opt for the latter). Is there any reason for this inconsistency and/or guidelines as to when to favour one approach over the other? (i.e. having the `__init__` method accept objects or lists-of-integers)

OTOH, having LEDBoard accept a list of ints, *does* seem to make more sense than passing in a list of LED objects. Hmmmm.....

lurch:
Hmmm, partially-answering my own question: the `TrafficLights`, `Buzzer` and `Button` init methods all take extra optional parameters, which it wouldn't make sense to try and squeeze into the `TrafficLightsBuzzer` init method; whereas (currently) the `Motor` init method **only** accepts two integers, with no extra optional parameters. Still, maybe it would have made sense to have Robot's init method accept two Motor objects, just in case the Motor class *does* grow extra optional arguments?

Would it be possible for Robot's `__init__` to support *both* `Robot((int, int), (int, int))` and `Robot(Motor, Motor)` by using `isinstance` on the `left` and `right` args?  And I guess with the pin-refactoring, it should also support `Robot((Pin, Pin), (Pin, Pin))`?
(and perhaps even `Robot((Pin, Pin), Motor)` or `Robot(Motor, (int, int))`, if that's not going too far down the rabbit hole)

PR #133, opened 2015-12-17
Add DotsBoard interface

bennuttall:
I'll explain later...

### DIFF
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -20,6 +20,7 @@ from .input_devices import (
     AnalogInputDevice,
     MCP3008,
     MCP3004,
+    Dot,
 )
 from .output_devices import (
     OutputDeviceError,
@@ -42,4 +43,5 @@ from .boards import (
     Robot,
     RyanteckRobot,
     CamJamKitRobot,
+    DotsBoard,
 )
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -12,7 +12,7 @@ except ImportError:
 from time import sleep
 from collections import namedtuple
 
-from .input_devices import InputDeviceError, Button
+from .input_devices import InputDeviceError, Button, Dot
 from .output_devices import OutputDeviceError, LED, PWMLED, Buzzer, Motor
 from .devices import CompositeDevice, SourceMixin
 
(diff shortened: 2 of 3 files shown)


bennuttall:
@rzrzr created an add-on called the DOTs Board, which has been used as an activity at Raspberry Pi's stands at various events like Maker Faire. It's a conductive paint dot-to-dot activity.

Dots board

The way it works is by painting over a dot, you short a pin to ground. When implementing the GPIO code last year using RPI.GPIO, I found that pulling up an input on setup, then reading it at some point later did not work. I had to set each pin up as an input, leaving it floating (no pull-up/down), then at the moment I wanted to check it, pull up and read.

I'd like to create a GPIO Zero interface to the board to allow people to easily make their own games and applications with the board.

In order to incorporate this into gpiozero, I created a `Dot` class which does not currently inherit any GPIO device class, because I did not want to have to register the pull- up/down until reading. This means it does not currently register in the GPIO pin lock set.

I also created a `DotsBoard` class which just extends `CompositeDevice`. I did not extend `SourceMixin` due to the nature of the board using all available GPIO pins (even the special use GPIO 0 and 1) so there's no way of connecting more devices (without using a compute module, and the intersection of Dots Board users with compute module users... you get the idea). I guess source could be used by non-GPIO things, like on-screen stuff? But I'm not sure how to get source to work so I've left it out for now.

I'm not 100% happy with the implementation, nor the choice of method names, but it's working code and it seems useful. I hope to get something like this into the library soon.

Any thoughts, @waveform80?

lurch:
Maybe there could also be e.g. :
```python
    def all_specific_dots_selected(self, specific_dots, exact_match=False):
        return all(dot in self.selected_dots for dot in specific_dots) and (not exact_match or (len(self.selected_dots) == len(specific_dots)))

    def any_specific_dot_selected(self, specific_dots):
        return any(dot in self.selected_dots for dot in specific_dots)
```

waveform80:
Hang on ... pulling on setup didn't work, but pulling up and reading later on did? That's ... hmm ...

Otherwise this basically sounds like a ButtonBoard (if that were the equivalent of what LEDBoard is to LED) although I'd have to understand a bit more about GPIOs 0 and 1 - haven't used those yet. Anyway, I'll take a look at the code in a bit - sounds like a good thing to have in the library at any rate.

lurch:
GPIOs 0 and 1 are 'reserved' for use by the HAT EEPROM ( https://github.com/raspberrypi/hats#new-add-on-boards-basic-requirements ) over i2c, and so (AFAIK) shouldn't normally be used by end-users. In most GPIO-breakout diagrams they're labelled as ID_SC and ID_SD, or DNC.

bennuttall:
Pins 0 and 1 can be used as regular GPIO pins if you don't mind ignoring this:

> The ID_SC and ID_SD pins must only be used for attaching a compatible ID EEPROM. Do not use ID_SC and ID_SD pins for anything except connecting an ID EEPROM, if unused these pins must be left unconnected

I believe they're reserved for reading a HAT's ID via EEPROM, though I'm not sure why using them is bad.

`ButtonBoard` would be good. I'll do some more testing on whether reading dots as if they're buttons really doesn't work...

lurch:
Just been looking at this with @bennuttall this evening (first time I've seen a dotsboard IRL ;) )

It's _really_ weird, but yeah it seems like if you treat the 'dots' as buttons (i.e. created with `pull_up=True`) then you don't get the values you expect, i.e. some of the dots report they're not connected to GND, when actually they are.
However going back to the 'original' way of doing it, by only turning on the pull-up just before reading the dot, and you get all the right values out :-/
My only conclusion is that there's something 'special' about the conductive paint, which means you can't just use it as an open-or-closed switch, and instead have to use a 'special' reading method :-S

bennuttall:
Ok so it looks like we need the pull behaviour from this PR but the API from #340.

I wonder, could we allow `OutputDevice` to change its pull after init to make this functionality possible?

lurch:
> I wonder, could we allow OutputDevice to change its pull after init to make this functionality possible?

I think as the dotsboard is a special-case, then the special-case code should be restricted to dotsboard, rather than 'polluting' the base OutputDevice ?

lurch:
> My only conclusion is that there's something 'special' about the conductive paint, which means you can't just use it as an open-or-closed switch, and instead have to use a 'special' reading method :-S

Hmm, or maybe having all the pullups enabled at once, means each individual pullup effectively becomes weaker?  I'm not a hardware engineer... ;-)

waveform80:
> Hmm, or maybe having all the pullups enabled at once, means each individual pullup effectively becomes weaker? I'm not a hardware engineer... ;-)

Reading through this I had a similar thought: no so much that the pull-ups become weaker but I have heard that, for example, trying to switch every single exposed GPIO on simultaneously is more or less impossible as there's an overall current limit that gets exceeded (can't remember the numbers though). If that's the case, then setting all GPIOs with pull-ups might have much the same effect - the GPIOs can no longer supply enough current to make them appear on when grounded.

Either way, this sounds like an electrical issue. Unfortunately, that also means there's no way we can do this with the regular Button class (with that, pull-up has to mean "always pulled up" rather than "pulled up when read" because people may be relying, electrically, on the GPIO being high when inactive.

That said, we could potentially make an alternate InputDevice descendent suitable for compositing into this, but it is more work.

PR #137, opened 2016-01-06
Add missing InputDeviceError import to devices.py

alexwlchan:
Line 316 raises an InputDeviceError, but the import for that error seems to be missing.

### DIFF
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -12,6 +12,7 @@ from collections import deque
 
 from RPi import GPIO
 
+from .input_devices import InputDeviceError
 
 _GPIO_THREADS = set()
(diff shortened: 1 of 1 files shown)


bennuttall:
Thanks!

waveform80:
Eurgh, sorry, that's my fault. Unfortunately, this PR introduces a circular dep between devices and input_devices. Not exactly an issue in the existing master but it's broken a couple of experiments I've been running with pins. One of the things I've been meaning to do is split the exceptions into their own module which would remove the circular dep and clean things up a little. I'll get this done with the metaclass fix-up I mentioned the other day, then I can rebase the pins stuff and get it pushed.

Issue #129, opened 2015-12-03
TrafficLightsBuzzerTuple unused?


bennuttall:
`TrafficLightsBuzzerTuple` is declared in `boards.py` but is not used.

waveform80:
Fixed in #138

PR #138, opened 2016-01-06
Fix return type of TrafficLightsBuzzer.value

alexwlchan:
The FishDishTuple was renamed to TrafficLightsBuzzerTuple in
commit b87e2a4, but the `value` property wasn't modified to match.

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -211,7 +211,7 @@ class TrafficLightsBuzzer(SourceMixin, CompositeDevice):
         the LEDs, and the buzzer. This property can also be set to a 4-tuple
         to update the state of all the board's components.
         """
-        return FishDishTuple(
+        return TrafficLightsBuzzerTuple(
             self.lights.red.value,
             self.lights.amber.value,
             self.lights.green.value,


bennuttall:
Ah, this fixes #129, right?

alexwlchan:
Looks like it.

Probably best to double-check with @waveform80 to be sure that's what he meant to do when he renamed the tuple.

waveform80:
Oh good grief ... yup, that's my stupid fault too, and the fix is indeed correct. Right after I've sorted out this pins stuff I really need to get on with a proper test suite to avoid nonsense like this.

alexwlchan:
For whatever it's worth, this and #137 were both spotted just by running flake8 over the codebase.

bennuttall:
Yeah probably worth bearing that in mind. I ran pylint over a file yesterday which contained more useful information than I'd expected! We do have a test suite in development but other things happening alongside.

waveform80:
I'd love to have a done a test suite by the time of 1.0 (as I had with picamera), but there was a bit of a rush to hit the jessie release so I'm afraid I'm rather behind with that. I can't say I'm terribly enamoured of automated test tools historically, though I am trialing some with my picraft project at the mo, and they've evidently improved significantly in the years since I last tried them. Perhaps I should add something similar to gpiozero when I get moving on the test suite (picraft's currently using codecov and landscape).

PR #141, opened 2016-01-07
Refactor pins implementation

waveform80:
@bennuttall - I finally got around to pushing it, but *don't merge this without testing it*! I've played around with it on my test board, but I'd really like some independent confirmation that I haven't screwed things up (especially given the current lack of a test suite). Still, there's some fun stuff in here including native PyPy GPIO support :)

-------

This commit is a fairly major piece of work that abstracts all pin operations (function, state, edge detection, PWM, etc.) into a base "Pin" class which is then used by input/output/composite devices to
perform all required configuration.

The idea is to pave the way for I2C based IO extenders which can present additional GPIO ports with similar capabilities to the Pi's "native" GPIO ports. As a bonus it also abstracts away the reliance on the
RPi.GPIO library to allow alternative pin implementations (e.g. using RPIO to take advantage of DMA based PWM), or even pure Python implementations (both of which are included in this PR; the native implementation works nicely in PyPy but doesn't support PWM currently, while the RPIO implementation requries a model 1 Pi, and root access but supports hardware-based PWM albeit with a currently fixed frequency).

The major downside to this PR is that is breaks backward compatibility for the `pin` property. For example, accessing `led.pin` will now return a Pin derived object rather than the GPIO pin number that the LED is attached to (you can access `led.pin.number` to get the GPIO pin number, but the `__repr__` of all pin implementations is reasonably friendly so if you just query it from the console you'll see "GPIO21" for example).

### DIFF
--- a/docs/api_pins.rst
+++ b/docs/api_pins.rst
@@ -1,0 +1,84 @@
+====
+Pins
+====
+
+.. currentmodule:: gpiozero
+
+As of release 1.1, the GPIO Zero library can be roughly divided into two
+things: pins and the devices that are connected to them. The majority of the
+documentation focuses on devices as pins are below the level that most users
+are concerned with. However, some users may wish to take advantage of the
+capabilities of alternative GPIO implementations or (in future) use GPIO
+extender chips. This is the purpose of the pins portion of the library.
+
+When you construct a device, you pass in a GPIO pin number. However, what the
+library actually expects is a :class:`Pin` implementation. If it finds a simple
+integer number instead, it uses one of the following classes to provide the
+:class:`Pin` implementation (classes are listed in favoured order):
+
+1. :class:`gpiozero.pins.rpigpio.RPiGPIOPin`
+
+2. :class:`gpiozero.pins.rpio.RPIOPin`
+
+3. :class:`gpiozero.pins.native.NativePin`
+
+You can change the default pin implementation by over-writing the
+``DefaultPin`` global in devices like so::
+
+    from gpiozero.pins.native import NativePin
+    import gpiozero.devices
+    # Force the default pin implementation to be NativePin
+    gpiozero.devices.DefaultPin = NativePin
+
+    from gpiozero import LED
+
+    # This will now use NativePin instead of RPiGPIOPin
+    led = LED(16)
+
+In future, this separation should allow the library to utilize pins that are
+part of IO extender chips. For example::
+
+    from gpiozero import IOExtender, LED
+
+    ext = IOExtender()
+    led = LED(ext.pins[0])
+    led.on()
+
+.. warning::
+
+    While the devices API is now considered stable and won't change in
+    backwards incompatible ways, the pins API is *not* yet considered stable.
+    It is potentially subject to change in future versions. We welcome any
+    comments from testers!
+
+
+Abstract Pin
+============
+
+.. autoclass:: Pin
+    :members:
+
+
+RPiGPIOPin
+==========
+
+.. currentmodule:: gpiozero.pins.rpigpio
+
+.. autoclass:: RPiGPIOPin
+
+
+RPIOPin
+=======
+
+.. currentmodule:: gpiozero.pins.rpio
+
+.. autoclass:: RPIOPin
+
+
+NativePin
+=========
+
+.. currentmodule:: gpiozero.pins.native
+
+.. autoclass:: NativePin
+
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,6 +37,8 @@ class Mock(object):
 
 sys.modules['RPi'] = Mock()
 sys.modules['RPi.GPIO'] = sys.modules['RPi'].GPIO
+sys.modules['RPIO'] = Mock()
+sys.modules['RPIO.PWM'] = sys.modules['RPIO'].PWM
 sys.modules['w1thermsensor'] = Mock()
 sys.modules['spidev'] = Mock()
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,5 +12,6 @@ Table of Contents
     api_output
     api_boards
     api_generic
+    api_pins
     changelog
     license
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -281,7 +281,7 @@ Each button plays a different sound!
 
     buttons = [Button(pin) for pin in sound_pins]
     for button in buttons:
-        sound = sound_pins[button.pin]
+        sound = sound_pins[button.pin.number]
         button.when_pressed = sound.play
 
     pause()
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -5,6 +5,23 @@ from __future__ import (
     division,
 )
 
+from .pins.exc import (
+    PinError,
+    PinFixedFunction,
+    PinInvalidFunction,
+    PinInvalidState,
+    PinInvalidPull,
+    PinInvalidEdges,
+    PinSetInput,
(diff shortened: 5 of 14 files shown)


bennuttall:
Great. I'll test it out when I get chance!

bennuttall:
Sweet!

lurch:
I never got round to testing or experimenting with this refactoring; but something I kept meaning to ask after reading through the code was: Is the GPIOFS class supposed to be more fully-featured? e.g. so that it can provide a pure SysFs-based GPIO implementation (with obviously much-limited functionality) without having to dive down into /dev/gpiomem or /dev/gpio ? Currently GPIOFS seems entirely unused.
Although maybe it's something to consider fleshing-out for v1.2 (and this refactoring will presumably also make writing unit tests much easier).

And I also kept meaning to say that when I first heard about GpioZero, I had *hoped* it was a sufficiently abstracted high-level API to allow the underlying implementation to be swapped out, and I guess this PR shows that it is. Yay! :-D

waveform80:
Okay, I should probably write some stuff about the pure Python implementation if only to get it documented somewhere. Firstly: yup, the underlying implementation can definitely now be swapped out and I'm planning to add another implementation in 1.2 (pigpiod).

So ... the pure Python implementation. You're absolutely correct that at the moment GPIOFS isn't used, but I am planning to use it when I get around to refining the implementation in 1.2...

When I embarked on writing the pure Python implementation I wondered if it'd be possible to write a purely sysfs based implementation and ran a few experiments along those lines. Well ... the answer is: yes, it's possible and no you don't want to do it :)

It's basically possible to do everything from sysfs that you can do via /dev/mem or /devgpiomem  (configure input/output, pull up/down, read values, write values, set edge detection and efficiently wait on it via epoll) however it's *horribly* slow. Just to test the waters I wrote a couple of C programs which flipped the state of an output GPIO pin 10 million times. The first did it via /dev/gpiomem and finished in under a second (0.7s). The second relied purely on sysfs and file I/O. I made it as efficient as I could (never open the file more than once, etc.) and still couldn't get its runtime much below 30 seconds. So ... for controlling GPIO pins, sysfs will work but memory bit-bang is at least an order of magnitude faster (and given software PWM isn't great even at those speeds, sysfs would be a complete non-starter there).

I may return to it out of pure theoretical interest at some point in the future (after all it does mean avoiding /dev/mem or /dev/gpiomem entirely which may have some utility in strict security settings), but for now a pure sysfs implementation is on the back-burner. So, what's the purpose of GPIOFS?

At the moment, the pure Python implementation only uses GPIOMemory which provides mmap'd access to /dev/gpiomem (preferably) or /dev/mem (as a fallback) with some magic to make it look like an array of 32-bit words instead of bytes (there is a more efficient way of doing this than `__getitem__` and `__setitem__` but I haven't tested if it works with pypy yet, and it might not because it relies on some dirty ctypes stuff).

Now ... this is extremely similar to how both RPi.GPIO and RPIO do stuff - they both flip hardware registers to handle configuring GPIO pins for input / output, for pull-ups, and for output setting. However, both of these libraries avoid touching the event-detection registers; they both use epoll against sysfs files for handling edge detection. I thought that was a little odd (as in, there's probably some good reason they're avoiding those registers), but thought "what the hell, let's give 'em a whirl". If you give the native pin implementation a try, you'll see why they avoid them! The first time you configure a pin, the native implementation will try and set "both" edge detection on the pin. As soon as you do that you'll see something like this in dmesg:

```
[180028.554528] gpiomem-bcm2835 3f200000.gpiomem: gpiomem device opened.
[180028.768668] irq 49: nobody cared (try booting with the "irqpoll" option)
[180028.768692] CPU: 0 PID: 5641 Comm: python Not tainted 4.1.13-v7+ #826
[180028.768701] Hardware name: BCM2709
[180028.768739] [<80018444>] (unwind_backtrace) from [<80013e08>] (show_stack+0x20/0x24)
[180028.768760] [<80013e08>] (show_stack) from [<8055a188>] (dump_stack+0x98/0xe0)
[180028.768781] [<8055a188>] (dump_stack) from [<8006e2fc>] (__report_bad_irq+0x30/0xd0)
[180028.768800] [<8006e2fc>] (__report_bad_irq) from [<8006e6e8>] (note_interrupt+0x274/0x2d8)
[180028.768817] [<8006e6e8>] (note_interrupt) from [<8006bcdc>] (handle_irq_event_percpu+0xc4/0x254)
[180028.768835] [<8006bcdc>] (handle_irq_event_percpu) from [<8006beb8>] (handle_irq_event+0x4c/0x6c)
[180028.768852] [<8006beb8>] (handle_irq_event) from [<8006f0ac>] (handle_level_irq+0xa4/0x13c)
[180028.768870] [<8006f0ac>] (handle_level_irq) from [<8006b5a0>] (__handle_domain_irq+0x84/0xec)
[180028.768886] [<8006b5a0>] (__handle_domain_irq) from [<8000933c>] (asm_do_IRQ+0x2c/0x30)
[180028.768904] [<8000933c>] (asm_do_IRQ) from [<8055fe74>] (__irq_svc+0x34/0x14c)
[180028.768913] Exception stack(0xb4113ed0 to 0xb4113f18)
[180028.768926] 3ec0:                                     80801cdc 00000001 00000000 00000000
[180028.768939] 3ee0: 8002a914 807ddc74 00000001 00000000 00000000 76f0e04c 00000003 b4113f5c
[180028.768951] 3f00: b4113f18 b4113f18 8002a24c 8002a250 60000113 ffffffff
[180028.768970] [<8055fe74>] (__irq_svc) from [<8002a250>] (__do_softirq+0xc8/0x3d4)
[180028.768985] [<8002a250>] (__do_softirq) from [<8002a914>] (irq_exit+0xcc/0x13c)
[180028.769001] [<8002a914>] (irq_exit) from [<8006b5a8>] (__handle_domain_irq+0x8c/0xec)
[180028.769017] [<8006b5a8>] (__handle_domain_irq) from [<8000933c>] (asm_do_IRQ+0x2c/0x30)
[180028.769032] [<8000933c>] (asm_do_IRQ) from [<8056023c>] (__irq_usr+0x3c/0x140)
[180028.769041] Exception stack(0xb4113fb0 to 0xb4113ff8)
[180028.769052] 3fa0:                                     00000000 98aa9200 769092ac 7ecd5cd8
[180028.769066] 3fc0: 0000004c 768250f8 003c3848 0035b7bc 00000fb4 76f0e04c 00000003 767cc738
[180028.769078] 3fe0: 00000000 7ecd5d30 001a2fe0 0023c860 60000010 ffffffff
[180028.769085] handlers:
[180028.769101] [<8031e4e8>] bcm2835_gpio_irq_handler
[180028.769111] Disabling IRQ #49
```

So, something in the kernel is doing something with those registers (something to do with IRQs which I don't fully understand at the moment) and as soon as we touch them, it throws a wobbly. This wobbly doesn't seem to matter too much in practice - the native pin implementation carries on working smoothly from there but I'm sure there'll be some horrible effect I haven't discovered yet (it'll only do this once after a boot; basically it kills that module and until you reboot it won't come back).

This is why there's a warning on the native pins docs, and I'm sure it's why the aforementioned libraries don't touch those registers. Hence, in 1.2 I'll be following their lead, sorting out GPIOFS so it's a bit more full featured, and switching to that for edge detection, leaving the mmap stuff for everything else.

I'm half tempted to leave the current implementation around for anyone that wants it though. Eventually there might be a "default" native pin implementation (part mmap, part sysfs), a "pure mmap" implementation (the current one) with a big fat warning attached that it might eat Pis, and a "pure sysfs" implementation with a big fat warning that it'll act like its eaten too many Pis already!

PR #150, opened 2016-01-19
 Adding MCP3208/3204 12 bit ADC with 8 or 4 channels

pcopa:
The text needs more editing, but the code is running stable on my Pi 2.

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -508,13 +508,14 @@ class AnalogInputDevice(CompositeDevice):
     Represents an analog input device connected to SPI (serial interface).
     """
 
-    def __init__(self, device=0, bits=None):
+    def __init__(self, device=0, bits=None, mode=1):
         if bits is None:
             raise InputDeviceError('you must specify the bit resolution of the device')
         if device not in (0, 1):
             raise InputDeviceError('device must be 0 or 1')
         self._device = device
         self._bits = bits
+        self._mode = mode
         self._spi = SpiDev()
         self._spi.open(0, self.device)
         super(AnalogInputDevice, self).__init__()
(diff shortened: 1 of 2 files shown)


waveform80:
Doh - there's me going through my github notifications and you've not only made commits, you've made a PR too! Awesome stuff - okay, forget my ramblings about branches and what-not. They might be useful for the future but let's deal with this PR first. There's a couple of minor changes it'd be nice to see if you're up for making them?

Firstly there's an odd doc-string/comment *above* the MCP3208 class definition which should probably be removed (or turned into a proper doc-string for the class ... I ought to make a proper doc-string for MCP3008 for that matter!). The channel property's doc-string mentions MCP3008 rather than MCP3208 (and likewise for MCP3204). And finally, I think the third result byte (`data[2]`) needs to be right-shifted 2 bits before being |'d into the result (just looking at the diagram for the protocol; incidentally many thanks for including that in the comments - makes it *much* easier to figure out what the code's doing!).

Now the tricky bit: how to update the PR with your changes. There's two ways of doing this:

## Simple method

The easy method (but one which is generally frowned upon) is to make the changes, commit them to your clone (`git commit -a`) and push that to GitHub (`git push`). This will add the new commit to your PR (and you can keep doing this as many times as you like).

The reason this is frowned upon is that your PR potentially winds up containing lots of commits so if we merge it to master we see all those commits appear. Given that each issue/PR is really meant to make "one" change (although what a "change" is is rather ill defined), having a change associated with lots of commits makes it harder to track down bugs later on (simply because it makes it harder to find the before and after states as it means crossing more than one commit in the history).

Frankly I'm not that bothered in this case - it'd only be two commits and I don't foresee any serious debugging issues arising so if you want to go with this method, that's fine.

## Hard(er) method

Just in case you want to have a crack at the hard(er) method, here it is. In this case you make your changes as before and then commit them. However, instead of making a new commit you do something naughty and "overwrite" the existing commit. You do this with: `git commit -a --amend` (the "amend" simply means "change the latest commit instead of making a new one"). You'll find it comes up with the commit message you wrote instead of a blank one.

At this point you've done something naughty in git's eyes: you've re-written history. But it's not that bad because it's in a repo that only you care about (your forked clone). If we did this in the master repo, that would be seriously bad because now everyone with a fork would need to re-write their history to match too.

Now you need to push your re-written commit up to GitHub. However, because you've re-written history a normal push won't work - you need to "force" the push: `git push -f`. Again, this is a "dangerous" command; it's perfectly safe as long as you're working on your own private fork which no one else cares about but woe betide the developer who does this on a master repo that other people have forked (don't worry - you can't do this with the master repo :).

At this point you'll have an updated PR. So, there you go: it's not really *that* hard as method - it's just harder because it involves re-writing history. Frankly, it's not even really dangerous for you because you're working on your own fork. However, if one day you have write access to the master repo for a popular project the commands above need to be used with serious caution (that's the reason I work on a fork of gpiozero rather than the master repo itself!)

gpiozero/input_devices.py
```diff
@@ -601,3 +601,59 @@ def __init__(self, channel=0, device=0):
...
+        # MCP3208 protocol looks like the following:
+        #
+        #     Byte        0        1        2
+        #     ==== ======== ======== ========
+        #     Tx   0001MCCC xxxxxxxx xxxxxxxx
+        #     Rx   xxxxxxxx x0RRRRRR RRRRRRxx
+        #
+        # The transmit and receive bits are shifted 4 bits to the left comared
+        # to the MCP3008. This leaves 3 preamble bits to warm up witch sould
+        # be more than adiquid.
```
carlosperate: `adequate` instead of `adiquid`

gpiozero/input_devices.py
```diff
@@ -601,3 +601,59 @@ def __init__(self, channel=0, device=0):
...
+
+    def _read(self):
+        # MCP3208 protocol looks like the following:
+        #
+        #     Byte        0        1        2
+        #     ==== ======== ======== ========
+        #     Tx   0001MCCC xxxxxxxx xxxxxxxx
+        #     Rx   xxxxxxxx x0RRRRRR RRRRRRxx
+        #
+        # The transmit and receive bits are shifted 4 bits to the left comared
```
carlosperate: `compared` instead of `comared`?

pcopa:
@waveform80 thanks to your rambling, I've been able to get this far.  And yes i'm committed to continue improving this "ADC" part for now.  As mentioned before the text needs editing, but I was more interested in the Github process and code at this time. 
I am dyslexic, grown up with 4 languages, so therefor avoid writing without a spellchecker, but do like formatting and editing.  Sorry, that's my rambling.

I will have to read your text in more detail before commenting. But as I was getting my grips on the structure of this subject. I think the "ADC" or whatever we want to call it, is the main group and then the MCP's, or others, are the sub groups. 
There are also a lot of other interesting sensors beyond pot, like temp, humidity, pressure, light, PH, ....., or any combination, that all use some form of "ADC" and we should make it simple to integrate them. 
The "Pot" coverts the "ADC" output  between "0" and "1"  and for the others we have to convert the "ADC" output to something else.  I see the "ADC" output as the main "hidden" value that then gets coveted to whatever according to it's main class use.

I'll have to think it over in more detail and make a proposal later.

gpiozero/input_devices.py
pcopa: Thanks, done

gpiozero/input_devices.py
pcopa: Thanks, done

pcopa:
@waveform80 

> Firstly there's an odd doc-string/comment above the MCP3208 class

Did you mean the text between the "***"?  I assumed that this was just another way of inserting comments.  I have removed this text. (Will have to lookup doc -string)

> The channel property's doc-string mentions MCP3008 rather than MCP3208 (and likewise for MCP3204)

I've made the change

> And finally, I think the third result byte (data[2]) needs to be right-shifted 2 bits

Your absolutely right, 
In my case i"ve hooked up Vref to channel 2. This should give me the full range of 4095
The result from the Bytes going to the spi.xfer2 is 26, 0, 0 or B0011010  00000000  00000000
The Bytes received from the spi.xfer2 are 0, 63, 255 or B00000000  01111111  11111111
The last 2 bits are don't care and need to be masked out or shifted right 2 times.
So the formula is now ((data[1] & 63) << 6) | (data[2] >> 2)  and it works.

pcopa:
In MHO we could base the 3008/3004 on the 3208/3204 with the only exception that the last data[2] byte needs to be shifted 4 times and the 13 bit could also be included by only shifting once.

pcopa:
I've made a bold step and added the @property def bits(self): and now the formula for the MCP 3x08 return value is universal up to 14 bits.

((data[1] & 63) << (self.bits - 6)) | (data[2] >> (14 - self.bits)

pcopa:
1) Have restructured the MCP3004/3008/3204/3208 variants and new variants can be added as needed.
2) Have added an absolute ADC value to the existing relative value between 0 and 1.  This way the value can be used for other things like a thermometer that I intend to add later.

Please check for English spelling. I've already changed "Digital" to "Digital"

gpiozero/input_devices.py
```diff
@@ -551,53 +552,94 @@ def _read(self):
...
+        # The channel number must be 0, 1, 2 or 3 
         if not 0 <= channel < 4:
             raise InputDeviceError('channel must be between 0 and 3')
-        super(MCP3004, self).__init__(channel, device)
+        super(MCP3004, self).__init__(channel, device, bits=10)
+        self._channel = channel
+
+class MCP3008(MCP3xxx):
+    def __init__(self, channel=0, device=0):
+        # MCP3208 is a 10 bit Analog to Ditytal Converter with 8 channels.
```
pcopa: Spelling: change "Ditytal" to "Digital"

gpiozero/input_devices.py
```diff
@@ -551,53 +552,94 @@ def _read(self):
...
+        # The channel number must be 0, 1, 2, 3, 4, 5, 6, 7 or 8 
+        if not 0 <= channel < 8:
+            raise InputDeviceError('channel must be between 0 and 7')
+        super(MCP3008, self).__init__(device=device, bits=10)
+        self._channel = channel
+
+class MCP3204(MCP3xxx):
+    def __init__(self, channel=0, device=0, bits=12):
+        # MCP3204 is a 12 bit Analog to Ditytal Converter with 4 channels.
```
pcopa: Spelling: change "Ditytal" to "Digital"

gpiozero/input_devices.py
```diff
@@ -551,53 +552,94 @@ def _read(self):
...
+        # The channel number must be 0, 1, 2 or 3 
+        if not 0 <= channel < 4:
+            raise InputDeviceError('channel must be between 0 and 3')
+        super(MCP3204, self).__init__(channel, device, bits)
+        self._channel = channel
+
+class MCP3208(MCP3xxx):
+    def __init__(self, channel=0, device=0, bits=12):
+        # MCP3208 is a 12 bit Analog to Ditytal Converter with 8 channels.
```
pcopa: Spelling: change "Ditytal" to "Digital"

gpiozero/input_devices.py
```diff
@@ -551,53 +552,94 @@ def _read(self):
...
 
+        data = self._spi.xfer2([(24 + self.channel), 0, 0])
+        return ((data[1] & 63) << (self.bits-6)) | (data[2] >> (14 - self.bits))
 
-class MCP3004(MCP3008):
+class MCP3004(MCP3xxx):
     def __init__(self, channel=0, device=0):
-        # MCP3004 protocol is identical to MCP3008 but the top bit of the
-        # channel number must be 0 (effectively restricting it to 4 channels)
+        # MCP3004 is a 10 bit Analog to Ditytal Converter with 4 channels.
```
pcopa: Spelling: change "Ditytal" to "Digital"

gpiozero/input_devices.py
```diff
@@ -551,53 +552,94 @@ def _read(self):
...
 
 
-class MCP3008(AnalogInputDevice):
-    def __init__(self, channel=0, device=0):
+
+class MCP3xxx(AnalogInputDevice):
+    def __init__(self, channel=0, device=0, bits=None):
         if not 0 <= channel < 8:
             raise InputDeviceError('channel must be between 0 and 7')
-        super(MCP3008, self).__init__(device=device, bits=10)
```
pcopa: Code is redundant in this spot and will be removed

gpiozero/input_devices.py
```diff
@@ -551,53 +552,94 @@ def _read(self):
...
-        return self._read() / (2**self._bits - 1)
+        return self._read(), (self._read() / (2**self._bits - 1))
 
 
-class MCP3008(AnalogInputDevice):
-    def __init__(self, channel=0, device=0):
+
+class MCP3xxx(AnalogInputDevice):
+    def __init__(self, channel=0, device=0, bits=None):
         if not 0 <= channel < 8:
```
pcopa: Code is redundant in this spot and will be removed

gpiozero/input_devices.py
```diff
@@ -551,53 +552,94 @@ def _read(self):
...
+        super(MCP3008, self).__init__(device=device, bits=10)
+        self._channel = channel
+
+class MCP3204(MCP3xxx):
+    def __init__(self, channel=0, device=0, bits=12):
+        # MCP3204 is a 12 bit Analog to Ditytal Converter with 4 channels.
+        # The channel number must be 0, 1, 2 or 3 
+        if not 0 <= channel < 4:
+            raise InputDeviceError('channel must be between 0 and 3')
+        super(MCP3204, self).__init__(channel, device, bits)
```
pcopa: edit error: Change bits to bits=12 to force 12 bits.

waveform80:
Sorry for the recent silence - horribly busy! Great to see you've managed to update the PR; that's great stuff (and many thanks for trying the harder route - it'll help with the few final changes that are needed as we'll just wind up with one nice clean commit :).

On the idea of making a generic MCP3xxx class: nice idea but I'm not as sure it's as simple as you've got it at the moment. Check the protocol for the bytes going to the MCP3008 (here's the [relevant bit](https://github.com/pcopa/python-gpiozero/commit/b89ecdbea13f74e095cd366a795ca7817eea00c5#diff-58c739be690512be707a7a9a3e174730L581) of the diff). As you've got it in your version I think it's different to the original MCP3008 transaction: the send bytes need shifting by different amounts for one.

I'm not opposed to the idea of the generic MCP3xxx class - it's a nice plan but if you want to go that route I think you'll need a little more finesse. Specifically, beyond "bits" you're going to need how much to shift the send bytes, and where to split the receive bytes because these differ.

pcopa:
Thanks for the comments. At first I didn't think it would be possible, but then I found a code on the internet, that did used the same code for both.  I've studied the two data sheets excessively and the MCP300x even starts with the start bit right away and without any preamble bits. I even give it 3 zeros. that should be more than adequate.
As a computer hardware engineer myself I believe this should work, but of coarse it's got to be tested.

Iv'e tested the MCP3208 and it works fine, but who has a MCP3008 and is willing to test it?  
The algorithm to shift bits is implemented and works for bits 6 to 14.  

At this point I don't want to make the MCP3xxx a public class, that's why I've not added it to the __init__  
I'm trying to keep things as simple as possible. The user only has to import the chip and every thing's loaded with thee defaults. The only two parameters that can be are overwritten, are channel and device.

waveform80:
Intriguing! @bennuttall @martinohanlon - either of you got an MCP3008 kicking around still? If so could you possibly clone pcopa/python-gpiozero and give it a whirl? If it works with MCP3008 too then this is good enough to merge (I can always clean up spelling after - that's minor stuff). I'm keen to get this in fairly quickly because I've got some fairly major stuff to merge into master which'll likely clobber the odd line here (and I'm not keen to start teaching rebasing PRs at this point - that's a bit more hairy :).

On keeping MCP3xxx out of `__init__`: that's perfect, exactly what I would've done. And yeah - I wasn't aware this might work but if it does then that's fantastic - I always like a simple implementation!

bennuttall:
I'll try to give it a try this week!

pcopa:
I've just ordered two of each Mcp3008/3208/3304 at Moser, but it will take
10 days before I get them.
!!Yes!!, there's a new MCP 330x series. This is a 13 bit device and as far
as I can tell from the datasheet, fits perfectly in my scheme.

The MCP 3301 can be found at:
http://ww1.microchip.com/downloads/en/DeviceDoc/21700D.pdf

The MCP 3302/04 can be found at:
http://ww1.microchip.com/downloads/en/DeviceDoc/21697F.pdf

 I'm intending to add these 3 variances /Classes, do some cleanup and
cosmetics and then upload a new commit.

Now that I'm beginning to understand this a little bit better, i'll start
tackling the documentation

gpiozero/input_devices.py
```diff
@@ -551,53 +552,185 @@ def _read(self):
...
+class MCP3304(MCP3xxx):
+    """
+    The MCP3304 is a 13 bit Analog to Digital Converter with
+    4 channels single ended or 2 channels diferential.
+    The channel number must be 0, 1, 2, 3, 4, 5, 6, 7 or 8
+    """
+    def __init__(self, channel=0, device=0, mode=1):
+        if not 0 <= channel < 8:
+            raise InputDeviceError('channel must be between 0 and 7')
+        super(MCP3304, self).__init__(device=device, bits=13)
```
pcopa: need to add "mode=mode"

pcopa:
I've received the MCP3008, MCP3208 and MCP3304 and did some extensive testing. 
The program worked fine on all the MCP3008 and MCP3208 variances including the mode bit so these are good to go.

But, even though the MCP3304 should have the exact same timing as the MCP3008 and MCP3208, according to the data sheets, the Rx bits were shifted by 1 bit to the right.
I've adapted the code to handle this shift for the 13 bit MCP330x family, but am uncomfortable with the fact that this chip is behaving different. It could be a timing issue in the spi.xfer2 module.

Has anyone else seen a shift in the Rx bits using the spi.xfer2 module?

waveform80:
Great stuff! There's a few things I'm going to change, but now it's been tested with the 3008 and 3208 I'm happy enough to merge this and press on with the other stuff that's going into 1.1. The things that I'll be changing (partially for my own reference) are:

* No need for `relVal` as it's just the same as value
* Rename `absVal` to `raw_value`. Firstly it's PEP-8 compliant; no camelCase in prop/method name, and secondly the whole rel/abs thing is just going to confuse people what with single/differential being supported too. It's really the raw-reading from the device without bit correction so "raw_value" seems a better name.
* Remove `mode` from the base `AnalogInputDevice` class; it's (so far) specific to MCP3xxx devices
* Rename `mode` in `MCP3xxx` to `differential`; "mode" could mean so many things that it pretty much demands people have to read the docs to figure out the meaning. If we simply call it "differential" and make it a straight-forward bool (default `False`), then most people should figure out the meaning without having to resort to the docs.

Anyway, I can't thank you enough firstly for sticking with my rambling explanations of Git and GitHub, secondly for providing an excellent PR, and finally for testing it with a variety of chips (and expanding the functionality to cope with all of them)! Brilliant work :)

waveform80:
Hmm ... just reading the data-sheets for the MCP330x series and it looks like they support full differential operation, providing potentially negative outputs (looks like in single-ended mode they're just 12-bit as the two's-complement sign bit is always 0). We should probably scale these from -1 to +1 when operating differentially instead of 0 to 1 (this'll be useful with things like the Motor class). Also looks like the MCP3301 only operates in differential mode (there's no input at all; it just outputs when the clock says to).

waveform80:
Also, I think the output for the 3302/3304 *is* the same alignment as the 3204/3208; it's just the section on using it with a micro-controller's SPI port does have everything right-aligned which might've caused the confusion. Anyway, I'll get all this done up nicely and we'll just call 330x support "preliminary" until someone can give it a good testing.

pcopa:
Thanks @waveform80 for all the complements and cleaning up. All your comments make sense and I've just checked out "/input_devices.py#L741" and like the structure. The only thing I found is on lines 741 and 774, the "mode" could be changed to something like "differential", now that we'r using "differential", but  differential is "0" and not "1".... I'm sure you can come up with a text.

Now I need to do some more reading on "differential" and binary before I start testing.

pcopa:
Looks good. Thanks for all your explanations and help.

PR #153, opened 2016-01-31
Move LED example from GPIO2 to GPIO17

### LINKED ISSUE #116: Use different pin for LED examples?
The LED recipes in http://pythonhosted.org/gpiozero/recipes/ (and also the example in http://pythonhosted.org/gpiozero/notes/ ) use GPIO 2, which is one of the I2C pins. This means that when you connect an LED to it, the weak pull-up on that pin means the LED dimly illuminates, before actually running any Python code!  (which of course may be surprising / confusing for newbies).

Maybe the LED recipes should be modified to use GPIO 17, so that they match http://pythonhosted.org/gpiozero/outputs/ ?

And / or maybe a note could be added to the Notes page about the weak pullups on GPIOs 2 & 3 ?

lurch:
Fixes #116

### DIFF
--- a/docs/notes.rst
+++ b/docs/notes.rst
@@ -13,7 +13,7 @@ The following script looks like it should turn an LED on::
 
     from gpiozero import led
 
-    led = LED(2)
+    led = LED(17)
     led.on()
 
 And it does, if you're using the Python (or IPython or IDLE) shell. However,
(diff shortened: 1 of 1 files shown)


lurch:
WRT #116 the [TrafficLights example](http://gpiozero.readthedocs.org/en/latest/recipes.html#traffic-lights) still uses GPIO2 and GPIO3, but I don't have a traffic-lights kit, so I dunno if the weak pull-ups are still an issue or not?

Issue #41, opened 2015-09-30
MCP3008 / Potentiometer


bennuttall:
I've incorporated @martinohanlon's MCP3008 class, and added a couple of examples to the recipes in the docs, including a cool one for controlling an RGB LED with 3 pots.

`MCP3008` could be used for a range of analogue inputs. Is this sufficient an interface, or should we provide wrappers for specific components?

martinohanlon:
How about creating a super class of ADC which describes the interface for an ADC of which an MCP3008 is one of them. You could then create wrappers for analog components (e.g. potentiometer) and you pass a reference to your ADC when you instantiate it 

myadc = MCP3008(channel = 2)
mypot = Potentiometer(myadc)

or

myadc = ADS1115(address=0x48)
mypot = Potentiometer(myadc)

waveform80:
Did you mean the `open` method to be called from `__enter__`? That implies that either users *have* to use the class as a context manager, or that they have to call `open` manually after construction. The more usual pattern is to perform initialization implicitly on construction, have `__enter__` return self, and then call `close` from `__exit__`.

waveform80:
BTW - I haven't had a chance to play with ADC or pots generally yet (at least in connection with gpiozero), but @martinohanlon's suggestion looks very nice. My one concern at the moment is that people may wish to use pots in a similar manner to the LDR class (in series with a capacitor, time the charging and average it for a value) but we could make this compatible with the suggestion: derive from SmoothedInputDevice, detect if pin is a plain integer and treat it in a similar manner to the LDR circuit if so; alternatively if it's a derivative of some generic ADC class, act accordingly - e.g. reduce queue_len to 1 (in which case I'd rename the `read` method to `_read` and let the rest of the SmoothedInputDevice machinery take over).

martinohanlon:
When I wrote the MCP3008 class, I wanted to give the consumer the option of using with if they wanted too, so you could consume it as either

    with MCP3008() as myadc:
        myadc.read()

which would have the benefit of opening and closing the SPI channel 'automatically'

or you could implement it as:

    myadc = MCP3008()
    myadc.open()
    try:
        myadc.read()
    finally:
        myadc.close()

I vaguely remember that for one of the use cases I had the above was more 'useful' for me, but I cant think why!

This was my first 'attempt' at creating a class which used the context manager, if there is a better way lets change it.

waveform80:
[...]

PR #156, opened 2016-01-31
Typo fix

lurch:


lurch:
Also, would there be any harm in adding classes to cover the MCP3001, MCP3002, MCP3201 and MCP3202 chips? https://github.com/RPi-Distro/python-gpiozero/issues/41#issuecomment-167552148
 (although I've not studied the dataseheets yet, I'd be very surprised if they use a different protocol?)

waveform80:
Doh, good catch! I'll have a look at those datasheets when I get a sec because if they're as similar as we suspect then, sure, the more the merrier!

PR #159, opened 2016-01-31
Fix prototypes in docs

waveform80:
The prototypes in the docs are rigged to make out the first parameter as
mandatory (as it effectively is); however this does mean you've got to
remember to update the prototype when you modify it in the code! :)

### DIFF
--- a/docs/api_output.rst
+++ b/docs/api_output.rst
@@ -16,25 +16,25 @@ everyday components. Components must be wired up correctly before use in code.
 LED
 ===
 
-.. autoclass:: LED(pin, active_high=True)
+.. autoclass:: LED(pin, active_high=True, initial_value=False)
     :members: on, off, toggle, blink, pin, is_lit
 
 PWMLED
 ======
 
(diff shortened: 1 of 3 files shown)


docs/api_output.rst
```diff
@@ -16,25 +16,25 @@ everyday components. Components must be wired up correctly before use in code.
...
 
-.. autoclass:: Buzzer(pin, active_high=True)
+.. autoclass:: Buzzer(pin, active_high=True, initial_value=False)
     :members: on, off, toggle, beep, pin, is_active
 
 RGBLED
 ======
 
-.. autoclass:: RGBLED(red, green, blue)
+.. autoclass:: RGBLED(red, green, blue, active_high=True)
```
lurch: I dunno if it might be taking things too far "just for the sake of it", but would it make sense to have an `initial_values` parameter which accepts a 3-tuple, to set the initial colour of the RGBLED?
Hmmm, probably not...

Issue #125, opened 2015-11-29
Adds a *args to LEDBoard on() and off() so individual pins can be controlled


martinohanlon:
At the moment, the LEDBoard class only allows you to turn on or off all the LEDs. The methods could be enhanced so optional led positions can be passed to the on and off methods to control individual leds.

While this can be done through setting a value within the properties its doesnt feel very 'friendly'.

martinohanlon:
Happy to assign this to myself.

waveform80:
Sounds good to me - this would allow control of multiple LEDs through a single call rather than having to make multiple calls to, e.g. `board.leds[0].on(); board.leds[1].on()` etc.

One question: should the `*args` be pin numbers as in the constructor, or LED indexes as in the `leds` property. I'd lean toward the latter for consistency with existing code (in particular I'm thinking about the classes that derive from LEDBoard) but I'd be interested in counter arguments?

Also, @bennuttall any thoughts?

martinohanlon:
Definitely indexes of the leds property. It would seem counter intuitive to access items within an abstract class by a specific value (which could change across implementations).

bennuttall:
Sorry I missed this one. I actually came here to make a similar comment myself!

I started using `LEDBoard` to control a row of LEDs and realised I'd be better off using LED in a list comprehension. The only thing `LEDBoard` provides is the ability to control them all together with on/off/blink/toggle.

I'm currently using:

```python
n = 2
leds = [LED(i) for i in (2, 3, 4)]
[led.on() for i, led in leds if n > i]
```

to turn on the first two.

waveform80:
Do we have a PR on this yet? I see @lurch has some commits which might form one, and @martinohanlon's volunteered to take this on?

Incidentally, just to counter the idea that using `value` isn't friendly - here's a little something for alternate patterns:

```python
from gpiozero import LEDBoard
from time import sleep

board = LEDBoard(5, 6, 13, 19, 26)
while True:
    board.value = (1, 0, 1, 0, 1)
    sleep(1)
    board.value = (0, 1, 0, 1, 0)
    sleep(1)
```

Or in source/values form (using the new source_delay stuff in 1.2):

```python
from gpiozero import LEDBoard
from signal import pause

board = LEDBoard(5, 6, 13, 19, 26)

def alternating():
    while True:
        yield (1, 0, 1, 0, 1)
        yield (0, 1, 0, 1, 0)

board.source_delay = 1
board.source = alternating()
[...]

PR #160, opened 2016-02-01
Add *args to the LEDBoard on() and off() methods.

lurch:
Implements #125 

As well as doing board.on() to turn on all LEDs, you can now do e.g. board.off(2) to turn off the second LED, or board.off(1,3) to turn off the first and third LEDs; or board.on(2, 4, others_off=True) to turn on the second and 4th LEDs and turn all the other LEDs off.


waveform80:
Nice idea - my one issue is with how the LEDs are numbered (I think I had this discussion with Ben a while ago, but I can't recall if we came to a conclusion). In your code the LEDs are numbered as they are in the `leds` tuple. So, let's say a user's got LEDs on GPIOs 17, 18, 27, 22, 23 and they want to switch between the left three on and the right three:

```python
from gpiozero import LEDBoard
from time import sleep

board = LEDBoard(17, 18, 27, 22, 23)
while True:
    board.on(0, 1, 2, others_off=True)
    sleep(1)
    board.on(2, 3, 4, others_off=True)
    sleep(1)
```

Firstly, I'm not wholly convinced by `others_off`; there'll be little execution time difference (and certainly no human-visible difference) to doing:

```python
from gpiozero import LEDBoard
from time import sleep

board = LEDBoard(17, 18, 27, 22, 23)
while True:
    board.off()
    board.on(0, 1, 2)
    sleep(1)
    board.off()
    board.on(2, 3, 4)
    sleep(1)
```

Which looks simpler and avoids having to remember a new keyword-only parameter. Then there's the issue that the LED numberings are now zero-based indexes (almost akin to BOARD numbering, but off-by-one) while the class was instantiated with BCM numbering. My suspicion is that users might find it simpler to stick with the numbering used on instantiation:

```python
from gpiozero import LEDBoard
from time import sleep

board = LEDBoard(17, 18, 27, 22, 23)
while True:
    board.off()
    board.on(17, 18, 27)
    sleep(1)
    board.off()
    board.on(27, 22, 23)
    sleep(1)
```

Thoughts?

lurch:
When adding the `others_off` parameter, I was thinking of the canonical knight-rider example ;-)
If we stick with index-based-numbering, that can be implemented as:

```python
from gpiozero import LEDBoard
from time import sleep

board = LEDBoard(17, 18, 27, 22, 23)
last_index = len(board.leds) - 1
index = 0
increasing = False
while True:
    board.on(index, others_off=True)
    sleep(0.1)
    if index == 0 or index == last_index:
        increasing = not increasing
    index += 1 if increasing else -1
```

But if you and @bennuttall decide that:
```python
board.off()
board.on(some_pins)
```

is always simpler / more-readable than:
```python
board.on(some_pins, others_off=True)
```

then that's fine with me and I'll remove that optional feature from my PR.


But I'm definitely :-1:  on passing in the GPIO numbers to `on()`/`off()`, rather than passing in array-indices. As is @martinohanlon https://github.com/RPi-Distro/python-gpiozero/issues/125#issuecomment-160479955
(and of course keeping it as indices also means that if the user decides to change which GPIOs their pins are connected to, they only have to edit their code in one place)
And @bennuttall 's example on #125 of:
```python
n = 2
leds = [LED(i) for i in (2, 3, 4)]
[led.on() for i, led in enumerate(leds) if n > i]
```
then becomes:
```python
n = 2
board = LEDBoard(2,3,4)
board.on(*range(n))
```

waveform80:
Knightrider ... ah, the memories ... the hair ... the Hoff! Right, that's enough nostalgia for dodgy 80s hair-dos, and fantastically silly plot lines. I'd probably code it like this, but this might be my obsession with itertools:

```python
from gpiozero import LEDBoard
from itertools import cycle, chain
from time import sleep

pins = [17, 18, 27, 22, 23]
board = LEDBoard(*pins)
for pin in cycle(chain(pins, reversed(pins))):
    board.off()
    board.on(pin)
    sleep(0.1)
```

Which still leaves the pin numbers in one place (in case of re-wiring) and doesn't require me to think about indexes at all (which is largely the point of Python's `for` construct): I simply want to iterate up the list of pins, then down again so chain(pins, reversed(pins)) does that nicely, then I just want to repeat that forever, so throw it in cycle() and off we go!

\<with the sound of a really cheap 80s synth\> di di di dee, di di di dee, di di di ... di ... di ... deeeee.

I'll get my coat.

lurch:
Oooh, very clever. I can see that I'll need to read up on all the fancy functions in itertools at some point...

waveform80:
Well, as I say, that's the way *I'd* code it which is why I think BCM numbering for on/off is a reasonable idea, but I'm getting the notion that this probably isn't the consensus of the masses (and I do have a hard time thinking like a kid new to programming - as that was 30 years ago in my case :). 

Given that the goal of the library is to be intuitive, if my way ain't intuitive then we should either go another way or make it really obvious in the recipes how to use the methods in a simple way for a variety of scenarios (like making knightrider LED displays).

Comments welcome!

martinohanlon:
If you are interacting with a named class I think the complexity should be abstracted away. As a user you have already defined all the pin numbers at initiation, you shouldn't have to worry about them after that, it seems very logical to me that you would use a position on the LEDBoard rather than an arbitrary pin number.

waveform80:
It took me a while to understand your point (I was thinking "named class? Do we have any anonymous ones?") but I think I get it now. There's precious little difference between my example above and simply using a list of LEDs:

```python
from gpiozero import LED
from itertools import cycle, chain
from time import sleep

pins = [17, 18, 27, 22, 23]
leds = [LED(pin) for pin in pins]
[led.off() for led in leds]
for led in cycle(chain(leds, reversed(leds))):
    led.on()
    sleep(0.1)
    led.off()
```

The only thing the LEDBoard class adds in that case is trivial methods for turning everything on and off without an (overt) loop. So, using index position instead of pin number would actually be *functionality* added by the LEDBoard class (to imply the location of each device relative to each other). Okay, it's functionality implied by any list or container (Dave's brain goes off on a tangent considering whether LEDBoard should be a Container derivative implementing `__len__` etc., but is quickly dragged back on topic...), but it's still functionality and it's worth having. Okay, I'm sold, index numbers it is.

lurch:
> Okay, I'm sold, index numbers it is.

Yay.

@martinohanlon Do you agree with @waveform80 that my `others_off` parameter is a feature that'll never be used (because it's 'simpler' to just add an extra call to `board.off()`), and therefore should be removed from this PR?

waveform80:
My major concern (and I think it's one shared by Ben) is API bloat. I don't terribly mind having lots of classes; classes can be categorized and categories separated into different pages keeping the docs reasonably neat and simple. However, add a method or even a method parameter to a base class (Device, InputDevice, or indeed LEDBoard which is the basis for several board classes) and now it'll appear in the docs all the way down the line (well ... assuming you include inherited members in the docs; we do because Ben felt, and I largely agree, that if we didn't include them then the docs for things like the LED class would include ... not very much, which means a kid browsing them now needs to understand inheritance to track down simple things like `on()` and `off()`).

So in this specific case, I'm not firmly against `others_off`, but wherever something's going to add to a base(ish) class, I'm likely to question it if there's an alternative that doesn't add to it. I'm happy to be persuaded of the burning need for it though!

martinohanlon:
I'm with Dave. others_off() seems superfluous.

lurch:
Okay, I'll update and re-push.

bennuttall:
Same here. Index numbers makes more sense than BCM numbers for LEDBoard, and we could do without `others_off`.

Let's merge once you've updated.

lurch:
PR updated.
But I see there's a merge-conflict, which will have to wait until this evening (unless someone else wants to fix it in the meantime).

lurch:
Arrghhh, tried fixing the merge conflicts (accidentally made a mistake), forced-push, realised (and fixed) my mistake, tried fixing things up with 'git rebase', and now it looks like I've got myself into a right muddle! :-(

I think the easiest thing is probably to just close this totally-screwed-up PR and open another one...

PR #161, opened 2016-02-01
Add a blink thread to LEDBoard

### LINKED ISSUE #94: Blink doesn't work for boards when background=False
Is this fixable? Or should we drop `background` in these cases?

lurch:
Fixes #94

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -11,11 +11,12 @@ except ImportError:
 
 from time import sleep
 from collections import namedtuple
+from itertools import repeat
 
 from .exc import InputDeviceError, OutputDeviceError
 from .input_devices import Button
 from .output_devices import LED, PWMLED, Buzzer, Motor
-from .devices import CompositeDevice, SourceMixin
+from .devices import GPIOThread, CompositeDevice, SourceMixin
 
 
 class LEDBoard(SourceMixin, CompositeDevice):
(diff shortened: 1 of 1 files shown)


lurch:
I dunno if this is _quite_ the fix that @waveform80 had in mind, but basically I copied all of the `blink`-related code out of `DigitalOutputDevice` and copied it into `LEDBoard` (adapting it where appropriate) and AFAICT it all **seems** to work okay?

waveform80:
I had an attempt at fixing this in a way that re-used the existing blink method(s) by having the board fire up threads for each blink it ran. But left long enough they do eventually go out of sync (such is life in a non-RTOS). So yeah, this is probably the only reasonable solution.

Unfortunately, now we support PWMLEDs in LEDBoard it does need a little refinement. If PWMLEDs are in use then we should allow `fade_in_time` and `fade_out_time` (and if they're non-zero and PWMLEDs aren't in use then throw an exception).

lurch:
> I had an attempt at fixing this in a way that re-used the existing blink method(s) by having the board fire up threads for each blink it ran. But left long enough they do eventually go out of sync (such is life in a non-RTOS).

Isn't having multiple threads also more inefficient than having a single thread? Although granted, blinking a few LEDs isn't going to use up much of the resources on a 700MHz RPi ;)

> Unfortunately, now we support PWMLEDs in LEDBoard it does need a little refinement.

Ah okay, I'll try to hack this in later...

waveform80:
Yup - multiple threads is definitely more inefficient than a single thread, but as you also note there's plenty of power on a pi for a thread per LED (thread overhead starts getting nasty at hundreds of threads, which is one of several reasons they're considered bad for solving heavily distributed I/O heavy workloads like the C10k problem). Still it'd be premature optimization to avoid them here for that reason.

Did you get a chance to look at the PWMLED bit? If not, don't worry - I'll merge this anyway and press on with 1.1 tonight :)

lurch:
> Did you get a chance to look at the PWMLED bit?

Nope.

> If not, don't worry - I'll merge this anyway and press on with 1.1 tonight :)

Thanks :)

PR #166, opened 2016-02-03
Minor tweak to PWMOutputDevice.is_active

lurch:
...so that its implementation is more consistent with the other is_active methods


bennuttall:
Looks sensible. @waveform80?

waveform80:
Yup, can't remember where @lurch and I discussed this, but it was somewhere ... e-mail maybe. Anyway, I agreed it was a decent idea and the existing implementation was probably born of historical conceptions of `value`.

PR #174, opened 2016-02-08
Fix #135

### LINKED ISSUE #135: RGBLED doesn't have blink() while docs say it does
I was trying out your awesome library and was slightly disapointed that RGBLED class doesn't actually have a blink() method while the docs here say, that it does:

http://pythonhosted.org/gpiozero/outputs/

It's probably cause RGBLED doesn't derive from DigitalOutputDevice like other two LED classes.

waveform80:
Yeah, okay, it's RGB linear interpolation. It's good enough...

### DIFF
--- a/docs/api_output.rst
+++ b/docs/api_output.rst
@@ -35,7 +35,7 @@ RGBLED
 ======
 
 .. autoclass:: RGBLED(red, green, blue, active_high=True)
-    :members: on, off, red, green, blue, value
+    :members: on, off, blink, red, green, blue, value
 
 Motor
 =====
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -302,6 +302,7 @@ class PWMOutputDevice(OutputDevice):
             raise
 
     def close(self):
+        self._stop_blink()
         if self._pwm:
             # Ensure we wipe out the PWM object so that re-runs don't attempt
             # to re-stop the PWM thread (otherwise, the fact that close is
@@ -484,11 +485,12 @@ PWMLED.is_lit = PWMLED.is_active
 
 
 def _led_property(index, doc=None):
-    return property(
-        lambda self: getattr(self._leds[index], 'value'),
-        lambda self, value: setattr(self._leds[index], 'value', value),
-        doc
-    )
+    def getter(self):
+        return self._leds[index].value
+    def setter(self, value):
+        self._stop_blink()
+        self._leds[index].value = value
+    return property(getter, setter, doc=doc)
 
 
 class RGBLED(SourceMixin, CompositeDevice):
@@ -523,6 +525,8 @@ class RGBLED(SourceMixin, CompositeDevice):
         (the :meth:`off` method always does the opposite).
     """
     def __init__(self, red=None, green=None, blue=None, active_high=True):
+        self._leds = ()
+        self._blink_thread = None
         if not all([red, green, blue]):
             raise OutputDeviceError('red, green, and blue pins must be provided')
         super(RGBLED, self).__init__()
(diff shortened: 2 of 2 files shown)


gpiozero/output_devices.py
```diff
@@ -572,9 +576,91 @@ def off(self):
...
+                (lerp(i * (1 / fps) / fade_out_time, False), 1 / fps)
+                for i in range(int(fps * fade_out_time))
+                ]
+        sequence.append((off_color, off_time))
+        sequence = (
+                cycle(sequence) if n is None else
+                chain.from_iterable(repeat(sequence, n))
+                )
+        for value, delay in sequence:
+            self._leds[0].value, self._leds[1].value, self._leds[2].value = value
```
lurch: Could this not simply be `self.value = value` ?

docs/api_output.rst
```diff
@@ -35,7 +35,7 @@ RGBLED
 ======
 
 .. autoclass:: RGBLED(red, green, blue, active_high=True)
-    :members: on, off, red, green, blue, value
+    :members: on, off, blink, red, green, blue, value
```
lurch: If we have `on` and `off`, should we also have `toggle` ?
And should we also have `RGBLED.is_lit` ?
Does the `color` property need to be mentioned here?


Definitely feels like there's an implicit "LED interface" shared between LED, PWMLED and RGBLED... any (easy) way that implicit interface could be formalised, or is this just an example of duck-typing?

gpiozero/output_devices.py
waveform80: Afraid not; that'll implicitly invoke _stop_blink

docs/api_output.rst
waveform80: Hmm, good point - it may be an example of duck typing, but that's not a bad thing. There probably ought to be a toggle. On color/value - color's just an alias for value but looking at the other classes we generally prefer including the "device specific" aliases over the generic names, so yeah I ought to change that too. I'll bung this stuff in a misc commit for a bit later - just ploughing through LEDBarGraph, then it's onto the big pins merge, then I'll do the misc stuff.

gpiozero/output_devices.py
lurch: Hmmm, https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L551 doesn't call `_stop_blink`, but I see that https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L490 does call `_stop_blink`. I haven't tried running the code yet, but from a first glance it looks like it'll only call _stop_blink on the LED instance, and not on the RGBLED? Is there some deeper magic going on here, or does the RGBLED value-setter need a `_stop_blink` too?

And all that said, would `self._leds[0].value, self._leds[1].value, self._leds[2].value = value` re-written as `self.red.value, self.green.value, self.blue.value = value` better match the rest of the 'style' of RGBLED?

Or maybe it's just too early in the day, and I'm not reading the code properly yet ;-)

docs/api_output.rst
lurch: Dunno if there's any 'need' for it, but for the sake of completeness would it be worth adding `initial_value` to `RGBLED.__init__` ?

waveform80: Yeah, that's a decent idea for the sake of similarity

gpiozero/output_devices.py
waveform80: I'm afraid you're not quite reading the code properly (but then it is a bit obscure largely because I'm so lazy!). The setter will call _stop_blink on the RGBLED instance, not the subordinate LEDs. Also, the getter doesn't return the LED itself; it returns the subordinate LED's *value* (this is because RGBLED's concept is not to look like a collection of LEDs a la LEDBoard, but rather to look like a single device which represents a colour, composed of three components - this is also why there's no direct way to get the individual LED components from it; just their values).

lurch: Thanks for the clarifications :-)

> this is because RGBLED's concept is not to look like a collection of LEDs a la LEDBoard, but rather to look like a single device which represents a colour, composed of three components

So (with apologies for going off on a tangent) I guess there's a subtle difference between a CompositeDevice like RGBLED (where we don't explicitly provide access to the 'inner components') and the CompositeDevices like LEDBoard and SnowPi where we do provide access to the collection of 'inner components'?

waveform80: Yeah - it's a rather woolly conceptual difference rather than anything concrete in code, but it's there nonetheless (if only in my head!). LEDBoard and SnowPi represent collections of components and thus it makes sense to present them as such (in fact I'm toying with the idea of literally making LEDBoard a collection so you could iterate over the LEDs within it, but I digress). RGBLED on the other hand is meant to represent a single component - albeit one that's controlled from several pins (and yeah, you could just use three LEDs ... in fact that's exactly what I'm doing on my test rig here because I haven't got an RGBLED ... but that's not really it's intended purpose).

PR #176, opened 2016-02-08
Merge-clean update of PR #126

waveform80:
This is an updated cherry-pick of Martin O'Hanlon's LEDBarGraph class.

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -19,7 +19,59 @@ from .output_devices import LED, PWMLED, Buzzer, Motor
 from .devices import GPIOThread, CompositeDevice, SourceMixin
 
 
-class LEDBoard(SourceMixin, CompositeDevice):
+class LEDCollection(SourceMixin, CompositeDevice):
+    """
+    Abstract base class for :class:`LEDBoard` and :class:`LEDBarGraph`.
+    """
+
+    def __init__(self, *pins, **kwargs):
+        self._blink_thread = None
+        super(LEDCollection, self).__init__()
+        pwm = kwargs.get('pwm', False)
+        LEDClass = PWMLED if pwm else LED
+        self._leds = tuple(LEDClass(pin) for pin in pins)
+
+    def close(self):
+        for led in self.leds:
+            led.close()
+
+    @property
+    def closed(self):
+        return all(led.closed for led in self.leds)
+
+    @property
+    def leds(self):
+        """
+        A tuple of all the :class:`LED` or :class:`PWMLED` objects contained by
+        the instance.
+        """
+        return self._leds
+
+    def on(self):
+        """
+        Turn all the LEDs on.
+        """
+        for led in self.leds:
+            led.on()
+
+    def off(self):
+        """
+        Turn all the LEDs off.
+        """
+        for led in self.leds:
+            led.off()
+
+    def toggle(self):
+        """
+        Toggle all the LEDs. For each LED, if it's on, turn it off; if it's
+        off, turn it on.
+        """
+        for led in self.leds:
+            led.toggle()
+
+
+
+class LEDBoard(LEDCollection):
     """
     Extends :class:`CompositeDevice` and represents a generic LED board or
     collection of LEDs.
@@ -41,29 +93,10 @@ class LEDBoard(SourceMixin, CompositeDevice):
         ``False`` (the default), construct regular :class:`LED` instances. This
         parameter can only be specified as a keyword parameter.
     """
-    def __init__(self, *pins, **kwargs):
-        self._blink_thread = None
-        super(LEDBoard, self).__init__()
-        pwm = kwargs.get('pwm', False)
-        LEDClass = PWMLED if pwm else LED
-        self._leds = tuple(LEDClass(pin) for pin in pins)
 
     def close(self):
         self._stop_blink()
-        for led in self.leds:
-            led.close()
-
-    @property
-    def closed(self):
-        return all(led.closed for led in self.leds)
-
-    @property
-    def leds(self):
-        """
-        A tuple of all the :class:`LED` or :class:`PWMLED` objects contained by
-        the instance.
-        """
-        return self._leds
+        super(LEDBoard, self).close()
 
     @property
     def value(self):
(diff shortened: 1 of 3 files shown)


gpiozero/boards.py
```diff
@@ -178,9 +200,96 @@ def _blink_device(self, on_time, off_time, fade_in_time, fade_out_time, n, fps=5
...
 
 
+class LEDBarGraph(LEDCollection):
+    """
+    Extends :class:`CompositeDevice` to control a line of LEDs representing a
+    bar graph. Positive values (0 to 1) light the LEDs from first to last.
+    Negative values (-1 to 0) light the LEDs from last to first.
+
+    The following example turns on all the LEDs on a board containing 5 LEDs
+    attached to GPIO pins 2 through 6::
```
lurch: Umm, looks like the example only turns on the first two LEDs and the last two LEDs, but not **all** LEDs?

gpiozero/boards.py
```diff
@@ -178,9 +200,96 @@ def _blink_device(self, on_time, off_time, fade_in_time, fade_out_time, n, fps=5
...
+
+        from gpiozero import LEDBarGraph
+
+        graph = LEDBarGraph(2, 3, 4, 5, 6)
+        graph.value = 2/5  # Light the first two LEDs only
+        graph.value = -2/5 # Light the last two LEDs only
+        graph.off()
```
lurch: I guess a few `sleep()`s sprinkled in here wouldn't go amiss?

gpiozero/boards.py
```diff
@@ -178,9 +200,96 @@ def _blink_device(self, on_time, off_time, fade_in_time, fade_out_time, n, fps=5
...
+    supported::
+
+        from gpiozero import LEDBarGraph, MCP3008
+        from signal import pause
+
+        graph = LEDBarGraph(2, 3, 4, 5, 6)
+        pot = MCP3008(channel=0)
+        graph.source = pot.values
+        pause()
+
```
lurch: Would it be worth copying the examples from here into the Recipes page too?

gpiozero/boards.py
```diff
@@ -102,6 +69,61 @@ def toggle(self):
...
+    def off(self):
+        self._stop_blink()
+        super(LEDBoard, self).off()
+
+    def toggle(self):
+        self._stop_blink()
+        super(LEDBoard, self).toggle()
+
     def blink(
```
lurch: I don't have time to experiment with the code right now, but would it be possible to move all the blink-related stuff into `LEDCollection` too? Or is there a reason it's *only* available for LEDBoard and not LEDBarGraph? (my brain's not really in "python mode" today so apologies if that's a stupid question!)

lurch:
I guess http://gpiozero.readthedocs.org/en/latest/api_generic.html needs updating to reflect this PR too?
And should the docstrings for LEDBoard and LEDBarGraph be updated to say that they inherit from `LEDCollection` and not `CompositeDevice`?

gpiozero/boards.py
waveform80: Doh! I'll get that cleaned up :)

gpiozero/boards.py
waveform80: If it was one of the recipes at the front, yeah - I'm a little less bothered about examples in the API docs (want to keep them short and sweet partly so they're easy to read at the command line; if they demonstrate the principle I figure that's enough).

gpiozero/boards.py
waveform80: Yeah, those are probably worth copying up to the recipes page.

gpiozero/boards.py
waveform80: Right now there isn't a reason ... but I didn't want to include blink because in future I do intend to add PWM to the LEDBarGraph class. Right now, both blink states (off and on) have a reasonable representation in the value of the bar graph (0 and 1). However, consider a blink which supports fading (with PWM) - if all the LEDs are on at half strength what value on the chart is that? Basically blink, once PWM is added, will utilise states that aren't valid in the graph so I figured it's better to leave it in the board class and just say "blinking's not a valid use-case for a graph component" (which doesn't seem too unreasonable to me).

waveform80:
Oooh - good point - forgot about the charts there. Okay, several things to tidy up here by the looks of things - I'll make a ticket for PWM in the bar graph and all this stuff can go under there.

gpiozero/boards.py
lurch: Sounds fair enough.
Does the same kind of argument apply to `toggle` or does that still make sense in the bar-chart context?

waveform80: At the moment, I *think* all the transforms that toggle could do (including with a theoretical PWM implementation) produce valid end states in the bar graph class. I'm struggling to think of a proper use-case for it in the context of a graph but given it's a valid transform I'm happy just leaving it in for now.

lurch: I've been trying to work out what mathematical operation LEDBarGraph.toggle() would be equivalent to, and I *think* it's:
```python
import math
self.value -= math.copysign(1,self.value)
```
The above technically converts 0 to -1 (rather than 1), but that's not a problem as for the LEDBarGraph 1 and -1 are indistinguishable.

gpiozero/boards.py
```diff
@@ -178,9 +200,96 @@ def _blink_device(self, on_time, off_time, fade_in_time, fade_out_time, n, fps=5
                 break
 
 
+class LEDBarGraph(LEDCollection):
+    """
+    Extends :class:`CompositeDevice` to control a line of LEDs representing a
+    bar graph. Positive values (0 to 1) light the LEDs from first to last.
+    Negative values (-1 to 0) light the LEDs from last to first.
```
lurch: Should this be changed to "Negative values (0 to -1) ..." to reflect that increasingly-larger negative values light up more LEDS "from last to first" ?

lurch:
I hope you don't think I'm being overly picky with all my comments?

I guess there's no reason that LEDBoard shouldn't have it's own `__init__` which allows you to pass in an appropriately-sized tuple as an `initial_value` kwarg?
Giving LEDBoard it's own `__init__` method would also allow you to move `self._blink_thread = None` out of LEDCollection's `__init__` which IMHO currently seems a bit of a misnomer (and is partly why I asked about https://github.com/RPi-Distro/python-gpiozero/pull/176#discussion_r52292334).


Hmmm, and it occurs to me that maybe the implementation of `initial_value` inside LEDBarGraph isn't quite correct? Given that the reason for adding initial_value parameters was so that GPIO-pins could have their initial-value set at "setup" time to avoid any startup-glitches, I guess what the current LEDBarGraph `__init__` is actually doing is creating all the individual LEDs set to off, and then (almost immediately) setting them to the 'desired' initial-value.
Perhaps LEDCollection itself should accept an optional `initial_value` tuple, which it then passes into the individual LED constructors as appropriate, and then the LEDBarGraph `__init__` could be rewritten to convert **its** initial_value parameter into an appropriate tuple-value, before calling the parent-class init-method? I hope that makes sense!

waveform80:
Not at all! That does actually sound like a much cleaner way of initialising everything and avoids any "glitch" at the beginning. The major reason for including an initial value in LEDBarGraph is to ensure that after construction the device is in a "known good" state (we don't have some weird set of led states like on/off/on/off/on which won't mean anything).

Anyway, I'll have a look at this when I get back home!

lurch:
> The major reason for including an initial value in LEDBarGraph is to ensure that after construction the device is in a "known good" state (we don't have some weird set of led states like on/off/on/off/on which won't mean anything).

In that case perhaps the LEDBarGraph constructor should explicitly disallow a `None` value? (which is allowed by e.g. LED)

gpiozero/boards.py
waveform80: Yeah, that's a good idea

waveform80:
Currently, it effectively does since setting value to `None` will fail with an exception (as `None` is not comparable to a float/int) - that'll probably be the case when I re-jig the implementation to construct the initial-tuple too

PR #177, opened 2016-02-08
Miscellaneous minor changes for 1.1

waveform80:


lurch:
Looks good. I still think RGBLED.is_lit would be useful for 'uniformity' purposes :)
(as mentioned earlier)

waveform80:
Doh, good point - I'll do that in a sec.

PR #183, opened 2016-02-10
Add *args to the LEDBoard on(), off() and toggle() methods.

lurch:
Implements #125

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -173,17 +173,29 @@ class LEDBoard(LEDCollection):
         self._stop_blink()
         super(LEDBoard, self).close()
 
-    def on(self):
+    def on(self, *args):
         self._stop_blink()
-        super(LEDBoard, self).on()
+        if args:
+            for index in args:
+                self[index].on()
+        else:
+            super(LEDBoard, self).on()
 
-    def off(self):
+    def off(self, *args):
         self._stop_blink()
-        super(LEDBoard, self).off()
+        if args:
+            for index in args:
+                self[index].off()
+        else:
+            super(LEDBoard, self).off()
 
-    def toggle(self):
+    def toggle(self, *args):
         self._stop_blink()
-        super(LEDBoard, self).toggle()
+        if args:
+            for index in args:
+                self[index].toggle()
+        else:
+            super(LEDBoard, self).toggle()
 
     def blink(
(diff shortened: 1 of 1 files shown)


lurch:
For background to this, see the discussion in #160.

When updating this to work with @waveform80 's new LEDCollection parent-class, I wasn't sure whether to:
a) add the `*args` parameter to LEDCollection.on() and LEDCollection.off(), and then add new LEDBarGraph.on() and LEDBarGraph.off() methods which **don't** accept `*args`
or
b) leave the base LEDCollection methods as-is and just add the `*args` parameter to LEDBoard.on() and LEDBoard.off()
so I went with option b. And while I was at it, it seemed to "make sense" to add a similar `*args` to the LEDBoard.toggle() method too.

bennuttall:
Seems ok. @waveform80 are you happy with this?

waveform80:
Unfortunately this isn't quite going to work. Now that LEDBoard is allowed to contain both LEDs and other LEDBoards (i.e. it can be hierarchical), the `leds` property is a generator which recursively flattens the hierarchy to yield each led within it (and generators can't be indexed like this).

So, there's a couple of ways to go about this but it's going to depend on exactly what you want the *args to do. If you want to args to refer to the indexes of LEDs in the flattened iterator, just do something like:

```python
if args:
    for index in args:
        tuple(self.leds)[index].on()
```

Alternatively, if you want args to refer to the indexes of the LEDs *or LEDBoards* contained in this particular LEDBoard then simply index the LEDBoard itself:

```python
if args:
    for index in args:
        self[index].on()
```

Have a look at `SnowPi` for an example of a hierarchical entity.

lurch:
I think the latter suggestion makes more sense. If you want the granularity of controlling individual LEDs within a sub-LEDBoard, then you can just use the on() method on that sub-LEDBoard.
Whereas with the former approach there's no easy way to say (with a single function-call) "turn on all the LEDs just on the right and left arms".

Any comments from @martinohanlon or @bennuttall before I update this PR?

waveform80:
Well, there's snowman.arms.on() but anyway, I take your point ;) There's no easy way to specify a subset of one of the devices that isn't a device itself (like a pattern of all the odd LEDs in a strip for example).

lurch:
PR has been updated to directly index `self`, rather than `self.leds`

waveform80:
Lovely stuff - thanks!

Issue #51, opened 2015-10-09
Add circuit diagrams to docs


bennuttall:
I have some @samalder diagrams I can add, others will need creating especially (initially with fritzing is fine but Sam's are much better).

bennuttall:
First few have been added!

bennuttall:
Can we list missing diagrams here? (fritzling diagrams for Sam to redraw would be ace)

waveform80:
I've been leaving IMAGE TBD comments in docs/recipes.rst for "images to be done"; when I've finished a bit more code I'll see if I can whip up some fritzing diagrams for them.

waveform80:
Here's some more diagrams - do check I've got 'em right! I've include Fritzing diagram source for ease of modification in case I've got anything wrong.

For anyone that wants to modify these: work on the Fritzing diagram, save it, then export as SVG. Then run "make doc" in an appropriately configured venv (`make develop`). This'll use Inkscape to generate the PNG and PDF output (because it does this much better than Fritzing). Commit the Fritzing source, and all the image outputs (SVG, PNG, PDF); the reason for all the formats is that the SVG is the "canonical image" but Sphinx will use PNG for web output and PDF/EPS for PDF/Epub output (this is the reason the reST source uses `.*` extensions for images; it lets the Sphinx pick the most appropriate image format for the requested output).

bennuttall:
Just a suggestion to add to this:

I think we should document wiring diagrams separately from docstrings, on their own separate docs page, rather than sometimes in a docstring. Every device class (except HATs and things) should have a wiring diagram with useful notes (e.g. pull-up/pull-down) included. Some recipes have wiring diagrams, and I guess that's fine when multiple devices are used.

@waveform80 @lurch any thoughts? If agreed, I'll get started on it after the v1.5 release, for inclusion in the v1.6 release.

lurch:
...presumably with the class-docs linking to the wiring-docs (where appropriate), and the wiring-docs linking back to the class-docs?

bennuttall:
Exactly

PR #194, opened 2016-02-13
Work on #51

waveform80:
Some more circuit diagrams; better check these are accurate - I haven't had time to wire them up and test them!


lurch:
I haven't checked the diagrams in depth, but a few small notes from a quick look at the images:

### motion_sensor
I believe the intention here is for PIR motion sensors, like the one shown in https://github.com/RPi-Distro/python-gpiozero/blob/master/docs/images/motion-sensor.png (and http://thepihut.com/collections/camjam-edukit/products/camjam-edukit-2-sensors )

### motor
Bit hard to see what the name of the chip is - perhaps the upper blue wire could be moved?

### rgb_led
While the diagram you've shown will work (and is fine for quick prototype purposes), I believe conventionally these are wired up with one resistor per colour, rather than just one resistor on the cathode - otherwise e.g. switching the blue part of the LED on and off will change the brightness of the red part of the LED. (and in more advanced applications, having a resistor-per-colour allows you to try 'balancing' the resistors, so that turning all the LEDs on results in 'white' and not e.g. pink - but this latter part is probably outside the scope of GPIO Zero)

waveform80:
On the motion sensor: yup, and if there was one like that in Fritzing I'd have used it - unfortunately the one that's there is the only IR sensor Fritzing had (and yeah, it's an "infrared distance sensor" - oh well, at least it's got vaguely similar pins to the CamJam one, except the OUT and GND are swapped).

On the motor: good point - I'll push that in a mo. It's the SN754410 (same as used on the Ryanteck board).

On the RGB LED: ahh, there goes Dave showing his lack of EE experience again. I'll fix it in a mo.

PR #196, opened 2016-02-13
Work on #51 - add pot diagram

waveform80:
Hope I've got this right!


lurch:
Should the breadboard be vertical, so that it matches all the other diagrams?

waveform80:
I would've liked to but I couldn't get the wiring neat enough to be understandable in that configuration - too many things crossing over. I'm half tempted to do the same with the Motor one, because I'll certainly have to do something similar for the Robot one (given motors pop off both sides of the dual H-bridge IC).

Issue #182, opened 2016-02-09
LEDBarGraph PWM support


waveform80:
As mentioned in comments in #176 there's several things to clean up:

* [x] Add PWM support to LEDBarGraph
* [x] Document LEDCollection in generic API docs (don't forget the hierarchy diagram)
* [x] Add pot->graph recipe to Recipes chapter
* [x] Fix the description of the first example in the docs

bennuttall:
What's your plan for PWM support? A value of 0.5 for 3 LEDs = `[1, 0.5, 0]`?

waveform80:
Yeah, that's basically the idea - every LED is either on (value > LED's threshold), off (value < prior LED's threshold) or on a PWM duty-cycle linearly proportional to how far value sits between the current and prior LED's thresholds.

bennuttall:
Cool. I might have a go.

waveform80:
Go for it :)

bennuttall:
Ok got distracted. I'm sure I can tackle this - not now though.

PR #198, opened 2016-02-14
Add PWM support to LEDBarGraph and PiLiterBarGraph, re: #182

bennuttall:
The start of fixing #182. I've implemented PWM on `LEDBarGraph` and extended it to `PiLiterBarGraph`. I think it's done reasonably sensibly but @waveform80 can be the judge of that.

I wish the issue had been related to `blink()`. Could have been blink #182.

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -265,7 +265,7 @@ class LEDBarGraph(LEDCollection):
     """
 
     def __init__(self, *pins, **kwargs):
-        super(LEDBarGraph, self).__init__(*pins, pwm=False)
+        super(LEDBarGraph, self).__init__(*pins, **kwargs)
         initial_value = kwargs.get('initial_value', 0)
         self.value = initial_value
 
@@ -307,12 +307,19 @@ class LEDBarGraph(LEDCollection):
     @value.setter
     def value(self, value):
         count = len(self.leds)
-        if value >= 0:
-            for index, led in enumerate(self.leds, start=1):
-                led.value = value >= (index / count)
-        else:
-            for index, led in enumerate(reversed(self.leds), start=1):
-                led.value = value <= -(index / count)
+        ratio = abs(value * count)
+        full_leds = ratio // 1
+        remainder = ratio % 1
+        leds = self.leds
+        if value < 0:
+            leds = reversed(leds)
+        for index, led in enumerate(leds):
+            if index < full_leds:
+                led.value = 1
+            elif index == full_leds:
+                led.value = remainder
+            else:
+                led.value = 0
 
 
 class PiLiter(LEDBoard):
@@ -360,9 +367,10 @@ class PiLiterBarGraph(LEDBarGraph):
 
     .. _Ciseco Pi-LITEr: http://shop.ciseco.co.uk/pi-liter-8-led-strip-for-the-raspberry-pi/
     """
-    def __init__(self, initial_value=0):
+    def __init__(self, initial_value=0, pwm=False):
         super(PiLiterBarGraph, self).__init__(
-                4, 17, 27, 18, 22, 23, 24, 25, initial_value=initial_value)
+                4, 17, 27, 18, 22, 23, 24, 25,
+                initial_value=initial_value, pwm=pwm)
 
 
 TrafficLightTuple = namedtuple('TrafficLightTuple', ('red', 'amber', 'green'))


bennuttall:
Oh just realised I forgot to change the `.value` getter.

Any reason that doesn't just return a the average LED count (or negative if it starts from the other end)?

bennuttall:
In fact why does it need to query the LEDs? Can't we just store the value as a regular property with a setter?

gpiozero/boards.py
```diff
@@ -265,7 +265,7 @@ class LEDBarGraph(LEDCollection):
     """
 
     def __init__(self, *pins, **kwargs):
-        super(LEDBarGraph, self).__init__(*pins, pwm=False)
+        super(LEDBarGraph, self).__init__(*pins, **kwargs)
         initial_value = kwargs.get('initial_value', 0)
```
lurch: If `initial_value` is being handled separately, does it need to be removed from the `**kwargs` passed to the base class?

bennuttall: I don't think so

waveform80: Unfortunately, while `*pins, pwm=False` is valid in Python 3, it isn't in Python 2 (only `**kwargs` can come after `*args` in Python 2 - nothing else, including keyword args with defaults). This is the sole reason we use **kwargs in such cases; to maintain Python 2 compatibility (while it's perfectly reasonable to use Python 3 only for applications at this point, libraries probably need to support Python 2 for several more years - there's still plenty of Python 2 users out there!)

waveform80:
Why not store the value as an internal variable? Good question. The cryptic answer is: "Confucius say: man with one watch knows the time; man with two watches is never sure...". Or to put it another way a single source of truth is always preferable.

We're defining an LEDBarGraph class: what is the state of the graph? Is it what is currently shown on the LEDs, or is it what we last set and stored inside the class instance? I'd argue the value of the graph is whatever is displayed on the LEDs *even if an external controller has modified it* because that's what the LEDBarGraph class is meant to represent: the LEDs as a graph. Hence, if we have the ability to read the LEDs we should use it to provide the state of the graph, rather than duplicating it in a variable which by some circumstance may not agree with reality.

* Is a variable simpler? Yes.
* Is a variable quicker? Most certainly.
* Will a variable represent reality? Not necessarily.
* Is the current code as strict as it should be? No (we should check all LEDs beyond the one found not lit are also off and throw a ValueError if not; but throwing an exception for merely reading a property is sufficiently odd behaviour that I'm not necessarily willing to subject inexperienced users to it :)

Still, the fact remains: the purpose of the class is to represent a value on a set of LEDs. It sets that value on the LEDs; thus (IMO) it should read that value back from the LEDs ensuring a "single source of the truth". Am I being too strict?

gpiozero/boards.py
bennuttall: Ah ok, we can just swap that out for **kwargs, right?

bennuttall:
No, that's fine. Well justified.

Back to by first point, any reason value doesn't just return a the average LED count (or negative if it starts from the other end)?

lurch:
What do you mean by "the average LED count" ?

pcopa:
Just some fruit for thought.
I was looking to use PWM for LED strips lighting and noticed that all new
strips are comprised of four colors. Red green blue and white. Just to be
future-proof I would like to suggest considering four channels instead of
the existing three channels.

bennuttall:
> What do you mean by "the average LED count" ?

As in:

```python
sum(led.value for led in self.leds) / len(self.leds)
```

bennuttall:
> Just to be future-proof I would like to suggest considering four channels instead of the existing three channels.

You mean for `RGBLED`?

lurch:
> > Just to be future-proof I would like to suggest considering four channels instead of the existing three channels.

> You mean for `RGBLED`?

I guess if that _is_ what he means, the best solution would probably be a new `RGBWLED` class?

bennuttall:
> I guess if that is what he means, the best solution would probably be a new RGBWLED class?

Agreed. Sounds like witchcraft to me though.

lurch:
> > What do you mean by "the average LED count" ?

> As in: `sum(led.value for led in self.leds) / len(self.leds)`

I guess the problem with that is you _still_ have to do additional checks to see if the LEDBarGraph is 'counting up' or 'counting down', so probably just easier to stick with the current code? And I guess it also ties into @waveform80 's "Is the current code as strict as it should be?" point - if the user has manually set e.g. the first and last LEDs to 'on' and the middle LED to 'off' (by "abusing" the `self.leds` property) then the current code and your proposed code would give very different results.

gpiozero/boards.py
waveform80: Yeah; it's also worth noting that while we use `**kwargs` here we actually document it as `*pins, pwm=False` - so we tell a white lie in the documentation for the sake of clarity.

lurch: In that case, should it be documented as `*pins, pwm=False, initial_value=0` ? (with the obvious caveat that `pwm` and `initial_value` can only be provided by name, and not by position)

waveform80: Sorry, I meant `*pins, initial_value=0` - I was thinking of LEDBoard (which has `pwm=False`, but which I see I need to add `active_high` and `initial_value` to).

On the caveat of keyword only parameters - I'm not sure we have to explain that. Any parameters after an *args style parameter are implicitly keyword only (as [PEP-3102](https://www.python.org/dev/peps/pep-3102/), which introduced keyword only parameters, specifies) because if they weren't keywords the *args would suck them up. But then given this is Python 3 only functionality, I could well believe there's plenty of Python 2 users out there that this wouldn't be immediately obvious to. Could be worth adding something like a "(keyword-only)" tag to such parameters in the docs...

waveform80:
Fixed in #241

PR #201, opened 2016-02-17
Add .cache directory to .gitignore

lurch:
Seems to get created during `make test`


waveform80:
Interesting, I hadn't noticed that. Evidently I need to relax some of the timing tests too - Travis' infrastructure is showing a lot more variance than my Pi at home (not surprising for virtualized stuff really).

PR #205, opened 2016-02-18
Documentation fix

lurch:
To stop http://gpiozero.readthedocs.org/en/latest/api_pins.html#abstract-pin saying "class gpiozero.pins.native.Pin"
(which AFAICT is wrong)

### DIFF
--- a/docs/api_pins.rst
+++ b/docs/api_pins.rst
@@ -105,6 +105,8 @@ NativePin
 Abstract Pin
 ============
 
+.. currentmodule:: gpiozero.pins
+
 .. autoclass:: Pin
     :members:
 


lurch:
I'm not familiar with how the docs work, so this is just my best-guess at the correct fix! ;-)

waveform80:
Doh - well caught! You're absolutely right with your fix as well. The reason it strangely comes up with gpiozero.pins.native is that the "current module" has been set to gpiozero.pins.native just prior to that, and naturally that module imports "Pin" (the abstract base class) in order to derive from it to define NativePin. Hence when sphinx sees "autoclass:: Pin" it first searches the "current module" for Pin ... and finds it (because it's been imported).

PR #207, opened 2016-02-19
Add DistanceSensor.within_range

lurch:
...as an alias of `is_active` (similar to LightSensor.light_detected)

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -754,6 +754,7 @@ class DistanceSensor(SmoothedInputDevice):
             # wrong (XXX raise a warning?)
             return 1.0
 
+DistanceSensor.within_range = DistanceSensor.is_active
 DistanceSensor.when_out_of_range = DistanceSensor.when_activated
 DistanceSensor.when_in_range = DistanceSensor.when_deactivated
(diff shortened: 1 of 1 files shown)


lurch:
Ooops! Just noticed that the code currently has
`DistanceSensor.when_in_range = DistanceSensor.when_deactivated`
which I guess means my PR should _actually_ be:
`DistanceSensor.within_range = not DistanceSensor.is_active` ?
I don't have a distance-sensor to test with.

lurch:
EDIT3: Doh, `is_active` is a property rather than a function, so perhaps `DistanceSensor.within_range = not DistanceSensor.is_active` *is* all that's needed? Need someone who has a distance-sensor to test for me...

lurch:
Looks like (a slight variation of) this was added as part of https://github.com/RPi-Distro/python-gpiozero/commit/23a63697b61487c3182c08077d2542a0e5683e6a

waveform80:
Ah, knew I'd forgotten to close something! Yeah, I'm afraid the distance sensor is a little odd in that its notion of within range is the inverse of active (a consequence of the test being greater-than-or-equal to though now I come to think of it I could've just changed that test in the class ... oh well, this maintains a little more backward compatibility in case anyone's actually testing is_active).

PR #211, opened 2016-02-20
RPiGPIOPin fixups

lurch:
* Imported missing exceptions
* Corrected wrong PinInvalidValue exception
* GPIO.PWM is a class (GPIO.HARD_PWM is the constant)
* GPIO.setup only supports setting input and output modes https://sourceforge.net/p/raspberry-gpio-python/code/ci/default/tree/source/py_gpio.c#l264

### DIFF
--- a/gpiozero/pins/rpigpio.py
+++ b/gpiozero/pins/rpigpio.py
@@ -13,6 +13,9 @@ from ..exc import (
     PinInvalidFunction,
     PinSetInput,
     PinFixedPull,
+    PinInvalidPull,
+    PinInvalidState,
+    PinPWMFixedValue,
     )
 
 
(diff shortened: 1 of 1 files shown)


waveform80:
Damned timing tests ... I must get around to fixing those. Unfortunately no time this weekend!

bennuttall:
No time to fix time

waveform80:
Oh the irony :) Yeah, unfortunately I've got some other contract work on this weekend and I know that fixing the timing tests is going to require a fair bit of experimentation (and time)!

lurch:
I've just written a fresh install of 2016-02-09-raspbian-jessie to an SD card and tried running all the unit tests (from before you added the pin stuff) on a Model A+, up until now I'd been doing all my dev and testing on a Pi2 running an older version of wheezy.

In both the py27 venv and the py34 venv, I always get 4 or 5 tests failing with `assert isclose(...` i.e. the same as the timing errors on TravisCI.

So when you do get time to investigate fixing the timing tests (after you've eaten Ben's thyme), I guess that's an easy to reproduce test-platform.

PR #213, opened 2016-02-20
NativePin fixups

lurch:
* rename FUNCTION / PULL / EDGES dictionaries so that they better-match the other `Pin` implementations
* throw `PinSetInput` if trying to set the state of an input pin

### DIFF
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -22,6 +22,7 @@ from ..exc import (
     PinInvalidEdges,
     PinInvalidFunction,
     PinFixedPull,
+    PinSetInput,
     )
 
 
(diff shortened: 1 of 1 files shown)


waveform80:
Huh ... wonder why Travis isn't doing it's thing. Oh well, all looks good to me :)

lurch:
> Huh ... wonder why Travis isn't doing it's thing.

I wondered that too - perhaps it's getting confused because this is a branch named `patch-3` which is a branch-name I'd already used for a previous PR? If I get time I'll have a look through the Travis issue-tracker.

waveform80:
Hmm, could be - don't think I've repeated a branch name yet (I tend to use relatively descriptive names). Can you remember if you pruned your origin's branch names prior to pushing the new `patch-3` branch? That might be another factor.

lurch:
`if you pruned your origin's branch names`

Um, huh? ;-)
Any branches of mine that are named `patch-X` will be ones that github created automatically when I clicked on the Edit-this-file pencil-icon in the github.com webinterface. And when GH tells me that "Pull request successfully merged and closed" I then click on the "Delete branch" button in the webinterface.

waveform80:
Ah, sorry - different workflow to me - I typically start a new branch on my workstation, hack away on it, pull it onto the test pi and try it out, tweak it more, rebase it onto the pi, then push it to GH when I'm happy with it. After I've merged it, I delete the branch on GH, then frequently forget to delete them on the workstation or Pi (until I remember to run "git remote prune origin" at some point and see a load of stuff removed :).

waveform80:
Must admit I did wonder why everyone kept using `patch-#` for branch names ... thought it was just a standard everyone was sticking to!

lurch:
> Must admit I did wonder why everyone kept using patch-# for branch names ... thought it was just a standard everyone was sticking to!

LOL!  Humans are never that predictable, only automated systems are ;-)

lurch:
Not that it really matters now, but I finally heard back from the Travis support team:

"I looked at the logs associated, and found that GitHub returned a 404 when we tried to fetch the information about the PR. It is not clear to me why. There was no subsequent pushes to the PR, so we never tried building that PR again.

I do not think the branch name reuse by GitHub, while it does happen, had anything to do with it, for the URL we tried was https://api.github.com/repos/RPi-Distro/python-gpiozero/pulls/213, which does not have any reference to the branch name.

If this happens again in the future, you could try another push to the branch in question."

So it sounds like it was merely a temporary error, rather than anything more fundamental.

PR #214, opened 2016-02-20
MockPin fixups

lurch:
* alter the PinSetInput exception message to match other `Pin` implementations
* constrain the state of MockPin to a `bool`, and the state of MockPWMPin to a `float`
* allow MockPin to have a `None` frequency set, matching the docs http://gpiozero.readthedocs.org/en/latest/api_pins.html#gpiozero.pins.Pin.frequency

### DIFF
--- a/gpiozero/pins/mock.py
+++ b/gpiozero/pins/mock.py
@@ -64,9 +64,10 @@ class MockPin(Pin):
 
     def _set_state(self, value):
         if self._function == 'input':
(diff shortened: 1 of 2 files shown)


waveform80:
Lovely stuff ... I'm tempted to factor out all the repeated writing stuff in the two `_set_state` into something like a `_change_state` method or something, but anyway - this is all great stuff and worth merging.

PR #216, opened 2016-02-20
Make sure closing a MockPWMPin sets its frequency to None

lurch:
just like all the other Pin classes

### DIFF
--- a/gpiozero/pins/mock.py
+++ b/gpiozero/pins/mock.py
@@ -161,6 +161,10 @@ class MockPWMPin(MockPin):
         super(MockPWMPin, self).__init__(number)
         self._frequency = None
 
+    def close(self):
+        self.frequency = None
+        super(MockPWMPin, self).close()
+
     def _set_state(self, value):
         if self._function == 'input':
             raise PinSetInput('cannot set state of pin %r' % self)


waveform80:
Oooh - interesting! You've got an actual test failure there and it's a really intriguing one: pypy's picked up an issue in `MockPWMPin` -- when you set the frequency to `None` on `MockPWMPin.close`, it attempts to set `state` to `False` but if the pin's `function` is `'input'` that amounts to setting an input pin's state which raises an error.

I've no idea why only pypy notices this (and I'd love to dig into that if only I had the time!) but it's definitely an issue.

lurch:
> Oooh - interesting! You've got an actual test failure there

Hah, I just assumed it was another timing-failure and so didn't bother looking into the build log - whoops! ;-)

> I've no idea why only pypy notices this...

It appears to only happen during interpreter shutdown, so possibly pypy shuts things down in a slightly different order?

The test is still easy to reproduce in 'normal' python though:
```python
>>> from gpiozero.pins.mock import MockPWMPin
>>> pin = MockPWMPin(3)
>>> pin.close()
  File "gpiozero/pins/mock.py", line 165, in close
    self.frequency = None
  File "gpiozero/pins/__init__.py", line 175, in <lambda>
    lambda self, value: self._set_frequency(value),
  File "gpiozero/pins/mock.py", line 188, in _set_frequency
    self.state = False
  File "gpiozero/pins/__init__.py", line 131, in <lambda>
    lambda self, value: self._set_state(value),
  File "gpiozero/pins/mock.py", line 170, in _set_state
    raise PinSetInput('cannot set state of pin %r' % self)
gpiozero.exc.PinSetInput: cannot set state of pin MOCK3
```

I'll add a unit-test, attempt to fix the base MockPWMPin implementation, and then update this PR.

waveform80:
Yup - it's definitely an interpreter shutdown thing (which are usually horribly difficult to track down, but thankfully, as you note, it's easy to test for in this case :). If I had the time, I'd love to dig into pypy to figure out the base cause, but for now I'll settle for just having pypy tests - always nice to have an alternate implementation pick up stuff like this!

lurch:
Now incorporated into #217

PR #217, opened 2016-02-20
MockPin improvements

### LINKED ISSUE #206: MockPin 'breaks' the Pin interface?
According to the "Abstract Pin" [documentation](http://gpiozero.readthedocs.org/en/latest/api_pins.html#abstract-pin) 

> Warning: Descendents must ensure that pin instances representing the same physical hardware are identical, right down to object identity.

and `MockPin` is a descendant of `Pin`. I realise the MockPin doesn't refer to _actual_ hardware, but it might be nice (for the sake of the unit-tests) if
```python
from gpiozero.pins.mock import MockPin
pin1 = MockPin(3)
pin2 = MockPin(3)
print(pin1 == pin2)
```
printed `True`, as it does for all the other `Pin` subclasses (with the current MockPin implementation it prints `False`).
In turn, this would allow us to have unit-tests for checking that e.g.
```python
from gpiozero.pins.mock import MockPin
from gpiozero import LED
led1 = LED(MockPin(3))
pin2 = LED(MockPin(3))
```
fails with `GPIOPinInUse` as all the other `Pin` subclasses do.

I tried having a go myself at modifying the MockPin class to behave like the other Pin implementations (by changing `__init__` to `__new__` , adding a `_PINS` dict, etc.) but that just caused a whole load of problems with `MockPWMPin` then not working.

Feel free to close this issue, if it's irrelevant and I've misunderstood things!

lurch:
Change MockPin (and MockPWMPin) to make them behave more like 'real' pins - fixes #206
Add new MockPin tests, and rework some of the existing ones
Incorporate #216

### DIFF
--- a/gpiozero/pins/mock.py
+++ b/gpiozero/pins/mock.py
@@ -25,18 +25,32 @@ class MockPin(Pin):
     A mock pin used primarily for testing. This class does *not* support PWM.
     """
 
-    def __init__(self, number):
+    _PINS = {}
+
+    @classmethod
+    def clear_pins(cls):
+        cls._PINS.clear()
+
+    def __new__(cls, number):
         if not (0 <= number < 54):
             raise ValueError('invalid pin %d specified (must be 0..53)' % number)
-        self._number = number
-        self._function = 'input'
-        self._state = False
-        self._pull = 'floating'
-        self._bounce = None
-        self._edges = 'both'
-        self._when_changed = None
-        self._last_change = time()
-        self.states = [PinState(0.0, False)]
+        try:
+            old_pin = cls._PINS[number]
+        except KeyError:
+            self = super(Pin, cls).__new__(cls)
+            cls._PINS[number] = self
+            self._number = number
+            self._function = 'input'
+            self._state = False
+            self._pull = 'floating'
+            self._bounce = None
+            self._edges = 'both'
+            self._when_changed = None
+            self.clear_states()
+            return self
+        if old_pin.__class__ != cls:
+            raise ValueError('pin %d is already in use as a %s' % (number, old_pin.__class__))
+        return old_pin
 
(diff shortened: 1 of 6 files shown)


lurch:
"The Travis CI build failed" - oh arse!  I'd only been testing with Python2.7 and Python3.2  :-(

lurch:
So, if I change https://github.com/RPi-Distro/python-gpiozero/pull/217/files#diff-cb8d4fe36ad4392bcf34acd42c33a0e9R172 to `super(MockPWMPin, self).__init__()` then I guess it'd stop Python3.3+ complaining; and surprisingly the tests all still pass in my Python2.7 and Python3.2 venvs with that change. Out of curiosity, how come 2.7 and 3.2 worked okay?

Is the error from Python3.3+ because neither `MockPin` or `Pin` have an `__init__` method of their own, and so `super(MockPWMPin, self).__init__` ends up calling `object.__init__` as the parent-class of `Pin`?

lurch:
I found the answer here: http://stackoverflow.com/questions/14300153/why-does-this-datetime-subclass-fail-on-python-3/  (my guess about `object.__init__` was correct; and this was behaviour only enforced in Py3.3 - the old behaviour being allowed in Py2.7 and Py3.2 was a bug).

PR updated, and now the Travis builds all pass! :-)

waveform80:
Hehe - and that's why I love having a test suite; I can't describe how nerve-wracking it was doing release 1.0 and 1.1 of gpiozero without any test suite in place! Anyway - this now looks excellent - the factoring out of `_change_state` is particularly nice with the return value determining whether anything *actually* changed state. Many thanks!

lurch:
> the factoring out of _change_state is particularly nice

Yeah, I thought so too :)  Thanks for merging!

PR #218, opened 2016-02-21
More small MockPin tweaks

lurch:


lurch:
In the coverage-output we now have:
```
Name                         Stmts   Miss  Cover   Missing
----------------------------------------------------------
gpiozero/pins/mock.py          132      1    99%   119
```
Awesome! The only part of MockPin that's *not* covered is
```python
    def _set_bounce(self, value):
        # XXX Need to implement this
        self._bounce = value
```
;-)
But of course that doesn't mean that every combination of functionality has been tested...

waveform80:
Nice!

waveform80:
> But of course that doesn't mean that every combination of functionality has been tested...

Indeed - we've got branch testing on in `coverage.cfg` but that's still not "complete" coverage testing. At some point I should have a play with the Python port of American Fuzzy Lop...

lurch:
Something I've been wondering about the coverage-stats (and never having looked at coverage before, this may well already be a solved problem) - given that gpio-zero covers both Python2 and Python3, there are obviously some (small) parts of the code that only get covered by one or the other. Is it possible to "combine" the coverage-stats from Py2 and Py3 to give a 'truer' picture?

waveform80:
Yes - I use codecov.io for this on other projects (e.g. [picraft](https://codecov.io/github/waveform80/picraft)). It's not entirely perfect (haven't read up on how to ignore things like `__repr__` as we do with coverage at the moment), but their [GitHub integration](https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf) is fantastic (there's also a [Firefox extension](https://addons.mozilla.org/en-US/firefox/addon/codecov-extension/?src=search)). I think I'd need to get @bennuttall to go sign in to codecov and activate the project, then I can add the necessary bits to the Travis config (codecov's stuff runs as a postscript to Travis).

PR #220, opened 2016-02-21
Add "real" pins tests

waveform80:
This is just a quicky for people to start playing with - it's not complete in any way, shape, or form. This is how I envisage the "real" pin tests being done; part of the test suite with a `skipif` to ensure they don't get run on non-Pi platforms, with a fixture to loop over whatever pin implementations are found (we can't always assume all of them: for example, RPIO doesn't work on a Pi 2), and a relatively simple
wiring for the test.

In this case I've assumed GPIOs 22 and 27 are wired together. They're next to each other, so a jumper is sufficient to run the test.

PRs extending the coverage are very welcome (I've already discovered and fixed several silly bugs in NativePin!). I've left all the interesting hard stuff for people to play with (PWM testing: statistical sampling? debounce compensation testing: timing?). When I've got a second, I'll looking into hooking up my Pi Zero as a Travis-esque test-bed for this, triggered by GitHub webhooks (not sure how I'll deal with reporting yet).

### DIFF
--- a/tests/test_real_pins.py
+++ b/tests/test_real_pins.py
@@ -1,0 +1,143 @@
+from __future__ import (
+    unicode_literals,
+    absolute_import,
+    print_function,
+    division,
+    )
+str = type('')
+
+
+import io
+import subprocess
+
+import pytest
+
+from gpiozero import PinFixedPull, PinInvalidPull, PinInvalidFunction
+
+
+# This module assumes you've wired the following GPIO pins together
+TEST_PIN = 22
+INPUT_PIN = 27
+
+
+# Skip the entire module if we're not on a Pi
+def is_a_pi():
+    with io.open('/proc/cpuinfo', 'r') as cpuinfo:
+        for line in cpuinfo:
+            if line.startswith('Hardware'):
+                hardware, colon, soc = line.strip().split(None, 2)
+                return soc in ('BCM2708', 'BCM2709', 'BCM2835', 'BCM2836')
+        else:
+            return False
+pytestmark = pytest.mark.skipif(not is_a_pi(), reason='tests cannot run on non-Pi platforms')
+del is_a_pi
+
+
+# Try and import as many pin libraries as possible
+PIN_CLASSES = []
+try:
+    from gpiozero.pins.rpigpio import RPiGPIOPin
+    PIN_CLASSES.append(RPiGPIOPin)
+except ImportError:
+    RPiGPIOPin = None
+try:
+    from gpiozero.pins.rpio import RPIOPin
+    PIN_CLASSES.append(RPIOPin)
+except ImportError:
+    RPIOPin = None
+try:
+    from gpiozero.pins.pigpiod import PiGPIOPin
+    PIN_CLASSES.append(PiGPIOPin)
+except ImportError:
+    PiGPIOPin = None
+try:
+    from gpiozero.pins.native import NativePin
+    PIN_CLASSES.append(NativePin)
+except ImportError:
+    NativePin = None
+
+
+@pytest.fixture(scope='module', params=PIN_CLASSES)
+def pin_class(request):
+    # pigpiod needs to be running for PiGPIOPin
+    if request.param.__name__ == 'PiGPIOPin':
+        subprocess.check_call(['sudo', 'pigpiod'])
+        # Unfortunately, pigpiod provides no option for running in the
+        # foreground, so we have to use the sledgehammer of killall to get shot
+        # of it
+        def kill_pigpiod():
+            subprocess.check_call(['sudo', 'killall', 'pigpiod'])
+        request.addfinalizer(kill_pigpiod)
+    return request.param
+
+@pytest.fixture
+def pins(request, pin_class):
+    # Why return both pins in a single fixture? If we defined one fixture for
+    # each pin then pytest will (correctly) test RPiGPIOPin(22) against
+    # NativePin(27) and so on. This isn't supported, so we don't test it
+    test_pin = pin_class(22)
+    input_pin = pin_class(27)
+    input_pin.function = 'input'
+    input_pin.pull = 'down'
+    def fin():
+        test_pin.close()
+        input_pin.close()
+    request.addfinalizer(fin)
+    return test_pin, input_pin
+
+
+def test_pin_numbers(pins):
+    test_pin, input_pin = pins
+    assert test_pin.number == 22
+    assert input_pin.number == 27
+
+def test_function_bad(pins):
+    test_pin, input_pin = pins
+    with pytest.raises(PinInvalidFunction):
+        test_pin.function = 'foo'
+
+def test_output(pins):
+    test_pin, input_pin = pins
+    test_pin.function = 'output'
+    test_pin.state = 0
+    assert input_pin.state == 0
+    test_pin.state = 1
+    assert input_pin.state == 1
+
+def test_output_with_state(pins):
+    test_pin, input_pin = pins
+    test_pin.output_with_state(0)
+    assert input_pin.state == 0
+    test_pin.output_with_state(1)
+    assert input_pin.state == 1
+
+def test_pull(pins):
+    test_pin, input_pin = pins
+    test_pin.function = 'input'
+    test_pin.pull = 'up'
+    assert input_pin.state == 1
+    test_pin.pull = 'down'
+    test_pin, input_pin = pins
+    assert input_pin.state == 0
+
+def test_pull_bad(pins):
+    test_pin, input_pin = pins
+    test_pin.function = 'input'
+    with pytest.raises(PinInvalidPull):
+        test_pin.pull = 'foo'
+
+def test_pull_down_warning(pin_class):
+    pin = pin_class(2)
+    try:
+        with pytest.raises(PinFixedPull):
+            pin.pull = 'down'
+    finally:
+        pin.close()
+
+def test_input_with_pull(pins):
+    test_pin, input_pin = pins
+    test_pin.input_with_pull('up')
+    assert input_pin.state == 1
+    test_pin.input_with_pull('down')
+    assert input_pin.state == 0
+
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -245,6 +245,7 @@ class NativePin(Pin):
     def close(self):
         self.when_changed = None
         self.function = 'input'
+        self.pull = 'up' if self.number in (2, 3) else 'floating'
 
     def _get_function(self):
         return self.GPIO_FUNCTION_NAMES[(self._MEM[self._func_offset] >> self._func_shift) & 7]
@@ -253,7 +254,7 @@ class NativePin(Pin):
         try:
             value = self.GPIO_FUNCTIONS[value]
         except KeyError:
-            raise PinInvalidFunction('invalid function "%s" for pin %r' % self)
+            raise PinInvalidFunction('invalid function "%s" for pin %r' % (value, self))
         self._MEM[self._func_offset] = (
             self._MEM[self._func_offset]
             & ~(7 << self._func_shift)
@@ -282,7 +283,7 @@ class NativePin(Pin):
         try:
             value = self.GPIO_PULL_UPS[value]
         except KeyError:
-            raise PinInvalidPull('invalid pull direction "%s" for pin %r' % self)
+            raise PinInvalidPull('invalid pull direction "%s" for pin %r' % (value, self))
         self._MEM[self._MEM.GPPUD_OFFSET] = value
         sleep(0.000000214)
         self._MEM[self._pull_offset] = 1 << self._pull_shift
--- a/gpiozero/pins/pigpiod.py
+++ b/gpiozero/pins/pigpiod.py
@@ -13,6 +13,7 @@ from ..exc import (
     PinInvalidFunction,
     PinSetInput,
     PinFixedPull,
+    PinInvalidPull,
     )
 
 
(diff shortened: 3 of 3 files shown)


tests/test_real_pins.py
```diff
@@ -0,0 +1,143 @@
...
+import subprocess
+
+import pytest
+
+from gpiozero import PinFixedPull, PinInvalidPull, PinInvalidFunction
+
+
+# This module assumes you've wired the following GPIO pins together
+TEST_PIN = 22
+INPUT_PIN = 27
```
lurch: Did you intend to use these constants later on in the code? ;-)

lurch:
> PWM testing: statistical sampling?

I'd always imagined that you'd test the PWM stuff by using e.g. an MCP3008? ;-)  (with an appropriate smoothing capacitor if necessary)

New pin-tests work fine on my A+ with Jessie (I didn't install either of RPIO or pigpiod yet though).
On my Pi2 with Wheezy (where the gpiomem stuff doesn't exist, so talking to the real pins requires `sudo python`) `make test` (which had obviously worked until now) suddenly doesn't work anymore.
So I tried running `sudo make test` but then it complained that it couldn't find `coverage` as that was only installed into my venv. I managed to work around this by editing line 7 of `Makefile` to say `COVERAGE?=coverage`, and then running ``sudo COVERAGE=`which coverage` make test`` and then it all worked :)  (but I don't know if that's a likely-enough scenario to warrant a PR)

But even if running on a Pi where you *don't* need `sudo`, I can imagine there'd be scenarios where you don't want to run the real hardware tests (e.g. maybe you've already got other stuff hooked up to your GPIO pins), so it might be nice if there was an env-var you could set, and/or alternative Makefile target that you could run, that would always skip the hardware-level tests?

P.S. Not sure why, but running ``sudo COVERAGE=`which coverage` make test`` inside my py32 venv (on Wheezy, on Pi2) is now occasionally failing inside `tests/test_outputs.py::test_rgbled_fade_background` which is weird, as it'd always been fine up until now. Dunno if it's because python running as root, or real hardware being tested as well, is adding some kind of latency somewhere?

tests/test_real_pins.py
```diff
@@ -0,0 +1,143 @@
...
+    PIN_CLASSES.append(NativePin)
+except ImportError:
+    NativePin = None
+
+
+@pytest.fixture(scope='module', params=PIN_CLASSES)
+def pin_class(request):
+    # pigpiod needs to be running for PiGPIOPin
+    if request.param.__name__ == 'PiGPIOPin':
+        subprocess.check_call(['sudo', 'pigpiod'])
```
lurch: What happens if pigpiod was already running before you start running the tests?
Would it make sense to add a `kill_pigpiod` inside a module-level setup function?

tests/test_real_pins.py
```diff
@@ -0,0 +1,143 @@
...
+    test_pin.output_with_state(1)
+    assert input_pin.state == 1
+
+def test_pull(pins):
+    test_pin, input_pin = pins
+    test_pin.function = 'input'
+    test_pin.pull = 'up'
+    assert input_pin.state == 1
+    test_pin.pull = 'down'
+    test_pin, input_pin = pins
```
lurch: Is this an accidentally-duplicated line of code, or is it part of the magic fixture stuff?

tests/test_real_pins.py
waveform80: Doh! I originally did ... must've undone something at some point!

tests/test_real_pins.py
waveform80: Hmm, that's a decent idea. If pigpiod's already running we can't be sure it's not (for example) doing PWM on our test pins, so we should definitely make sure it's disabled before we dive into testing.

tests/test_real_pins.py
waveform80: Er yup, that's an accidental copy'n'paste (this was all rather hastily bashed together - and evidently it shows :)

waveform80:
> So I tried running sudo make test but then it complained that it couldn't find coverage as that was only installed into my venv. I managed to work around this by editing line 7 of Makefile to say COVERAGE?=coverage, and then running sudo COVERAGE=`which coverage` make test and then it all worked :) (but I don't know if that's a likely-enough scenario to warrant a PR)

Yeah,  that's worth a PR - frankly I ought to make all that stuff environment configurable.

> But even if running on a Pi where you don't need sudo, I can imagine there'd be scenarios where you don't want to run the real hardware tests (e.g. maybe you've already got other stuff hooked up to your GPIO pins), so it might be nice if there was an env-var you could set, and/or alternative Makefile target that you could run, that would always skip the hardware-level tests?

Good point - probably worth another env-var (might even warrant being off by default - banging away at GPIO pins by default might be construed as rude!)

> P.S. Not sure why, but running sudo COVERAGE=`which coverage` make test inside my py32 venv (on Wheezy, on Pi2) is now occasionally failing inside tests/test_outputs.py::test_rgbled_fade_background which is weird, as it'd always been fine up until now. Dunno if it's because python running as root, or real hardware being tested as well, is adding some kind of latency somewhere?
> EDIT: Even weirder, it only seems to fail if RPi.GPIO isn't pip-installed inside my py32 venv. Definitely a heisenbug ;)

That's ... yeah, I've got no idea what's going on there!

lurch:
> > So I tried running `sudo make test` but then it complained that it couldn't find coverage as that was only installed into my venv. I managed to work around this by editing line 7 of Makefile to say `COVERAGE?=coverage`, and then running ``sudo `COVERAGE=which coverage` make test`` and then it all worked :) (but I don't know if that's a likely-enough scenario to warrant a PR)

> Yeah, that's worth a PR - frankly I ought to make all that stuff environment configurable.

Yeah, and RPIO still uses `/dev/mem` rather than `/dev/gpiomem` and so also requires testing with `sudo`, even on the latest Jessie release of Raspbian.

tests/test_real_pins.py
lurch: And perhaps there ought to be a mechanism for optionally specifying the full path to the `pigpiod` binary, rather than always assuming that it'll be found on the PATH?
(and *if* the unit-tests are already being run as root - see RPIO - then I guess in that case there's no need for `sudo` here. Which suggests that perhaps there ought to be externally-settable `START_PIGPIOD` and `STOP_PIGPIOD` commands / env-vars?)

And I guess some way of unit-testing the 'remote connection' mode of pigpiod http://gpiozero.readthedocs.org/en/latest/api_pins.html#pigpiopin would be really neat too, but that's obviously a set of tests you'd always need to ask for manually, and could never be run automatically. Although I dunno how the setup/teardown code would work in that situation...?

waveform80: Yup, all good points; though I think I'm happy to leave off testing the remote capabilities of pigpiod for the time being (given that, when running locally it's still operating over a localhost connection, so if it works there it should all work remotely too). Maybe eventually I could imagine having a dual pi rig setup for testing remote stuff too, but that's way off yet.

waveform80:
And before I forget!

> > PWM testing: statistical sampling?

> I'd always imagined that you'd test the PWM stuff by using e.g. an MCP3008? ;-) (with an appropriate smoothing capacitor if necessary)

It's certainly the sensible thing to do but I was really hoping for a test suite that anyone could run without having to order (significant) extra parts. In other words, if it's possible (and reliable) to test everything using nothing more than a wire across two pins I'd love to do it that way. But if it's not, sure - an ADC is obviously the way to go.

lurch:
>  In other words, if it's possible (and reliable) to test everything using nothing more than a wire across two pins I'd love to do it that way.

I just knocked up a simple test app, and pleasingly it works much better than I expected :-)

```python
from __future__ import (
    print_function,
    division,
    )

import time

from gpiozero.pins.rpigpio import RPiGPIOPin

# This test assumes you've wired the following GPIO pins together
TEST_PIN = 22
INPUT_PIN = 27

def measure_pwm(pin, duration=1):
    if pin.function != 'input':
        raise ValueError('pin %r must be in "input" mode' % pin)
    states = [0,0]
    start = time.time()
    while time.time() - start < duration:
        states[pin.state] += 1
    total = states[0] + states[1]
    print('Average = %.3f after %d measurements over %d seconds (%.1f reads/sec)' % (states[1] / total, total, duration, total / duration))


pin_class = RPiGPIOPin
test_pin = pin_class(TEST_PIN)
input_pin = pin_class(INPUT_PIN)

test_pin.function = 'output'
test_pin.frequency = 100

for value in range(0, 11, 2):
    test_pin.state = value / 10
    time.sleep(0.1)
    print('\nSet PWM value to %.3f' % test_pin.state)
    measure_pwm(input_pin)
```

Running it in Py2.7 virtualenv on a Pi1 (A+) I get:
```
Set PWM value to 0.000
Average = 0.000 after 17376 measurements over 1 seconds (17376.0 reads/sec)

Set PWM value to 0.200
Average = 0.198 after 18342 measurements over 1 seconds (18342.0 reads/sec)

Set PWM value to 0.400
Average = 0.401 after 16192 measurements over 1 seconds (16192.0 reads/sec)

Set PWM value to 0.600
Average = 0.601 after 18997 measurements over 1 seconds (18997.0 reads/sec)

Set PWM value to 0.800
Average = 0.800 after 15555 measurements over 1 seconds (15555.0 reads/sec)

Set PWM value to 1.000
Average = 1.000 after 19680 measurements over 1 seconds (19680.0 reads/sec)
```

Running it in Py3.4 virtualenv on a Pi1 (A+) I get:
```
Set PWM value to 0.000
Average = 0.000 after 13594 measurements over 1 seconds (13594.0 reads/sec)

Set PWM value to 0.200
Average = 0.202 after 14930 measurements over 1 seconds (14930.0 reads/sec)

Set PWM value to 0.400
Average = 0.400 after 12806 measurements over 1 seconds (12806.0 reads/sec)

Set PWM value to 0.600
Average = 0.601 after 15592 measurements over 1 seconds (15592.0 reads/sec)

Set PWM value to 0.800
Average = 0.801 after 12291 measurements over 1 seconds (12291.0 reads/sec)

Set PWM value to 1.000
Average = 1.000 after 15810 measurements over 1 seconds (15810.0 reads/sec)
```

Running it in Py2.7 virtualenv on a Pi2 I get:
```
Set PWM value to 0.000
Average = 0.000 after 92475 measurements over 1 seconds (92475.0 reads/sec)

Set PWM value to 0.200
Average = 0.205 after 92468 measurements over 1 seconds (92468.0 reads/sec)

Set PWM value to 0.400
Average = 0.399 after 92518 measurements over 1 seconds (92518.0 reads/sec)

Set PWM value to 0.600
Average = 0.600 after 92502 measurements over 1 seconds (92502.0 reads/sec)

Set PWM value to 0.800
Average = 0.795 after 92442 measurements over 1 seconds (92442.0 reads/sec)

Set PWM value to 1.000
Average = 1.000 after 92413 measurements over 1 seconds (92413.0 reads/sec)
```

Running it in Py3.2 virtualenv on a Pi2 I get:
```
Set PWM value to 0.000
Average = 0.000 after 84751 measurements over 1 seconds (84751.0 reads/sec)

Set PWM value to 0.200
Average = 0.204 after 85022 measurements over 1 seconds (85022.0 reads/sec)

Set PWM value to 0.400
Average = 0.403 after 82339 measurements over 1 seconds (82339.0 reads/sec)

Set PWM value to 0.600
Average = 0.596 after 85251 measurements over 1 seconds (85251.0 reads/sec)

Set PWM value to 0.800
Average = 0.796 after 85008 measurements over 1 seconds (85008.0 reads/sec)

Set PWM value to 1.000
Average = 1.000 after 84710 measurements over 1 seconds (84710.0 reads/sec)
```

As expected, the Pi2 is able to do **much** faster GPIO reads than the Pi1. Strange that Py2 is consistently faster than Py3 though.

Given how well this has worked, I'll try to work it into the unit-tests this evening. Using edge-detection to attempt to measure the frequency might be interesting too ;)

waveform80:
Wow ... that's *way* better than I was expecting! I was worried that either we'd have to use unacceptably long sample times, or that at short sample times (sub-second) we'd have to use unacceptably large tolerances (< 0.01 or something) to get the tests passing. But that's a brilliant result!

waveform80:
> As expected, the Pi2 is able to do much faster GPIO reads than the Pi1. Strange that Py2 is consistently faster than Py3 though.

Not surprised that 2.7 was faster than 3.2; initially Python 3.x was much slower than 2.x as the initial focus was on cleaning up the code-base rather than performance. However, I'm quite surprised that 3.4 on the A+ was slower. I vaguely recall a big push on performance in either 3.3 or 3.4 (I know one of those made the major change of minimal unicode string representations which significantly lowered Py3's memory footprint over 2, fast Decimal types, etc). I think there was another load of optimizations in 3.5 but obviously that's not relevant yet.

lurch:
Another set of data-points to add to the collection above ;-)

Python 2.7 virtualenv on a Pi3B:
```
Set PWM value to 0.000
Average = 0.000 after 146121 measurements over 1 seconds (146121.0 reads/sec)

Set PWM value to 0.200
Average = 0.204 after 155086 measurements over 1 seconds (155086.0 reads/sec)

Set PWM value to 0.400
Average = 0.401 after 155232 measurements over 1 seconds (155232.0 reads/sec)

Set PWM value to 0.600
Average = 0.598 after 155816 measurements over 1 seconds (155816.0 reads/sec)

Set PWM value to 0.800
Average = 0.796 after 154823 measurements over 1 seconds (154823.0 reads/sec)

Set PWM value to 1.000
Average = 1.000 after 155775 measurements over 1 seconds (155775.0 reads/sec)
```

Python 3.4 virtualenv on a Pi3B:
```
Set PWM value to 0.000
Average = 0.000 after 132698 measurements over 1 seconds (132698.0 reads/sec)

Set PWM value to 0.200
Average = 0.205 after 140551 measurements over 1 seconds (140551.0 reads/sec)

Set PWM value to 0.400
Average = 0.403 after 140740 measurements over 1 seconds (140740.0 reads/sec)

Set PWM value to 0.600
Average = 0.598 after 141008 measurements over 1 seconds (141008.0 reads/sec)

Set PWM value to 0.800
Average = 0.795 after 140871 measurements over 1 seconds (140871.0 reads/sec)

Set PWM value to 1.000
Average = 1.000 after 141064 measurements over 1 seconds (141064.0 reads/sec)
```

PR #221, opened 2016-02-22
RPIOPin fixups

lurch:
...as discovered using the new real_pins unit-tests.


waveform80:
Nice - I hadn't managed to get around to digging out a Pi1 and testing RPIO yet!

PR #233, opened 2016-03-19
Correction on Traffic Lights Alternative recipe

ukBaz:
I got an error when using the documented example.
`NameError: name 'lights' is not defined`

This PR fixes that error


bennuttall:
Thanks Barry!

PR #239, opened 2016-03-30
Fix #140, fix #69, fix #185

### LINKED ISSUE #69: Energenie
It was suggested in #66 that we add an [Energenie](https://energenie4u.co.uk/index.phpcatalogue/product/ENER314) class.

I maintain the [energenie](https://pypi.python.org/pypi/energenie) module, which provides a simple interface to the Energenie Pi-mote. It's just a case of turning pins on and off in sequence.

Seeing as I'm the maintainer and it's not going to upset anyone, should we consider incorporating Energenie functionality into GPIO Zero?

I'd consider this low priority, would rather concentrate on other things before the 1.0 release, but wanted to start a thread now.

### LINKED ISSUE #140: Cannot install gpiozero due to missing dependency
Hi! I've been trying to install the python-gpiozero package on Raspbian but am told that it cannot be installed due to one of the dependencies being unavailable. Anyone who can help me solve this? I've run `apt-get update` and made sure I'm on the latest firmware (in case that helps). Other than that I'm out of ideas.

Here's the output from apt-get (in English for your convenience):

```
pi@raspberrypi ~ $ LANG=en_GB.UTF-8 sudo apt-get install python-gpiozero
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 python-gpiozero : Depends: python-spidev but it is not installable
E: Unable to correct problems, you have held broken packages.
```
```
pi@raspberrypi ~ $ LANG=en_GB.UTF-8 sudo apt-get install python-spidev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package python-spidev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'python-spidev' has no installation candidate
```

Output from apt-cache seems to indicate that python-spidev is a virtual package and that the functionality needs to be provided by another package. This is sadly going far beyond my knowledge of APT and I don't even know what type of information to look for at this point.

```
pi@raspberrypi ~ $ LANG=en_GB.UTF-8 apt-cache depends python-gpiozero
python-gpiozero
  Depends: python
  Depends: python
  Depends: python-rpi.gpio
  Depends: <python-spidev>
```

### LINKED ISSUE #185: Add error message when SPI not enabled
When you try to use an SPI component like `MCP3008`, and SPI has not been enabled, we should catch that and provide a helpful error message like "Try running sudo raspi-config" (I believe picamera does this) instead of "file not found" or whatever.

waveform80:
This PR adds a software SPI implementation. Firstly this removes the absolute necessity for spidev (#140), which also means when it's not present things still work (effectively fixes #185), and also enables any four pins to be used for SPI devices (which don't require the hardware implementation).

The software implementation is simplistic but still supports clock polarity and phase, select-high, and variable bits per word. However it doesn't allow precise speeds to be implemented because it just wibbles the clock as fast as it can (which being pure Python isn't actually that fast).

Finally, because this PR involves creating a framework for "shared" devices (like SPI devices with multiple channels), it made sense to bung Energenie (#69) in as well as this is a really simple shared device.

P.S. My apologies for the ridiculous size of this PR - messing around with the base hierarchy tends to result in such things, but I think after this the base hierarchy should be fairly "done". It might *extend* in the future (when we get to I2C components and such like) but I don't think there'll be many (any?) more major re-writes. Especially as this PR also greatly simplifies composite device construction (yeah ... this one sprawled quite a bit ...). Oh, and I still haven't written tests for all the new stuff (though I have done some limited testing with actual hardware).


bennuttall:
@waveform80 I played with Energenie today and realised the interface uses 0-indexing. Were you aware that the 4-socket Energenie strip is numbered 1 to 4? https://energenie4u.co.uk/catalogue/product/ENER010

The usage in my energenie library was you could provide `1` to `4` for a specific socket, and `0` for all four.

waveform80:
Ahh ... nope, wasn't aware of that product - I'll update it to use 1-4 instead. Not sure about the "all four" thing though ... if someone constructs a device representing socket 1, and then another that represents all four, given that it's impossible to read the state of the socket that leads to interesting issues keeping the states up to date. For now, I'd stick with controlling individual interfaces and leave the "all four" thing for future work.

lurch:
Just out of curiosity, how come this PR has seemingly modified every single image, even though most of them still look the same?

waveform80:
Good question - the PNGs, SVGs, and PDFs are all generated by `docs/Makefile` when it detects the source files have changed. I did write the Energenie stuff on my netbook whilst we were at my mother-in-law's for our daughter's first birthday, and that had a fresh checkout of my clone of the repo ... it's possible that messed up the relationship between the source and output's modification timestamps and `make` decided to regenerate everything as a result (then I just hit "gca" to commit everything as usual and didn't notice a crap-load of images going along for the ride!)

PR #243, opened 2016-04-02
Fix #237

### LINKED ISSUE #237: 'Keyboard controlled robot' requires python3-evdev
re: http://gpiozero.readthedocs.org/en/v1.1.0/recipes.html#keyboard-controlled-robot

This recipe needs python3-evdev to be installed as a prerequisite: this is not installed by default by raspbian-jessie. 

The installation instructions at https://python-evdev.readthedocs.org/en/latest/ are clear enough, but have further prerequisites that are not met by default.

The overall effect is that using this recipe is not quite 'frictionless.'

Also, it was not immediately obvious to me that the evdev library only deals with directly-connected physical devices. It does not interpret keystrokes from a 'virtual keyboard' such as might be the case if ssh'ing into a robot via a wifi connection.

waveform80:
Converts the keyboard controlled robot recipe to use curses. Advantage
is that no external deps are required, disadvantages are that it'll only
work from a terminal (or terminal window, basically not IDLE) and that
an explicit "stop" key (I've used the space bar) is required as
terminals only record key-presses, there's no way to get key release
events under them without resorting to stuff like evdev (which'd get us
right back to where we started).

I'd forgotten about the "no key release" thing under terminals, so I
won't merge this one myself - let me know what you think!

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -567,10 +567,52 @@ Use four GPIO buttons as forward/back/left/right controls for a robot::
 Keyboard controlled robot
 =========================
 
-.. XXX Rewrite this using curses (to avoid evdev dep, which isn't packaged
-   on Raspbian)
-
 Use up/down/left/right keys to control a robot::
+
+    import curses
+    from gpiozero import RyanteckRobot
+
+    robot = RyanteckRobot()
+
+    actions = {
+        curses.KEY_UP:    robot.forward,
+        curses.KEY_DOWN:  robot.backward,
+        curses.KEY_LEFT:  robot.left,
+        curses.KEY_RIGHT: robot.right,
+        }
+
+    def main(window):
+        next_key = None
+        while True:
+            curses.halfdelay(1)
+            if next_key is None:
+                key = window.getch()
+            else:
+                key = next_key
+                next_key = None
+            if key != -1:
+                # KEY DOWN
+                curses.halfdelay(3)
+                action = actions.get(key)
+                if action is not None:
+                    action()
+                next_key = key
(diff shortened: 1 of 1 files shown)


waveform80:
@bennuttall ping! Do you reckon it's worth merging this (with the explicit stop key requirement), or should we stick with the evdev requirement?

bennuttall:
I wondered if we should use both, and explain the advantages and disadvantages?

> There should be one-- and preferably only one --obvious way to do it.

*cough*

lurch:
I did some google-searching and found http://cboard.cprogramming.com/c-programming/123853-tracking-key-release-linux.html - any use?

But I also think (given the pros and cons of each) it's worth keeping and documenting each approach.

waveform80:
@lurch Oooh ... the penultimate message is intriguing. I hadn't noticed halfdelay in curses (despite trawling the docs recently). Okay, that's a decent compromise; I'll have a go in a bit when I've finished some more tests...

waveform80:
Okay, here's an updated version - have a play and let me know if it's good enough to replace the evdev one. If not, it can go just go in alongside it (I've had a brief play here with LEDs substituting for motor bridges and it looks decent).

bennuttall:
Haven't had time to test this yet, will try to soon!

waveform80:
@bennuttall any news? If you haven't got time to test today, pick from one of the following because 1.2 is going out by hell or high water today ;)

* Not needed: close it
* Might need it: bump the milestone to 1.3
* Compromise: let me know and I'll update the PR so both recipes are included

Issue #148, opened 2016-01-16
Add recipe for controlling onboard ACT LED (led0) (and camera LED)?


shmick:
I set the led to gpio mode with:
`echo gpio | sudo tee /sys/class/leds/led0/trigger`

Yet I'm still unable to address the onboard green / activity led.

Raspberry Pi = A+
OS = raspbian jessie
python2.7 = 2.7.9-2
python-gpiozero = 1.0.0

Tested with and without `active_high=False`
```
from gpiozero import LED
from signal import pause
from time import sleep

led1 = LED(16, active_high=False)
led1.blink()

while True:
  print(led1.is_lit)
  sleep(.3)

pause()
```
I get a stream of True and False results, but the LED never comes on.

```
True
True
True
True
False
False
False
True
True
True
```

shmick:
Blurg.

Turns out I was using the wrong GPIO number. 16 and active_high=False is only for the mode B.

The newer B+, A+, 2B, and Zero** use:
```
red = LED(35)
green = LED(47)
```

**No onboard red LED on the zero

waveform80:
Huh ... didn't even know that was possible - thanks for the info! I wonder if that should be bunged in the docs somewhere (in an advanced section, obviously).

bennuttall:
And camera LED? Or leave it to picamera

bennuttall:
Knew it was possible to toggle the activity light, never knew which pin number it was! Thanks for the info.

I forget, do we have an upper limit on pin numbers? Don't think so. I must give the compute module a try :)

shmick:
I think it's worthwhile to add this into the docs. It's a no-cost solution to 1 or 2 LED outputs, depending on the board you have.

waveform80:
After the pins PR is merged, the pin limit will be defined by whatever module is implementing the pins (naturally). Both RPi.GPIO and the native implementation allow all pin numbers from 0 to 53 (as that's all the GPIO registers in /dev/mem) - I think RPIO was stricter but can't recall off the top of my head.

waveform80:
Yup, let's add these to the docs (I'll add a little recipe somewhere at the end); as mentioned above - it's a straight-forward way of playing with the library without any additional hardware at all which is a decent use-case.

bennuttall:
Look ma: no additional components!

https://vine.co/v/i1pEVeJ2X9Y

waveform80:
Very nice :)

waveform80:
[...]

PR #246, opened 2016-04-03
Add pins database

waveform80:
Related to @lurch's comments on #148, this PR contains a database of
pins for each Pi revision, along with various other bits of miscellany
(I might've gotten a bit carried away here...).

Any corrections/extensions welcome!

### DIFF
--- a/docs/api_exc.rst
+++ b/docs/api_exc.rst
@@ -93,6 +93,12 @@ Errors
 
 .. autoexception:: PinPWMFixedValue
 
+.. autoexception:: PinMultiplePins
+
+.. autoexception:: PinNoPins
+
+.. autoexception:: PinUnknownPi
+
 Warnings
 ========
 
--- a/docs/api_pins.rst
+++ b/docs/api_pins.rst
@@ -46,7 +46,8 @@ can pass a :class:`Pin` object itself::
     led = LED(NativePin(16))
 
 This is particularly useful with implementations that can take extra parameters
-such as :class:`PiGPIOPin` which can address pins on remote machines::
+such as :class:`~gpiozero.pins.pigpiod.PiGPIOPin` which can address pins on
+remote machines::
 
     from gpiozero.pins.pigpiod import PiGPIOPin
     from gpiozero import LED
@@ -73,40 +74,46 @@ to utilize pins that are part of IO extender chips. For example::
 RPiGPIOPin
 ==========
 
-.. currentmodule:: gpiozero.pins.rpigpio
-
-.. autoclass:: RPiGPIOPin
+.. autoclass:: gpiozero.pins.rpigpio.RPiGPIOPin
 
 
 RPIOPin
 =======
 
-.. currentmodule:: gpiozero.pins.rpio
-
-.. autoclass:: RPIOPin
+.. autoclass:: gpiozero.pins.rpio.RPIOPin
 
 
 PiGPIOPin
 =========
 
-.. currentmodule:: gpiozero.pins.pigpiod
-
-.. autoclass:: PiGPIOPin
+.. autoclass:: gpiozero.pins.pigpiod.PiGPIOPin
 
 
 NativePin
 =========
 
-.. currentmodule:: gpiozero.pins.native
-
-.. autoclass:: NativePin
+.. autoclass:: gpiozero.pins.native.NativePin
 
 
 Abstract Pin
 ============
 
-.. currentmodule:: gpiozero.pins
-
 .. autoclass:: Pin
     :members:
+
+
+Utilities
+=========
+
+The pins module also contains a database of information about the various
+revisions of Raspberry Pi. This is used internally to raise warnings when
+non-physical pins are used, or to raise exceptions when pull-downs are
+requested on pins with physical pull-up resistors attached. The following
+functions and classes can be used to query this database:
+
+.. autofunction:: pi_info
+
+.. autoclass:: PiBoardInfo
+
+.. autoclass:: PinInfo
 
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -8,6 +8,11 @@ from __future__ import (
 from .pins import (
     Pin,
 )
+from .pins.data import (
+    PiBoardInfo,
+    PinInfo,
+    pi_info,
+)
 from .exc import (
     GPIOZeroError,
     DeviceClosed,
@@ -37,6 +42,9 @@ from .exc import (
     PinPWMError,
     PinPWMUnsupported,
     PinPWMFixedValue,
+    PinUnknownPi,
+    PinMultiplePins,
+    PinNoPins,
     GPIOZeroWarning,
     SPIWarning,
     SPISoftwareFallback,
--- a/gpiozero/exc.py
+++ b/gpiozero/exc.py
@@ -97,6 +97,15 @@ class PinPWMUnsupported(PinPWMError, AttributeError):
 class PinPWMFixedValue(PinPWMError, AttributeError):
     "Error raised when attempting to initialize PWM on an input pin"
 
+class PinUnknownPi(PinError, RuntimeError):
+    "Error raised when gpiozero doesn't recognize a revision of the Pi"
+
+class PinMultiplePins(PinError, RuntimeError):
+    "Error raised when multiple pins support the requested function"
+
+class PinNoPins(PinError, RuntimeError):
+    "Error raised when no pins support the requested function"
+
 class GPIOZeroWarning(Warning):
     "Base class for all warnings in GPIO Zero"
 
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -1,0 +1,696 @@
+from __future__ import (
+    unicode_literals,
+    absolute_import,
+    print_function,
+    division,
+    )
+str = type('')
+
+import io
+from collections import namedtuple
+
+from ..exc import PinUnknownPi, PinMultiplePins
+
+
+# Some useful constants for describing pins
+
+V1_8   = '1V8'
+V3_3   = '3V3'
+V5     = '5V'
+GND    = 'GND'
+NC     = 'NC' # not connected
+GPIO0  = 'GPIO0'
+GPIO1  = 'GPIO1'
+GPIO2  = 'GPIO2'
+GPIO3  = 'GPIO3'
+GPIO4  = 'GPIO4'
+GPIO5  = 'GPIO5'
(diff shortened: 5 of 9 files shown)


lurch:
Yes, you've probably got a bit carried away, but I guess more info is better than less ;-)

This is also probably the ideal place for adding the ACT and PWR LED gpio-info? :)

If you're looking for other things to add, you could also add whether a board has a RUN header, and whether it has separate audio and video headers (A, B) or a combined AV header (A+/B+/2/3) or only a solderable video header (Zero, maybe CM?).

I assume you've already cross-checked against http://elinux.org/RPi_BCM2835_GPIOs and http://elinux.org/RPi_Low-level_peripherals and http://elinux.org/RPi_HardwareHistory and http://pinout.xyz/ ( https://github.com/Gadgetoid/Pinout2/blob/master/src/en/pi-pinout.yaml specifically may contain useful machine-readable info)

waveform80:
It's definitely the place to put ACT and PWR ... eventually. That's partially going to depend on exactly how we address things like GPIO expanders (given that's where they sit on the Pi3) - but I'm having some thoughts about that at the moment. I hadn't dug into Phil's yaml but those are indeed all the sources I've been using so far (plus the CM schematics for its enormous pin list).

So ... pin addressing. Just to float some ideas: at the moment when creating a device we accept one of two things: an integer (which we convert into a Pin implementation) or a Pin implementation. What if we added to that strings in certain formats. For example:

* `"GPIO12"` would mean GPIO12 (i.e. the same as `12` would at the moment)
* `"P1/3" would mean the third physical pin on header P1 (whatever that happened to be on the current Pi)
* `"P5/3" would mean the third physical pin on header P5
* `"I2C:0x4A/1" would mean the first physical pin on the IO extender attached to I2C at address 0x4A (still need to look into this sort of thing, but this is a rough sketch).
* Perhaps we could include `"WP8"` to mean WiringPi pin 8 for those familiar with this means of addressing.

I realize that some of this sounds like BOARD mode (oh no!) and in some ways it is (permits addressing of pins by physical location) but it lacks the one significant drawback of BOARD mode: no global state. In other words, there's still no "mode" that the module is switched into which would potentially result in modules fighting over the mode or copy'n'pasted code using the wrong mode.

Basically I'm thinking I'd add another module under pins which would define a factory function which, given some specification like the above, would try and figure out what it meant for the current Pi and would return an appropriate Pin implementation (then devices.GPIODevice simply calls this factory function with whatever it's given).

Thoughts?

lurch:
Hmmmm, +0 on having device-constructors additionally accept and parse string arguments. Does it lead to a slippery slope? (Should "GPIO06" be parsed the same as "GPIO6"? What about "gpio6"? etc. How many other alternative pin-naming schemes will get shoe-horned into these strings? Should there be an optional extra part to the string to signify whether RPi.GPIO or RPIO or pigpiod should be used?)

Definitely -1 on the "P1/3" notation - to me it reads as "p one-third". (so if you do go down this route, perhaps `_` (or something else) would be a better separator character than `/` ?)
And for your "I2C:..." example (if you add that), you'd also need to specify exactly which chip it is you're talking to (as each chip uses it's own communication protocol - all I2C defines is a way to read and write bytes (as does SPI)). This reminds me slightly of the 'opener' syntax used with http://docs.pyfilesystem.org/en/latest/commands.html

> Basically I'm thinking I'd add another module under pins which would define a factory function

I'd be quite happy to have something like CurrentPiBoard().get_physical_pin(x) (or IOExpander(address).get_pin(y) ) returning an already-instantiated Pin object (which I guess you'll be doing anyway), which could then be passed into a device-constructor; but I can see that typing out CurrentPiBoard().get_physical_pin(x) would quickly get tiresome, so perhaps there ought to be some shorter route of accomplishing the same.

OTOH, it seems to already be established that "GPIO Zero uses BCM pin numbering" (e.g. http://raspi.tv/2016/raspio-pro-hat-putting-pi-ports-in-perfect-positions-for-productive-play )

I guess @bennuttall (with his links to Picademy) may have a better idea about whether this would simply over-complicate the whole thing.

waveform80:
My main rationale for "/" was thinking in terms of URIs but that's a good point about P-one-third. Perhaps "P1:3" or something else. I'll have to think more about I2C when we get to tackling that (next release most likely). The IOExpander().pins[n] or some similar variant is already planned but as you note it's a bit of a mouthful (or keyboard-ful) when hacking away on the terminal (hence my desire for a simpler short-hand parsed out of some string format).

On the picademy side of things it probably won't matter that much: I've already made sure all the [worksheets](https://github.com/waveform80/picademy_worksheets) use BCM numbering exclusively and it's almost the first thing we cover in the first workshop of the day (I think the general idea is if we can get them over that hurdle quickly then we can concentrate on the more important things like coding skills).

That said, I do still have some (mild) sympathy for the viewpoint that physical pin numbering can be easier (it's not something I'd try and sway picademy towards, but I acknowledge some people are deeply set against anything but physical numbering). My main reasons for objecting to physical numbering (i.e. BOARD mode) in the past were:

1. First and foremost if the library *only* supports physical numbering then non-physical GPIOs (camera LEDs, power LEDs, etc.) become inaccessible (not *that* important for newbies, but I considered it important to keep the library as useful as possible to all skill ranges so we don't force people away from it as they grow their skills).
2. Secondly that the global state of the pin-numbering mode in RPi.GPIO/RPIO is nasty (import any module that changes the mode and you potentially break existing setups).

Well, the global state thing is basically solved: we use BCM mode under the hood and that's that. And as for the extended addressing it doesn't remove the capability to address GPIO numbers - so no worries about denying use-cases. On the other hand, it does introduce the possibility of mixing GPIO and physical numbering but I figure that's a case of "if the user wants to be deliberately silly, then that's their business".

Anyway - just some food for thought - not going to dive into implementing it right now (just finishing off the new-revision-parsing) and even if I do it's not something I'd merge without @bennuttall's approval as it'd be a fairly controversial change.

lurch:
I think in the past (when both B rev1 and B rev2 were common) there was more of an argument for using BOARD mode, as it "insulated" the users' code against being run on different Pis (but even then, to access the P5 header on a rev2 you still had to use BCM mode). With all the 'modern' Pis (A+, B+, B2, Zero, B3) using the identical 40-pin-layout, I guess this is now much less of an issue.

waveform80:
Yeah, that's certainly true. The only place where it currently causes an issue in picademy is that we typically hand out GPIO "shields" (either the [RasPIO ruler](http://rasp.io/gpioruler/) or the [portsplus](http://rasp.io/gpioruler/)) so that people can easily locate the GPIOs they need in the workshops. Unfortunately, because they've got both BCM and BOARD numbering on them (depending on which side you flip them) there's always a subset who stick it on the BOARD numbering side and then find their code doesn't work. Doesn't take long to figure out what's going on but we generally strive to try and eliminate even the possibility of such things happening.

Not that adding physical numbering possibilities to the library would help that in any way; I guess the "proper" solution for that is to hope that future shields won't include physical pin numbering at all and just use the BCM numbering ... but I rather doubt that's going to happen.

bennuttall:
ping @raspitv

lurch:
> Unfortunately, because they've got both BCM and BOARD numbering on them (depending on which side you flip them) there's always a subset who stick it on the BOARD numbering side and then find their code doesn't work.

At some of the CamJam workshops I've helped out at in the past (that were using the portsplus), the worksheets have explicitly said something like "make sure the keyring-hole is on the side away from the Pi".

The ProHAT I linked to earlier only has GPIO numbers (and see the stretch goals https://www.kickstarter.com/projects/raspitv/raspio-pro-hat-protect-position-raspberry-pi-ports/description )

And http://raspi.tv/2015/raspio-gpio-ruler-goes-on-general-sale#comment-59771 ;-D

raspitv:
The Pro Hat is BCM only, The ruler has BCM one side and Alt+BCM on the other. The only board with pin numbering is the Portsplus. It would be possible to do an edition with BCM on one side and Alt+BCM on the other, but it means forking the product, which carries a financial risk and additional complications of people buying the wrong one etc. Unless I abolish the pin numbering altogether, which then renders the product useful to less people (and mean I am deciding for them what is best - which is, maybe not so good, even if I always use BCM myself). That doesn't mean it can't be done, but it isn't something to be just done on a whim. There has to be a business case for it. I will get complaints either way (you know what people are like). Perhaps the best solution is if I did a Picademy version with BCM on one side and Alt+BCM on the other?

lurch:
@raspitv If you _did_ do a slightly-different version (which is something I have no opinion about), it might be worth producing it with a different-colour PCB, so it's easy to tell at-a-glance which is which?

Alternatively, perhaps @bennuttall and @waveform80 could paint over the pin-numbers-only side of the picademy portspluses with black paint? ;-)

waveform80:
@raspitv Yeah, the Pro Hat solves the issue nicely, but I don't believe we're planning to use that (yet?) in the GPIO workshop, while portsplus (and the ruler) are pretty heavily used (I must admit we had a few issues with the ruler last time - the weight of it hanging off the device levered a few people's jumper cables off their Pis, though this may have been slightly "timid" wiring by them - but it was the first time using it - we'll have another go next week :).

I figured it probably wouldn't be an easy decision commercially to drop BOARD numbering - I suspect to the majority of your customers portsplus is probably more useful with both numbering systems on it (as they may encounter both in the wild); picademy is a rather more "narrow" case and it's not reasonable for us to expect special treatment here. That said, I'm kicking myself I didn't think of @lurch's black paint suggestion ... I'm sorely tempted to try that!

waveform80:
Okay, nearly there with this - just got to get the tests working on non-Pi machines (for Travis) then I can merge this and rebase the source/values stuff...

bennuttall:
> Alternatively, perhaps @bennuttall and @waveform80 could paint over the pin-numbers-only side of the picademy portspluses with black paint? ;-)

> That said, I'm kicking myself I didn't think of @lurch's black paint suggestion ... I'm sorely tempted to try that!

Also kicking self! (admittedly a bit of a faff)

waveform80:
Ah, you've merged source/values toolkit - okay, no worries :) Kind of astounded this doesn't conflict with that ridiculously large patch-set ... git's merge skills are impressing me!

lurch:
> > > Alternatively, perhaps @bennuttall and @waveform80 could paint over the pin-numbers-only side of the picademy portspluses with black paint? ;-)

> > ....
That said, I'm kicking myself I didn't think of @lurch's black paint suggestion ... I'm sorely tempted to try that!

> Also kicking self! (admittedly a bit of a faff)

self.kick() ?  ;-)

PR #247, opened 2016-04-03
Fix up GPIO device hierarchy diagram

waveform80:
Oops - missed this last time


lurch:
Looks like it's still missing DistanceSensor?

waveform80:
Doh ... completely forgot that ... I'll deal with it in a mo.

PR #256, opened 2016-04-06
Fix #204 and start readying the release

### LINKED ISSUE #204: Should the 'duplicate pin checking' be more robust?
With the variety of "pin backends" now in GpioZero (see http://gpiozero.readthedocs.org/en/latest/api_pins.html ), I wonder if the checking-for-duplicate-pins inside GPIODevice could/should be stricter?
Consider the following:
```python
>>> from gpiozero import LED
>>> from gpiozero.pins.rpigpio import RPiGPIOPin
>>> from gpiozero.pins.native import NativePin
>>> led1 = LED(21)
>>> led2 = LED(RPiGPIOPin(21))
  File "gpiozero/devices.py", line 92, in __call__
    result = super(GPIOMeta, mcls).__call__(*args, **kwargs)
  File "gpiozero/output_devices.py", line 104, in __init__
    super(DigitalOutputDevice, self).__init__(pin, active_high, initial_value)
  File "gpiozero/output_devices.py", line 42, in __init__
    super(OutputDevice, self).__init__(pin)
  File "gpiozero/devices.py", line 167, in __init__
    super(SourceMixin, self).__init__(*args, **kwargs)
  File "gpiozero/devices.py", line 236, in __init__
    'pin %r is already in use by another gpiozero object' % pin
gpiozero.exc.GPIOPinInUse: pin GPIO21 is already in use by another gpiozero object
>>> led3 = LED(NativePin(21))
>>> led1.pin
GPIO21
>>> led3.pin
GPIO21
>>> led1.pin == led3.pin
False
```

It's good that `LED(21)` and `LED(RPiGPIOPin(21))` are recognised as referring to the same pin, but I wonder if it should also recognise that `LED(NativePin(21))` is pointing to the same pin?

Given that the `repr()` of `RPiGPIOPin(21)` is the same as the repr of `NativePin(21)` I guess one possible solution would be to change the `_PINS` in `gpiozero/devices.py` to be a `dict` (indexed by the pin repr-strings) instead of a `set` ?

waveform80:
Also re-numbers energenie sockets 1-4 (as noted by @bennuttall in
comments to #239), and adds several "real pins" tests and board tests.
The bad-PWM stuff is currently disabled as it causes segfaults when
running the tests and I can't seem to trace the cause at the moment.
Finally, I've tweaked the deb config to suggest gpiozero, removed spidev
as a mandatory dep (which'll fix installs on wheezy for py3), and
there's some more miscellaneous last-minute stuff here that I can't
recall...

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -11,7 +11,8 @@ X-Python3-Version: >= 3.2
 Package: python-gpiozero
 Architecture: all
 Section: python
-Depends: ${misc:Depends}, ${python:Depends}, python-rpi.gpio, python-spidev
+Depends: ${misc:Depends}, ${python:Depends}, python-rpi.gpio
+Suggests: python-spidev, python-gpiozero-docs
 Description: Simple API for controlling devices attached to the GPIO pins.
  gpiozero builds on RPi.GPIO to provide a set of classes designed to simplify
  interaction with devices connected to the GPIO pins, from simple buttons and
@@ -23,7 +24,8 @@ Description: Simple API for controlling devices attached to the GPIO pins.
 Package: python3-gpiozero
 Architecture: all
 Section: python
-Depends: ${misc:Depends}, ${python3:Depends}, python3-rpi.gpio, python3-spidev
+Depends: ${misc:Depends}, ${python3:Depends}, python3-rpi.gpio
+Suggests: python3-spidev, python-gpiozero-docs
 Description: Simple API for controlling devices attached to the GPIO pins.
  gpiozero builds on RPi.GPIO to provide a set of classes designed to simplify
  interaction with devices connected to the GPIO pins, from simple buttons and
--- a/docs/api_boards.rst
+++ b/docs/api_boards.rst
@@ -99,3 +99,10 @@ CamJam #3 Kit Robot
     :inherited-members:
     :members:
 
+Energenie
+=========
+
+.. autoclass:: Energenie
+    :inherited-members:
+    :members:
+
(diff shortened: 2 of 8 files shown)


bennuttall:
> The bad-PWM stuff is currently disabled

Bad-PWM?

> I've tweaked the deb config to suggest gpiozero

Suggest gpiozero? You mean gpiozero-doc?

waveform80:
Argh ... still had some failing tests there! Hang on and I'll push another fix...

waveform80:
Bad PWM stuff ... I've been trying to test the pin's PWM implementations statistically as discussed with @lurch in #220 - it works but if I throw in more than a single test I get segfaults which I've been unable to trace (as soon as I switch the venv to use a python with debug-symbols the segfaults disappear, and native-pin is suddenly unable to switch pin functions ... it's bizarre). Anyway, I'm happy enough with the progress made on the test suite to call it quits for this release and look into that stuff later.

And yes, I meant the docs ;)

waveform80:
But seriously - if the checks have failed, *don't merge it*. We should be using the test suite to determine whether stuff is suitable for merging or not - and sometimes I make a PR just to run it through Travis and see if it fails on 3.x (I tend to develop and test in a 2.x environment just to ensure 2.x compatibility).

lurch:
> sometimes I make a PR just to run it through Travis and see if it fails on 3.x

Is it worth looking into something like [tox](https://tox.readthedocs.org/en/latest/) ?

waveform80:
Yeah - it's been something I've been meaning to read up on for ages but I never seem to get a free moment (and for the most part Travis fulfils my needs).

Sorry - I'm being thick - what needs updating in DigitalOutputDevice?

lurch:
> Sorry - I'm being thick - what needs updating in DigitalOutputDevice?

The bit where it says "This class extends OutputDevice with a toggle() method to switch the device between its on and off states" - toggle is now in OutputDevice, not DigitalOutputDevice.

PR #259, opened 2016-04-07
Ensure CompositeDevice._order has deterministic behaviour

lurch:
... by sorting the kwargs, if no _order was specifically supplied.

### DIFF
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -261,7 +261,7 @@ class CompositeDevice(Device):
         self._tuple = None
         self._order = kwargs.pop('_order', None)
         if self._order is None:
-            self._order = kwargs.keys()
+            self._order = sorted(kwargs.keys())
         self._order = tuple(self._order)
         for missing_name in set(kwargs.keys()) - set(self._order):
             raise CompositeDeviceBadOrder('%s missing from _order' % missing_name)


lurch:
I'm not convinced by the failure-mode of the Travis test. Could you restart it for me please @bennuttall ?

waveform80:
Yeah, looks like something broke horribly on that travis instance - that's not a test-suite failure. I've kicked off that job again. Oh, and nice idea too!

lurch:
With this PR now merged, does the "If omitted, an arbitrary order will be selected for keyword arguments." text in http://gpiozero.readthedocs.org/en/latest/api_generic.html#gpiozero.CompositeDevice need to be updated?

And does `_order` need to be 'documented' as a parameter to http://gpiozero.readthedocs.org/en/latest/api_generic.html#gpiozero.CompositeOutputDevice ?

P.S. Does readthedocs have another level of sub-menu available? It's a bit confusing that `CompositeDevice` is in the http://gpiozero.readthedocs.org/en/latest/api_generic.html#base-classes section, but `CompositeOutputDevice` is in the http://gpiozero.readthedocs.org/en/latest/api_generic.html#composite-devices section. So if the left-hand menu could be expanded to show the class-names too, that would make things **much** easier to find!

P.P.S. Seems like "class gpiozero.SPIDevice" is actually documented in http://gpiozero.readthedocs.org/en/latest/api_generic.html twice?!?

waveform80:
> With this PR now merged, does the "If omitted, an arbitrary order will be selected for keyword arguments." text in http://gpiozero.readthedocs.org/en/latest/api_generic.html#gpiozero.CompositeDevice need to be updated?

Yup, good call - I'll bung that in in a sec.

> P.S. Does readthedocs have another level of sub-menu available? It's a bit confusing that CompositeDevice is in the http://gpiozero.readthedocs.org/en/latest/api_generic.html#base-classes section, but CompositeOutputDevice is in the http://gpiozero.readthedocs.org/en/latest/api_generic.html#composite-devices section. So if the left-hand menu could be expanded to show the class-names too, that would make things much easier to find!

Sure, reST supports lots of heading levels and how these are translated into navigation links is essentially theme defined. However, most (and I think the default which we're using is no exception) use the `:maxdepth:` setting from the toc (which you can find in `docs/index.rst`). So if we bump that to 3 I think it should add another level to the nav tree. That said, when I've done that in the past it looked kinda huge and scary so I generally default to 2 nowadays. I could just move `CompositeOutputDevice` up to the base classes section - would that be clearer?

> P.P.S. Seems like "class gpiozero.SPIDevice" is actually documented in http://gpiozero.readthedocs.org/en/latest/api_generic.html twice?!?

Ooops ... yup, it's documented under base classes and SPI devices (see `docs/api_generic.rst`)! Erm ... I'll remove it from the SPI devices section as base class seems more appropriate to me, but it's a rather arbitrary decision.

waveform80:
#271 implements those changes

waveform80:
Hmmm ... I see what you mean about things being hard to find in the generic devices section. It's become a bit of a behemoth. On the other hand, I'm rather reluctant to change the toc's maxdepth to 3 (just tried on a local build of the docs and the result just looks scarily huge on the front page.

I wonder if that section could do with being broken up into several chapters? Sticking with the theme of documenting things from the bottom of the hierarchy upwards we could start with generic input devices, then generic output devices. We could probably throw away the SPI devices section for now (just bung AnalogInputDevice in the SPI devices chapter). Same for the Composite devices section (throw LEDCollection into the Boards and Accessories chapter). Then finally have a Base classes chapter which contains the base classes section and maybe mixins (although they might warrant their own chapter at this point).

waveform80:
I'm out most of the day doing family things but feel free to have a play with this stuff - I'll take a look when we're back this evening. I'm planning on releasing 1.2 tomorrow as I've got a bit more time then and I want to get all this out before the madness of another picademy next week, so that's the deadline for changes.

To do a local docs build, run "make develop" to install dependencies in your venv, then "make doc" will construct a local docs build with the results under `docs/_build/html` and `docs/_build/latex`. You probably want `docs/_build/html/index.html` as your starting point (local builds don't use the RTD theme as it's not available in the version of sphinx packaged under Raspbian, but it should illustrate the structure that RTD will construct just as well).

PR #260, opened 2016-04-07
Fix #115

### LINKED ISSUE #115: Button.when_held(function_ref, seconds)
I'm looking for some pointers...  I think it would be a good to add Button.when_held which would fire a function when a button has been held down for a period of time. Its achievable by timing when_pressed and when_released but its a function I use a lot to determine between a short press and a long press. Presumably it would also be a good idea to have a wait_for _held(seconds, timeout=None)?

Providing you agree Im happy to make the change and issue a pull but could I get some pointers as to where the change should be made as Im struggling to find my path through the modules structure.

waveform80:
Adds when_held event hook to Button (via extension of the EventsMixin
class)


waveform80:
@bennuttall - don't merge this yet - need to finish off hold_repeat.

gpiozero/mixins.py
```diff
@@ -270,13 +280,119 @@ def _fire_events(self):
...
+    owning device is active.
+    """
+    def __init__(self, parent):
+        super(RepeatingThread, self).__init__(target=self.held, args=(parent,))
+        self.holding = Event()
+        self.start()
+
+    def held(self, parent):
+        while not self.stopping.wait(0):
+            if self.holding.wait(0.1):
```
lurch: Just out of curiosity, where does the 0.1 value here come from? I'm not that familiar with threaded code, but if it's necessary, should it be configurable somehow?

gpiozero/mixins.py
```diff
@@ -270,13 +280,119 @@ def _fire_events(self):
             if not old_state and new_state:
```
lurch: Would this be neater if the `else:` on the line above this was changed to `elif old_state != new_state:`, and then `if not old_state and new_state:` could be changed to just `if new_state:` and `elif old_state and not new_state:` could be changed to `else:`
?

gpiozero/input_devices.py
```diff
@@ -255,8 +255,12 @@ class Button(DigitalInputDevice):
         performed. Otherwise, this is the length in time (in seconds) that the
         component will ignore changes in state after an initial change.
     """
-    def __init__(self, pin=None, pull_up=True, bounce_time=None):
+    def __init__(
+            self, pin=None, pull_up=True, bounce_time=None,
+            hold_time=1, hold_repeat=False):
```
lurch: Will these new params need adding to the docstring?

waveform80: Yup - and I still need to write tests for all this :)

gpiozero/mixins.py
waveform80: Unfortunately that wouldn't work in one rare (but possible) circumstance. Just to help us visualize the code, you're proposing:

```python
if old_state is None:
    # initial state
elif new_state:
    # state change to active
else:
    # state change to inactive
```

Now, consider what happens if the device is currently active and _fire_events is called while the event is still active. It's rare, but in cases of extremely rapid switching it can occur (and different pin implementations make this more or less likely according to their implementation). This is largely because of the way the underlying hardware works. The GPIO registers have bits to set level/edge detection and then a register that indicates whether a level/edge was detected - however that register doesn't tell you *which* edge was detected. In the case you're only looking for a single edge you can obviously figure it out, but in the case you're detecting "both" edges (a common scenario), you can't. Moreover, multiple edges can occur while that register is set (it's just on/off, not a counter), so it's entirely possible to "miss" edges if your code isn't reacting to them fast enough.

So, what do we do in such circumstances? We could report the event anyway - the device (probably) was activated after all, but given that we can't guarantee losing events anyway, it seemed cleaner to only report events on clear transitions between states where we can definitely say we've gone from not active to active and vice versa.

Basically, it's a rather arbitrary design choice - do feel free to open a ticket about it though; it may be worth having a discussion with everyone about what they want.

gpiozero/mixins.py
waveform80: It's a vaguely arbitrary number plucked from the air, but with some background knowledge. The most important rule for background threads is generally: don't suck up all the CPU time. The key to that is to make sure the thread spends the vast majority of its time in an efficient wait state (basically anything like `Event.wait`, `Lock.acquire`, or even `sleep` although the latter isn't that useful).

So, let's look at the code with "n" as our wait there and walk through the various possibilities:

```python
        while not self.stopping.wait(0):
            if self.holding.wait(n):
                while not (
                        self.stopping.wait(0) or
                        parent._inactive_event.wait(parent.hold_time)
                        ):
                    parent._is_held = True
                    parent._fire_held()
                parent._is_held = False
```

Let's say *n* is 0. Now, if the device is inactive `holding` won't be set so we'll be looping around touching only `self.stopping.wait(0)` and `self.holding.wait(0)`. In other words, we're not waiting at all so our thread's going to consume loads of time. Once the device becomes active we start hitting `parent._inactive_event.wait(parent.hold_time)` so assuming the hold time hasn't been set to zero we'll be fine, but as soon as it's inactive again we'll be back to eating CPU.

Let's say *n* is 1. Now our loop's nice and CPU friendly even when the device is inactive. However, what happens when `self.stopping` is set to terminate the loop (when the owning component is closed/destroyed)? Now we've got up to a 1 second wait before the loop will terminate. Okay, not that big of a deal in most scenarios but generally you do want components to shut down in a reasonable length of time (if only because our test suite is likely to go creating and destroying lots of these things, and we don't want to spend too much time hanging around for that!).

So, 0.1 is a vague "guess" I plugged in initially figuring I wanted it to be CPU friendly but not take ages shutting down. I could probably lower that to 0.01 without any serious difference ... perhaps even 0.001. But 0.1's doing shut down quick enough for me at the moment (though I might change my mind once I write some tests for all this!).

gpiozero/mixins.py
lurch: Not quite, I was proposing:
```python
if old_state is None:
    # initial state
elif old_state != new_state:
    if new_state:
        # state change to active
    else:
        # state change to inactive
```
i.e. _exactly_ the same logic as before, just restructured slightly. Apologies for not explaining it very well.

waveform80: Ahhh, I see - yeah that's nicer.

gpiozero/mixins.py
lurch: Thanks Dave (as always!) for your detailed explanation. When I was looking at this code late last night, I hadn't realised the subtlety of there actually being _two_ nested while loops. And I still haven't quite got my head around there being so many different Event objects... ;)
But don't worry, I'm sure I'll get there eventually, so please focus on the code rather than taking time out to explain things to threading novices!

gpiozero/mixins.py
```diff
@@ -266,17 +300,141 @@ def _fire_events(self):
...
+        The length of time (in seconds) to wait after the device is activated,
+        until executing the :attr:`when_held` handler. If :attr:`hold_repeat`
+        is True, this is also the length of time between invocations of
+        :attr:`when_held`.
+        """
+        return self._hold_time
+
+    @hold_time.setter
+    def hold_time(self, value):
+        self._hold_time = value
```
lurch: Is it worth making this `self._hold_time = float(value)` ?

waveform80: Yeah, may as well - just in case someone does something weird like assigning a string

gpiozero/mixins.py
```diff
@@ -266,17 +300,141 @@ def _fire_events(self):
...
+        return self._held_from is not None
+
+    @property
+    def held_time(self):
+        """
+        The length of time (in seconds) that the device has been held for.
+        This is counted from the first execution of the :attr:`when_held` event
+        rather than when the device activated, in contrast to
+        :attr:`active_time`. If the device is not currently held, this is
+        ``None``.
```
lurch: So the first time that `when_held` gets fired, this is basically always zero?

gpiozero/mixins.py
waveform80: Hmm, should also check for a negative value (looking at the setter for source_delay which is similar). I'm going to be slightly naughty and rejig some of the exceptions as well because now a lot of functions are devolved into mixins they're really not specific to GPIODevice anymore so they shouldn't be GPIODeviceError descendants but GPIOZeroError descendants (yeah, okay, it's a compatibility break but I'll be amazed if anyone notices it :)

gpiozero/mixins.py
waveform80: Yup, that's the idea. Here's a quick script I threw together to test it out (it's curses so run it from the terminal):

```python
import curses
from gpiozero import *

btn = Button(17)
btn.hold_time = 1

def main(window):
    while True:
        window.addstr(0, 0, 'is_pressed: %-5s' % btn.is_pressed)
        window.addstr(1, 0, 'pressed_time: %-18s' % btn.pressed_time)
        window.addstr(2, 0, 'is_held: %-5s' % btn.is_held)
        window.addstr(3, 0, 'held_time: %-18s' % btn.held_time)
        window.refresh()

curses.wrapper(main)
```

gpiozero/mixins.py
```diff
@@ -266,17 +299,143 @@ def _fire_events(self):
...
+        until executing the :attr:`when_held` handler. If :attr:`hold_repeat`
+        is True, this is also the length of time between invocations of
+        :attr:`when_held`.
+        """
+        return self._hold_time
+
+    @hold_time.setter
+    def hold_time(self, value):
+        if value < 0:
+            raise BadWaitTime('source_delay must be 0 or greater')
```
lurch: hold_time, surely? ;)

waveform80: Doh!

PR #262, opened 2016-04-08
Additional parameter-type validation

lurch:
...in line with the property setters

Is it also worth checking that average is callable?


waveform80:
Hmm, not so worried about average - it's unlikely many users are going to customize that and even if we do check it's callable that still doesn't guarantee it's got an appropriate prototype or that it returns something sensible. But this the additional stuff here is useful nonetheless.

PR #265, opened 2016-04-08
Add LEDBarGraph to the composed_devices graph

lurch:
Feel free to reject if this ends up making the [graph](http://gpiozero.readthedocs.org/en/latest/api_generic.html) too cluttered. Alternatively I guess the "LEDBoard" node could be changed to "LEDBoard / LEDBarGraph" ?


waveform80:
Ah, good point - I'd forgotten those. And no, it's nowhere near the clutter of the other graphs (yet ;)

PR #272, opened 2016-04-10
Remove the CompositeDevice.tuple() method

lurch:
IMHO having a public method called tuple() just makes things too confusing - the first time I saw `self.tuple(some_args)` I automatically assumed it was a typo of just `tuple(some_args)`. And the way it was used in CompositeDevice.value as `return self.tuple(some_args)` when it was defined as being `def tuple(self):` just looks weird. And it's not included in the API documentation either.

I'm not sure I can see a use for it being a public property anyway - why would a user actually want/need to get at the underlying namedtuple type (rather than just relying on whatever the `value` property returns) ?
OTOH if you do want to keep it as a public property (and reject this PR), perhaps it should be renamed to `namedtuple` or `device_tuple_type` and then added to the documentation.

### DIFF
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -329,12 +329,8 @@ class CompositeDevice(Device):
         return all(device.closed for device in self)
 
     @property
-    def tuple(self):
-        return self._tuple
-
-    @property
     def value(self):
-        return self.tuple(*(device.value for device in self))
+        return self._tuple(*(device.value for device in self))
 
     @property
     def is_active(self):
--- a/tests/test_devices.py
+++ b/tests/test_devices.py
@@ -90,7 +90,7 @@ def test_composite_device_sequence():
         assert len(device) == 2
         assert device[0].pin.number == 2
         assert device[1].pin.number == 3
-        assert device.tuple._fields == ('_0', '_1')
+        assert device._tuple._fields == ('_0', '_1')
 
 def test_composite_device_values():
     with CompositeDevice(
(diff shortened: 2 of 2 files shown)


waveform80:
The main reason for having the function was so the user could get hold of the tuple type used by the CompositeDevice in order to manually construct objects to feed to `value`. For example, imagine you've got some device with the sub-components `foo` and `bar` and you have a function which yields values for those components. It'll work perfectly well without names:

```python
class MyDevice(CompositeOutputDevice):
    def __init__(self):
        super(MyDevice, self).__init__(foo=LED(4), bar=LED(5))

def my_source():
    v = True
    while True:
        yield tuple(v, not v)
        v = not v

my_device = MyDevice()
my_device.source = my_source()
```

But I think it's a little nicer with them:

```python
class MyDevice(CompositeOutputDevice):
    def __init__(self):
        super(MyDevice, self).__init__(foo=LED(4), bar=LED(5))

def my_source(target):
    v = True
    while True:
        yield target.tuple(foo=v, bar=not v)
        v = not v

my_device = MyDevice()
my_device.source = my_source(my_device)
```

Still, I take your point that it probably shouldn't be called tuple. namedtuple sounds a lot better to me, and it should be added to the docs (that's purely an oversight on my part).

I'm just working on a re-org of the docs after some experiments last night. I *think* I've got something that's a little nicer, so I'll rename tuple to namedtuple for that and include it in the docs.

waveform80:
Okay, we've gone with `namedtuple` in the 1.2 release, so I'll close this - but thanks for the suggestion - it was definitely worth doing something about it!

lurch:
The reason I didn't close this myself after the release of 1.2, was because you haven't added `namedtuple` to the docs yet... ;)

waveform80:
Doh! Doc-string's missing. I'll get to this in a bit (too busy with picamera stuff right now)

PR #273, opened 2016-04-10
Check that the sub-devices of CompositeDevice are actually Device instances

lurch:
Not sure if `CompositeDeviceBadDevice` is the best name, so feel free to rename the exception


lurch:
Hmmm, does codecov not get re-triggered when a PR is force-updated?

waveform80:
Hmm, apparently not! Anyway, good call on the PR - thanks!

PR #274, opened 2016-04-10
Doc reorganization

waveform80:
A fairly major re-organization of the docs; the generic API section had become excessively long and difficult to navigate. In this version, where base classes pertain to particular section (inputs, outputs, composites, etc.) they are listed in that section at the end (so that the first things users come across in a section is still the concrete classes).

The generic section now consists purely of Device and the mixins (this *might* be extended to include GPIOBase and GPIOMeta in future if there's demand for it). The generic section also includes a near-complete class hierarchy diagram for those particularly interested (given that the section is now of extremely limited interest anyway, it seemed a reasonable place to put such a thing).

Finally, several of @lurch's suggestions are incorporated (renaming CompositeDevice.tuple to CompositeDevice.namedtuple, conjunction to all_values, disjunction to any_values, etc.)


lurch:
Looks good.
Seems like the "Base classes" in each chapter are included 'backwards' though? e.g. given that CompositeOutputDevice extends CompositeDevice, and LEDCollection extends CompositeOutputDevice, should they be listed in the opposite order to what they are now?

And it seems like `GPIODevice` is included at both http://gpiozero.readthedocs.org/en/v1.2.0/api_input.html#gpiodevice and http://gpiozero.readthedocs.org/en/v1.2.0/api_output.html#gpiodevice - perhaps it should be moved to http://gpiozero.readthedocs.org/en/v1.2.0/api_generic.html ?

> The generic section also includes a near-complete class hierarchy diagram for those particularly interested

As this image is "too big for the page" it gets automatically scaled down. Is it possible to have it so that clicking on the image automatically opens it (full-size) in a new window?

waveform80:
> Looks good.
> Seems like the "Base classes" in each chapter are included 'backwards' though? e.g. given that CompositeOutputDevice extends CompositeDevice, and LEDCollection extends CompositeOutputDevice, should they be listed in the opposite order to what they are now?

The main reason for sticking them at the end is two-fold: first, don't scare the newbies away with frightening things like abstract base classes unless they go looking for them, and secondly we're documenting everything backwards anyway (concrete stuff first, to abstract last - largely for the same reasons, so that anyone mad enough to read the docs end to end ... ahem ... works their way up to the more abstract, yet complex, portions of the hierarchy).

I'm wondering whether I should include the *whole* hierarchy including GPIOBase and GPIOMeta now that Base Classes is slimmed down, but they really should go at the end for the crazy people who make it that far up the class hierarchy :)

> And it seems like GPIODevice is included at both http://gpiozero.readthedocs.org/en/v1.2.0/api_input.html#gpiodevice and http://gpiozero.readthedocs.org/en/v1.2.0/api_output.html#gpiodevice - perhaps it should be moved to http://gpiozero.readthedocs.org/en/v1.2.0/api_generic.html ?

I deliberate included GPIODevice twice, but the second instance is marked with "noindex" to avoid conflicts (it's explicitly redundant). I could place it in the base-classes bit, but then it sort of felt like SPIDevice and CompositeDevice ought to be in there as well (seeing as they're the bases of their respective bits of the hierarchy), and it was exactly that bloated situation in the base classes chapter we were trying to get away from. I could combine the input and output sections (as a "GPIO Device" section) but I rather like the input/output separation at the moment.

>>    The generic section also includes a near-complete class hierarchy diagram for those particularly interested

> As this image is "too big for the page" it gets automatically scaled down. Is it possible to have it so that clicking on the image automatically opens it (full-size) in a new window?

Unfortunately not that I know of. However, in most browsers you can simply right click and select "View Image" (or whatever the equivalent is) to view the whole thing solo, and as it's an SVG you can always zoom the page without any loss of quality.

Issue #219, opened 2016-02-21
RFC: Polymorphic Parameters for CompositeDevice subclasses


lurch:
I've been playing around with the ideas I had in https://github.com/RPi-Distro/python-gpiozero/pull/131#issuecomment-186243356 and https://github.com/RPi-Distro/python-gpiozero/pull/131#issuecomment-186251847 and I've managed to knock up a prototype which does exactly what I want, **and** still passes all the tests! :-)
https://github.com/lurch/python-gpiozero/tree/polymorphic_parameters

I'm not sure if it's the 'best' way of implementing it though, or whether it's just too confusing an idea to be practical? Which is why I'm opening it up for comments before I even _begin_ to think about creating a PR.
(and I guess this sort-of ties into the ideas in https://github.com/RPi-Distro/python-gpiozero/issues/107#issuecomment-156827276 and https://github.com/RPi-Distro/python-gpiozero/issues/189#issuecomment-183126169 too?)

In my polymorphic_parameters tree, you can do everything I suggested in https://github.com/RPi-Distro/python-gpiozero/pull/131#issuecomment-186251847
i.e. the standard `TrafficLightsBuzzer(TrafficLights, Buzzer, Button)` and `Robot((int, int), (int, int))` still work as normal; but you can also now do `TrafficLightsBuzzer((int, int, int), int, int)` and `Robot(Motor, Motor)`. However you decide to give the parameters to the CompositeDevices, the constructed objects end up exactly the same.

And since a `Robot` is a CompositeDevice of `Motor`s, and `Motor` is a CompositeDevice of `PWMOutputDevice`s, and a `PWMOutputDevice` can be constructed from a `Pin` or an `int`...... in princicple you can do totally wacky things like `robot = Robot((PWMOutputDevice(int), Pin(int)), Motor(int, PWMOutputDevice(Pin(int))))` and it all magically just works (obviously you'd *never* write code like that, but it illustrates how flexible this approach is).
And if you really wanted to, you could do `led = RGBLED(PWMLED(int), PWMLED(int), PWMLED(int))`

Comments? Criticisms? Suggestions?   Have I simply gone mad? ;)

lurch:
https://github.com/lurch/python-gpiozero/tree/polymorphic_parameters2 contains the same code, but 'rebased' on top of the 1.3 release of gpiozero.

lurch:
[...]

PR #276, opened 2016-04-12
Add Phase/Enable motor (and robot)

MrHarcombe:
As it turns out, the Pololu driver board I bought for my "little yellow robot" project uses phase/enable by default (which I hadn't heard of before, but now realise means 1 pin for direction, and 1 PWM pin for power). So I added a PhaseEnableMotor and a PhaseEnableRobot to get me going.

The robot probably needs to be done differently - if I were back in Java, I would be looking to create a "Motor" interface then have 2 concrete classes for the standard motor, and the new phase/enable one. But I don't (currently) have the Python nouse for that. Give it time...

Happy to be directed to improve this, as required - I'm an ex-developer (now a teacher), so I can probably handle the critique :)


lurch:
Interesting... I'm sure @waveform80 will have some good pointers on how this should be "done properly" once he's finished being busy with picademy, but in the meantime here's some brief comments from me:

 * It would be nice if you could squash the commits together, and then force-push the branch, which will make this PR much neater.

 * I'm not sure that `PhaseEnableMotor` is the best name... perhaps `SpeedDirectionMotor` (or something)? In hindsight, perhaps the current `Motor` class should have been named `HBridgeMotor`?

 * If only the `Robot` class could optionally take `Motor` objects in its constructor... (hint hint #219  @waveform80 ;-) ) ... then your separate PhaseEnableRobot wouldn't be necessary. (and you could even use two different Motor types within the same Robot)

 * In version 1.3 of gpiozero, the way CompositeDevices get constructed has changed slightly (and you need to implement fewer methods) - see https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L723

 * I wonder if your PhaseEnableMotor.forward should actually do something like:
```python
    self.power_device.off()
    self.direction_device.on()
    self.power_device.value = speed
```
(and similarly for PhaseEnableMotor.backward) ? Might be important if e.g. you do forward(0.8) followed by backward(0.1), without there being a 'phantom' backward(0.8) in there? (but perhaps this isn't actually a problem, given how quickly the Raspberry Pi runs??)

 * There's probably no _real_ reason for PhaseEnableMotor.stop needing to do `self.direction_device.off()` ?


But on the whole, it looks fairly good.
With regards to your interface comments, Python tends to use "implicit interfaces" - see https://en.wikipedia.org/wiki/Duck_typing

MrHarcombe:
@lurch thanks for the comments - very useful indeed. I will look at how to squash commits together, and do that, for starters.

Your polymorphic tree does exactly the sort of thing I was talking about with Java interfaces - given the lack of parameter typing constraints. Thank you - fabulous example to work from.

Naming is always a challenge - but the motor board I'm using also calls itself an [H-bridge interface](https://www.pololu.com/product/2753), which is also why I went for PhaseEnable (as opposed to it's other way of operating, IN/IN).

Your code suggestions are absolutely spot-on, I suspect and so I will at least put those in before I do anything else.

Thanks, again.

lurch:
If the board you're using is specific to the RaspberryPi, and will always connect to specific GPIO pins, then it makes sense to make a dedicated class for it, like http://gpiozero.readthedocs.org/en/latest/api_boards.html#ryanteck-mcb-robot or http://gpiozero.readthedocs.org/en/latest/api_boards.html#camjam-3-kit-robot

And if you modified that board to work in IN/IN mode, then it would work as-is with the current Motor and Robot classes - the Motor class in gpiozero supports the [0, 0], [0, PWM] and [PWM, 0] modes (using software PWM) described on the page you linked to.

MrHarcombe:
I did wonder if that was what I meant with its in/in mode. Couldn't quite get my head around what the table was trying to tell me; and, as I'd already got the code modified to work with the phase/enable, I left it at that and got back to the rest of my project 😀

I quite agree, re the specific robot implementation, but without the way of changing the robot motor instances, I went with what we used to call "code re-use method 2" (ie copy/paste).

waveform80:
> It would be nice if you could squash the commits together, and then force-push the branch, which will make this PR much neater.

Oddly enough GitHub now allows us to auto-squash all the commits in a PR together. You can also view all the commits squashed together (click on the "Showing changes from 1 commit" link at the top left). So, no having to teach people about rebase any more \o/

> If only the Robot class could optionally take Motor objects in its constructor... (hint hint #219 @waveform80 ;-) ) ... then your separate PhaseEnableRobot wouldn't be necessary. (and you could even use two different Motor types within the same Robot)

Hehe - I was thinking *exactly* the same thing while reading the PR! I'll try and find some time to look at #219 and this further at the weekend, but right now I really need to concentrate on getting a long overdue picamera release out so my gpiozero time will be a bit reduced over the next week.

lurch:
Just for info, there's yet another style of motor-driver-chip discussed in #300

lurch:
And another one: #314 

And maybe there's also a need for a UnidirectionalMotor sub-class of Motor, which only has values settable in the 0->1 range, and calling `backward()` would raise an Excepton. (Or perhaps it would be better to have the current Motor be a subclass of UnidirectionalMotor?)

MrHarcombe:
Okay, so I've tried to keep this PR "current" by pulling in the latest code, fixing the minor conflict and bringing my alternative PhaseEnableRobot in line with the default one. What have I not understood, about "coverage", please?

lurch:
"coverage" is simply complaining because you've added a load of new code, but none of it is covered by the unit-tests, and so the proportion of tested-lines to total-lines has decreased. I wouldn't worry about it for now.

I'm afraid I think you've made things harder for yourself by creating a pull-request directly from your `master` branch, rather than forking e.g. a `phase_enable_motor` branch off your `master`, and then PR-ing from that. E.g. this PR is still showing `docs/api_boards.rst` as having changed, which I guess is something you didn't intend?

MrHarcombe:
I'm beginning to realise this. Is it going to be easier to start again with
a new branch and a new PR? Or branch and make any subsequent changes on
that branch, this keeping it away from the PR?

Probably wouldn't hurt to consider test cases for my little bit of code, at
some point 😃

Ah, the joys of novices joining in...

lurch:
> easier to start again with a new branch and a new PR?

That would be my recommendation (there may be a 'neater' way using git commands, but taking a backup of your current code, and starting with a fresh branch off a 'clean' master and reapplying your changes is probably the simplest approach)

> Probably wouldn't hurt to consider test cases for my little bit of code, at some point

Shouldn't be too difficult to copy the existing `Motor` and `Robot` test cases, and modify them as needed for your new classes.

> Ah, the joys of novices joining in...

No problem, every little helps! We were all novices at one point ;-)  I find contributing to be a great way of learning.

lurch:
Couple of other quick comments...

"Extends :class:`CompositeDevice` to represent a generic dual-motor robot."
should probably be clarified to point out that it'll create a robot using PhaseEnableMotors rather than 'normal' Motors.

I wonder if the PhaseEnableMotor constructor needs a `direction_active_high` parameter (defaulting to `True`), in order to be able to invert the logic of the `direction_device`?

MrHarcombe:
Following @lurch 's advice, I'm going to close this PR and start again with the same modification, but on a separate branch on my fork, next time.

PR #277, opened 2016-04-13
Docs update - Motor is composed of two PWMOutputDevices

lurch:
If this gets merged, I'll leave @waveform80 to do all the image-files creation ;)

Hmmm, and looking at the inheritance graphs on http://gpiozero.readthedocs.org/en/latest/api_output.html#base-classes and http://gpiozero.readthedocs.org/en/latest/api_boards.html#base-classes should the Motor docs be moved from api_output.html to api_boards.html ?

### DIFF
--- a/docs/images/composed_devices.dot
+++ b/docs/images/composed_devices.dot
@@ -16,4 +16,5 @@ digraph classes {
     TrafficLightsBuzzer->Button;
 
     Robot->Motor;
+    Motor->PWMOutputDevice;
 }


waveform80:
I'll deal with re-generating the charts in a moment. For anyone interested in the process behind this, it's fairly simple:

1. Make sure you've got graphviz and Inkscape installed on your system and accessible in your PATH (which they will be by default in just about any Linux).

2. Set up a development build of gpiozero in the usual way (venv, git clone, make develop), then when you `make doc` within that venv after editing any `.dot` files you should find it uses graphviz (dot specifically) and Inkscape to generate the necessary SVG, PNG, and PDF files.

The specific commands are in `docs/Makefile` for those interested in the mechanics of this. You'll also find gnuplot and mscgen mentioned in there simply because I've used those for [certain](http://compoundpi.readthedocs.org/en/release-0.4/protocol.html) [charts](http://picraft.readthedocs.org/en/release-0.6/vectors.html#vector-vector-operations) in other bits of documentation I've done over the years.

lurch:
I found that in order for `make doc` to not error, I had to install a bunch of **huge** latex packages too :-/
(looking through my apt logfiles, looks like I needed texlive-latex-base, texlive-fonts-recommended and texlive-latex-extra)

Maybe it'd be worth having smaller doc 'sub-targets' in the Makefile that you could use to just build parts of the documentation?

lurch:
And what about my question regarding moving the Motor docs from api_output.html to api_boards.html, given that Motor itself is a CompositeDevice rather than an OutputDevice?

waveform80:
Ah yes - I keep forgetting I tend to install most of texlive by default. But yes, some separate targets would be handy (or perhaps some sensible detection that always builds the HTML but only builds the PDF if the right stuff's available) - someone remind me about this next week when I've got a sec (this week I'm concentrating on picamera and bits - because dammit that's *waaaaay* overdue :)

> And what about my question regarding moving the Motor docs from api_output.html to api_boards.html, given that Motor itself is a CompositeDevice rather than an OutputDevice?

Hmm ... you're absolutely right at the code level. My only concern is that from a simpler point of view, people don't see a motor as a board or a composite device; it's "just" a motor (although the driver is quite obviously a board). @bennuttall - any comments about Motor going to boards and accessories in the docs?

bennuttall:
Hmm. Unsure. I think you'd expect to find it in outputs, because we treat a motor like a component, but I can understand why you might want to move it away from there, because it requires a MCB. What do you think?

bennuttall:
Surely the same applies to RGBLED?

lurch:
> Surely the same applies to RGBLED?

Well in code-terms, `RGBLED` [derives](http://gpiozero.readthedocs.org/en/latest/api_output.html#base-classes) from `Device`, but `Motor` [derives](http://gpiozero.readthedocs.org/en/latest/api_boards.html#base-classes) from `CompositeDevice`.
And in electronics-wiring-terms, you **do** connect an RGB LED directly to the Pi's GPIO pins (with current-limiting resistors, of course), but you never connect an actual motor directly to the Pi's GPIO pins.

So with half a mind on #276 as well, maybe the current `Motor` class could be deprecated (so as not to break backwards-compatibility?), an abstract `MotorDriver` class could be created, and `HalfBridgeMotorDriver` (based on the current `Motor` class) and `PhaseEnableMotorDriver` classes could be concrete implementations, and then `Robot` could be modified to expect two `MotorDriver` instances? 
(although just as `Motor` proved to be too generic a name (as shown by #276 ), perhaps `Robot` is also too generic a name, and should be named something like `TwoWheeledRobot` or `DualMotorRobot` instead?)

In terms of 'expecting' to find the documentation for Motor in the "Output Devices" section of the docs, maybe this could remain in-place as just a small explanation that motors can't be used "bare" but must always be used with a driver board, with a link to the MotorDriver docs in the "Boards and Accessories" section?

waveform80:
Yup, liking just about all of this. Unfortunately I won't have the time this week to jump into it, and I've still got to review #219 properly which'd alleviate the backwards compat concerns on `Robot` but this sounds like a good direction to go.

Generally I'm not concerned with where things derive from (from the documentation perspective); I doubt most new users care what class descends from what - that's more for people hacking on the library or advanced users deriving from it, but this path keeps backwards compatibility, keeps the docs nice and simple, and adds plenty more possibilities for robots in future.

(not sure about robot class names yet either though :)

MrHarcombe:
FWIW, I created a [branch on my fork](https://github.com/MrHarcombe/python-gpiozero/tree/lurch_polymorphism) that brought @lurch's latest polymorphic branch for [#219](https://github.com/RPi-Distro/python-gpiozero/issues/219) which meant I could define a simpler PololuRobot that constructs itself around two PhaseEnableMotor instances on the default pins.

Look much neater :)

MrHarcombe:
Regarding Robot class names, other systems (I'm thinking Lego, here) would be looking at this as a "tank" robot, as the steering is done by the two separate drive motors, as opposed to more of a "car" robot which would only really need one drive motor (with or without differential) and then a separate steering motor.

PR #284, opened 2016-04-17
Include tests in release source tarball

EdwardBetts:
I'm thinking about packaging gpiozero so it can be included in Debian. It would be great to include the test suite in the pypi tarball so the tests can be run when the Debian package is built.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include README.rst
+recursive-include tests *.py


waveform80:
Sounds reasonable (the tests won't increase the source-dist much - a few kb at most). Just wondering why Debian would be interested in such a pi-specific package ... or are you thinking about the remote control pin possibilities? :)

EdwardBetts:
- I see a Debian package as another way of distributing the software, like making it available via pypi.
- Including gpiozero in Debian will be useful for anybody wanting to run Debian on their Raspberry Pi.
- A developer might want to write an application that uses the gpiozero library. That application can be included in Debian with a dependency on the gpiozero module.

waveform80:
Ah, if you mean a package for Raspbian - that's already in hand. Raspbian packages have already been built and shipped up to the Raspbian servers - we're just waiting on the maintainer to push them into production (the 1.1.0 packages are still the most recent available as I write this, but the 1.2.0 packages should be in staging somewhere).

bennuttall:
It'd be great to see this is Debian too, if that's what you mean. Is there any reason the package couldn't go into standard Debian currently? Any rules the Raspbian version is breaking?

EdwardBetts:
Yes, I'm interested in uploading the package to Debian. The existing packaging looks great, I'm just going to make a few changes.

Currently it is a native package. The version I'm going to upload to Debian will be a non-native package. 

Lintian gives some warnings, I'll investigate and fix them.
```
W: python-gpiozero-doc: extra-license-file usr/share/doc/python-gpiozero-doc/html/_sources/license.txt
X: python-gpiozero-doc: privacy-breach-generic usr/share/doc/python-gpiozero-doc/html/index.html (https://badge.fury.io/py/gpiozero.svg)
X: python-gpiozero-doc: privacy-breach-generic usr/share/doc/python-gpiozero-doc/html/index.html (https://travis-ci.org/rpi-distro/python-gpiozero.svg?branch=master)
I: python-gpiozero: description-synopsis-might-not-be-phrased-properly
I: python3-gpiozero: description-synopsis-might-not-be-phrased-properly
```

waveform80:
Huh - I don't see the privacy breach stuff - guessing that's new to lintian on Debian. I've seen the extra license file thing for a while, but figured that was ignorable (it's referring to the copy of the license in the docs vs. the license file in the root of the package that gets included in everything, including the docs package redundantly). Haven't see the description-synopsis stuff before either (new feature again?); that said I do build stuff on wheezy to ensure the py3 packages work so I'm using a *very* out of date debuild.

Anyway, this is all good enough to merge - if you come up with any neat fixes for the stuff above, do let me know!

lurch:
https://lintian.debian.org/tags/privacy-breach-generic.html

While it's useful to have the Travis-build-status displayed on http://gpiozero.readthedocs.org/en/latest/ it's probably less-useful in the offline packaged docs?

waveform80:
Ahh - yes, I've been wondering how to remove those from the offline docs. Haven't looked into conditionals in reST yet (or whether I'll need to do something more complex like including a generated file).

waveform80:
My head's getting full of picamera stuff right now, and I know I'm going to forget this - any chance someone can open a reminder ticket for the privacy-breach bits? I'll grab it when I get a spare second

PR #285, opened 2016-04-17
correct spelling of Artificial

EdwardBetts:


bennuttall:
Thanks! But I've a feeling that breaks the RST header. Can you add another `=` to the next line to fix it?

waveform80:
Oh dear - Dave's speling fails again ;)

PR #291, opened 2016-04-19
Add docs/_build to .gitignore

lurch:


lurch:
Wow, two codecovs for the price of one?!

PR #292, opened 2016-04-19
Add extra Source Tools functions: smoothed, summed and multiplied

lurch:
I also started to add methods to add-a-fixed-scalar and multiply-by-a-fixed-scalar, but then realised those could both be accomplished by careful use of the existing `scaled` function.

Feel free to come up with better usage-examples before merging this PR ;-)

### DIFF
--- a/docs/api_tools.rst
+++ b/docs/api_tools.rst
@@ -43,6 +43,8 @@ Single source conversions
 
 .. autofunction:: queued
 
+.. autofunction:: smoothed
+
 .. autofunction:: scaled
 
 Combining sources
@@ -53,6 +55,10 @@ Combining sources
 .. autofunction:: any_values
 
 .. autofunction:: averaged
+
+.. autofunction:: summed
+
+.. autofunction:: multiplied
 
 Artificial sources
 ==================
--- a/gpiozero/tools.py
+++ b/gpiozero/tools.py
@@ -218,6 +218,54 @@ def averaged(*values):
         yield mean(v)
 
 
+def summed(*values):
+    """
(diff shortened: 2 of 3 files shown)


waveform80:
Lovely stuff - here's a challenge for you, if you're up for it (I've been wanting to do this and haven't found the time yet):

* Make source-tool that converts a floating-point value to a boolean value (False/True) given a target threshold.
* Optional parameters should enable specification of the input range (default 0-1), threshold (default 0? 0.5?), and here's the fun one: hysteresis (default none).

bennuttall:
Neat!

lurch:
Challenge accepted... I guess the easiest option will just be to have a float min_value and a float max_value, and return True when the 'input value' is between those values - which should neatly incorporate all the input range and threshold stuff.

I'll have to give the hysteresis stuff a little bit more thought ;-)

BTW did you get my email @waveform80 ?

waveform80:
> Challenge accepted... I guess the easiest option will just be to have a float min_value and a float max_value, and return True when the 'input value' is between those values - which should neatly incorporate all the input range and threshold stuff.

Ah - much nicer than my idea! Yes, great stuff.

> BTW did you get my email @waveform80 ?

Yup - I'll respond as soon as I get a second but I'm really trying to get picamera 1.11 done this weekend before picademy next week (which'll just wind up pushing it back another week). Stereoscopic's proving ... well, I've got the basics working but there's a preview issue which is difficult to sort out without breaking lots of other stuff!

lurch:
> here's a challenge for you, if you're up for it

Ta-daa! #298 :-)

PR #293, opened 2016-04-19
Make Potentiometer Recipe more efficient

lurch:


bennuttall:
I saw this recipe yesterday and wondered why we bother using `with` at all? It's inconsistent with the rest of the examples and (like picamera) just means people assume they *have* to do it that way.

waveform80:
Hmm, demonstration that such a thing was possible - but you're right - it's inconsistent.

bennuttall:
I found a `with PiCamera()...` block too. Removed that.

Accidentally referred to the wrong issue #...

PR #299, opened 2016-04-28
Update recipes.rst

raspitv:
Small correction. LED to PWMLED in the Light Sensor 2nd example


bennuttall:
Thanks Alex!

One small tip: use descriptive commit messages and PR titles :)

PR #301, opened 2016-04-28
Update output_devices.py

goloplo:
To be able to use all the Pins when using a Motor. Maybe future definition for modifying the speed (PWM) should be added as separate command.

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -753,8 +753,8 @@ class Motor(SourceMixin, CompositeDevice):
                 'forward and backward pins must be provided'
             )
         super(Motor, self).__init__(
-                forward_device=PWMOutputDevice(forward),
-                backward_device=PWMOutputDevice(backward),
+                forward_device=DigitalOutputDevice(forward),
+                backward_device=DigitalOutputDevice(backward),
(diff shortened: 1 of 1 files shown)


lurch:
I don't think this fix is correct as-is (it would break existing code that relies on the Motor class supporting variable speeds), but it does raise an interesting point - there should probably be some way to use the Motor class with digital-only pins (which would only give you forwards / off / backwards controls rather than variable speed controls), in order to be able to use Motor objects with pins (e.g. on GPIO-expanders) that aren't capable of PWM (and are too slow for software PWM).
Whether that should be a new 'DigitalMotor' class, or some variant of the existing Motor class (that then raises Exceptions if you try to set a `value` other than 1 / 0 / -1) will need to be thought about. This also overlaps nicely with the discussion in #276 and https://github.com/RPi-Distro/python-gpiozero/pull/277#issuecomment-211608633

waveform80:
Why on earth did I write 0.01µF there... oh well, decimals on the brain :)

waveform80:
What on earth am I doing ... wrong comment on the wrong PR ... damnit, get a grip man!

lurch:
This has now been fixed with #303

Issue #300, opened 2016-04-28
Using L298N-Motor with gpiozero


goloplo:
Using this code on RPI2 with L298N-Motor:

```python
from gpiozero import Motor
from time import sleep

motor = Motor(forward=17, backward=27)

while True:
    motor.forward()
    sleep(5)
```

gives the following error:

```
> Traceback (most recent call last):
  File "motpi.py", line 3, in <module>
    motor = Motor(forward=17, backward=27)
  File "build/bdist.linux-armv7l/egg/gpiozero/devices.py", line 122, in __call__
  File "build/bdist.linux-armv7l/egg/gpiozero/output_devices.py", line 756, in __init__
  File "build/bdist.linux-armv7l/egg/gpiozero/devices.py", line 122, in __call__
  File "build/bdist.linux-armv7l/egg/gpiozero/output_devices.py", line 306, in __init__
  File "build/bdist.linux-armv7l/egg/gpiozero/pins/__init__.py", line 179, in <lambda>
  File "build/bdist.linux-armv7l/egg/gpiozero/pins/__init__.py", line 175, in _set_frequency
gpiozero.exc.PinPWMUnsupported: PWM is not supported on pin GPIO17
Exception AttributeError: AttributeError(u'no such attribute _source_thread',) in  ignored
```

If I use `LED(17)`, `LED(27)` is working perfect.

lurch:
I've just tried reproducing this myself on both Python2.7 and Python3.4 on each of a RPi2 and RPI3, using the latest version 1.2 of gpiozero, and in all cases your sample script worked without any problem :-/
Are you definitely using the latest v1.2 of python-gpiozero? How did you install it? Which version of Raspbian and/or Python are you using?

bennuttall:
That looks like it's not using RPi.GPIO - because its software PWM is permitted on any pin. It must be falling back to RPIO, or more likely the built-in pins implementation.

@goloplo Are you running this in a virtualenv? If so, you probably want to install RPi.GPIO in the virtualenv. If not, have you uninstalled RPi.GPIO?

@waveform80 can you tell from the stack trace which implementation is being used?

bennuttall:
@goloplo Note I've edited your post to use fenced code blocks to help make it more readable. Click edit and see how it's laid out for future reference.

lurch:
Just to keep things tied together - see my comments on #301 too.

lurch:
And see also #303 :-)
[...]

PR #302, opened 2016-04-28
Update test_outputs.py

goloplo:


waveform80:
Erm, not sure why you want to make this change - `Motor` *has* to use PWM-supporting pins (as the test failures make clear) and `MockPin` doesn't support PWM?

waveform80:
Ah, just read the rest of the discussion in #301, #300, etc. I'll close this for now as I think the (eventual) solution to #276, #277, et al. (i.e. #219) will probably encompass being able to construct a Motor made of DigitalOutputDevices.

PR #303, opened 2016-04-29
Add a `pwm` option to the RGBLED and Motor constructors

lurch:
...along with the other necessary changes required, to allow them to optionally be used with non-PWM-capable pins.

IMHO this is the 'proper' way of implementing the change suggested in #301


lurch:
Please merge #312 before merging this, and then I'll update this PR as appropriate.

waveform80:
#312 merged so this is now showing a conflict

lurch:
PR now rebased and updated.

It may _look_ like this PR is removing `test_output_pwm_pulse_background` and `test_output_pwm_pulse_foreground`, but in this morning's merge-fest they somehow got added twice (I guess I must have had slightly-different implementations of the same test function in two different PRs or something).

waveform80:
No prob - and yes, I think this probably the correct fix to #301 (though I still think #219 is a great idea too - I just don't have time to dig into it thoroughly right now).

waveform80:
Anyway, I'll merge this and we can consider #219 later when I'm back from picamera!

lurch:
Thanks for merging! Now to rebase all my other branches... ;)

PR #304, opened 2016-04-30
Minor Energenie tweaks

lurch:


waveform80:
Great stuff - incidentally, I'm not bothered about testing repr's (they're explicitly excluded in our coverage.cfg but codecov.io unfortunately doesn't take note of that). The reason for excluding them is that no-one should rely on the output of the repr as "useful to code"; they're simply meant to be human-readable object descriptions for debugging purposes.

lurch:
I only added the `repr` test to nudge the coverage towards 100% ;-)  And in adding the repr-test in #307 I actually uncovered a bug.
So perhaps the best approach would be to simply:
```python
assert type(repr(device)) == str
assert repr(device) != ''
```
which would pick up any reprs that throw exceptions or don't return strings?

waveform80:
Actually, yeah - that's a decent idea. I recently came across a repr that threw an exception in the new picamera mmalobj layer so I probably ought to at least that.

PR #306, opened 2016-04-30
Allow Robot to turn-while-moving

lurch:
by adding an optional `curve` parameter to the forward() and backward() methods.

I also added a few more Robot test-cases while I was at it ;-)

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -794,27 +794,55 @@ class Robot(SourceMixin, CompositeDevice):
     def value(self, value):
         self.left_motor.value, self.right_motor.value = value
 
-    def forward(self, speed=1):
+    def forward(self, speed=1, curve=0):
         """
         Drive the robot forward by running both motors forward.
 
         :param float speed:
             Speed at which to drive the motors, as a value between 0 (stopped)
             and 1 (full speed). The default is 1.
-        """
-        self.left_motor.forward(speed)
-        self.right_motor.forward(speed)
 
-    def backward(self, speed=1):
+        :param float curve:
+            The amount to curve left (if ``curve`` is less than 0) or curve
+            right (if ``curve`` is greater than 0) while moving forwards, by
+            driving one of the motors at a slower speed. Maximum left curve is
+            -1, maximum right curve is 1. The default is 0 (no curve).
+        """
+        left_speed = speed
+        right_speed = speed
+        if not -1 <= curve <= 1:
+            raise ValueError('curve must be between -1 and 1')
+        if curve < 0:
+            left_speed *= 1 + curve
+        elif curve > 0:
+            right_speed *= 1 - curve
+        self.left_motor.forward(left_speed)
+        self.right_motor.forward(right_speed)
+
+    def backward(self, speed=1, curve=0):
         """
         Drive the robot backward by running both motors backward.
 
         :param float speed:
             Speed at which to drive the motors, as a value between 0 (stopped)
             and 1 (full speed). The default is 1.
+
+        :param float curve:
+            The amount to curve left (if ``curve`` is less than 0) or curve
+            right (if ``curve`` is greater than 0) while moving backwards, by
+            driving one of the motors at a slower speed. Maximum left curve is
+            -1, maximum right curve is 1. The default is 0 (no curve).
         """
-        self.left_motor.backward(speed)
-        self.right_motor.backward(speed)
+        left_speed = speed
+        right_speed = speed
+        if not -1 <= curve <= 1:
+            raise ValueError('curve must be between -1 and 1')
+        if curve < 0:
+            left_speed *= 1 + curve
+        elif curve > 0:
+            right_speed *= 1 - curve
+        self.left_motor.backward(left_speed)
+        self.right_motor.backward(right_speed)
 
     def left(self, speed=1):
         """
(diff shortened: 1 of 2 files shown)


waveform80:
It's good stuff (love the extra robot tests!), and the idea is great. My one concern is whether we can come up with a slightly better API. Specifically, I'm not sure -1=left and +1=right on `forward()` and `backward()` is that intuitive or easy to think about.

Perhaps it'd be better to incorporate this functionality into the `left()` and `right()` methods? Something like an optional curve parameter there which as it varies between 0..1 (or -1..+1) affects whether one motor works in total opposition to the other or in the same direction at a different speed? Just seems a little easier to comprehend than assigning directions to numeric signs (however intuitive it may be for us thinking about a number line running left-to-right, negative-to-positive).

lurch:
I dunno... is -1 for left and +1 for right any less confusing than -1 for backwards and +1 for forwards? ;-)

At the moment, `forward()` does forwards-in-a-straight-line, `backward()` does backwards-in-a-straight-line, `right()` does rotate-right-on-the-spot, and `left()` does rotate-left-on-the-spot.
IMHO it seemed more natural to extend `forward()` to include turn-left-while-moving-forwards, than it did to extend `left()` to mean move-forwards-while-turning-left. \*shrug\*

And the reason I added a -1..+1 `curve` parameter, is that I was imagining it might be fun in future to have a robot controlled by a slider-potentiometer for the speed (with the 0..1 of the slider `scaled` to -1..+1 for both backwards & forwards) and a rotary-potentiometer for the 'amount of curve' i.e. left/right (again with the 0..1 of the pot scaled to -1..+1 for both left and right). I.e. similar to a traditional RC-car controller ([this](https://github.com/RPi-Distro/python-gpiozero/pull/277#issuecomment-213119691) is what gave me the original idea for this PR).

But if the consensus is that `forward(speed=1, curve=-0.5)` meaning turn-left-while-moving-forwards and `forward(speed=1, curve=0.5)` meaning turn-right-while-moving-forwards is too confusing (see also the docstrings I added in the PR); what about `forward(speed=1, curve_left=0.5)` and `forward(speed=1, curve_right=0.5)` ? (with an Exception raised if both `curve_left` and `curve_right` are non-zero)

Any comments? @bennuttall @martinohanlon @MrHarcombe @goloplo @SteveAmor

MrHarcombe:
I see what @waveform80 means by confusion - having two -1->1 ranges in the same parameter list could confuse those, like myself, bears of very little brain.

But I do like your idea of separate named parameters for the curve `left` xor `right`. I haven't checked your code, but presume there would be a similar addition to `reverse`? That would be a useful method for those messing with distance sensors, perhaps - if you detect an obstruction, reverse and curve then try to proceed. (on a tangent... I seem to recall having a remote controlled car as a child that, to me, confusingly had forwards, but then only a reverse-and-turn button).

lurch:
Hmmm, I hadn't added 'curve' to `reverse()` but I had added it to `backward()` (but perhaps that's what you meant?).

With the way the code works currently, if you did `forward(speed=0.8, curve=0.4)` and then did `reverse()` then the combined effect will be identical to `backward(speed=0.8, curve=0.4)` - I just checked. Which at least to me seems to make perfect sense. Which is purely by luck, as I hadn't considered the case of calling `reverse()` while driving in a curve ;-) Once we decide what the desired behaviour should be, I'll add a unit-test for it.

> I seem to recall having a remote controlled car as a child that, to me, confusingly had forwards, but then only a reverse-and-turn button).

Yeah, I think my younger brother once had one like that - IIRC it just had a single drive wheel on a spring-mounted pivot-plate.
The first RC car I had, had one backwards-speed, two forwards speeds (normal and "TURBO", and a couple of electromagnets* to steer the front wheels. (Haha, it was the red version of [this](https://rctoymemories.com/2014/06/03/nikko-black-fox-1985/))

\* yes, at some point I took it apart, and I don't think it worked properly afterwards.

bennuttall:
I just tested this and it's an awesome feature. We just need to decide on the API and let's get it merged. Love it.

https://vine.co/v/i3UpK2AVHMv

I personally don't mind the proposal as is, but I guess `curve_left` and `curve_right` do make more sense. @lurch do you want to update the PR?

bennuttall:
(trying to assign both @lurch and @waveform80, just waiting on @lurch's team membership)

lurch:
Okay... the branch in this PR (`robot_curve`) has been rebased from master, and I've made a tweak to the way the Robot tests are done.
On `robot_curve2` https://github.com/lurch/python-gpiozero/tree/robot_curve2 there's an extra commit which splits the single -1 -> +1 `curve` parameter into two separate 0 -> +1 `curve_left` and `curve_right` keyword-only parameters. Have a look, and let me know if you'd like me to push that extra commit onto this PR too.

(and I've just realised that I forgot to add the [reverse-while-curving](https://github.com/RPi-Distro/python-gpiozero/pull/306#issuecomment-217749798) tests)

waveform80:
Looks good to me - if you want me to fix it up for merge I can do that quickly (only the tests need tweaking). The one thing I'd add (which I can during the rebase) is a test to throw an exception in the case of unrecognized kwargs (so that if people accidentally do something like `forward(1, curveleft=0.5)` they get an exception instead of silently ignoring the invalid keyword).

bennuttall:
Sounds good :+1:

waveform80:
Merged in #619

PR #318, opened 2016-05-09
Mark optional sleeps with no coverage

waveform80:
Codecov frequently whinges about "unexpected coverage changes" on PRs;
this is because the sleep lines affected sometimes execute and sometimes
don't (they're only there to ensure something exists before we test it).
Hopefully codecov notices the pragmas...


lurch:
Hurrah! :+1:

waveform80:
There's probably a few more of these to deal with, and I'm still not 100% sure codecov notices those pragmas - we'll just have to see on future PRs, but it'd be nice to squash those warnings.

Issue #315, opened 2016-05-07
UnidirectionalMotor suggestion


lurch:
As I suggested in https://github.com/RPi-Distro/python-gpiozero/pull/276#issuecomment-217633718 I just had a go at coding up a UnidirectionalMotor which e.g. uses a wiring diagram like http://raspi.tv/2013/how-to-use-soft-pwm-in-rpi-gpio-pt-2-led-dimming-and-motor-speed-control

It would seem to make sense to have the UnidirectionalMotor class be the parent of the current (bidirectional) Motor class, since a bidirectional motor adds extra functionality on top of a unidirectional motor.
I managed to code it up and get it working (and passing the associated unit-tests), but I'm not really happy with the way the `__init__` parameters of the two classes now work with the interaction of CompositeDevice, which is why I'm only submitting this as an Issue and not a PullRequest. https://github.com/lurch/python-gpiozero/tree/unidirectional_motor

Is there a neater way of doing what I'm trying to do here, while still having Motor be a subclass of UnidirectionalMotor?
(Maybe there simply *isn't* a 'nice' way for CompositeDevice / CompositeOutputDevice subclasses to have different numbers of subordinate devices?)

.......
Alternatively, given that UnidrectionalMotor isn't *really* a CompositeDevice (as it only has a single subordinate device) maybe a different approach would be to have a UnidirectionalMotorMixin, and then UnidirectionalMotor could inherit from both UnidirectionalMotorMixin and PWMOutputDevice. And then we could have a MotorMixin which inherits from UnidirectionalMotorMixin, and Motor could inherit from both MotorMixin and CompositeDevice. And then all the other 'Motor' classes ( #276 #300 #314 etc.) could also inherit from MotorMixin, with Robot then being re-written to operate on MotorMixin instances?

lurch:
...and jumping ahead of myself slightly, I guess if we went for the Mixin approach, then https://github.com/lurch/python-gpiozero/commit/501ee0273ae00a1d9f0ed88dc4bf9680fa9d9f6f#diff-87f09163a5a18c26559f2252fbd91abbR226 could be modified to 
```python
    @classmethod
    def construct_object(cls, value, _interface_class=None, **kwargs):
        if (_interface_class is not None and isinstance(value, _interface_class)) or isinstance(value, cls):
            # if we already have an object of the right class _and_ we're
            # given kwargs, we ignore the kwargs and assume the object has
            # already been properly constructed
            return value
        else:
[...]

Issue #212, opened 2016-02-20
'state' property inconsistency


lurch:
According to http://gpiozero.readthedocs.org/en/latest/api_pins.html#gpiozero.pins.Pin.state the `state` property is 0 or 1 (i.e. an integer), or between 0.0 and 1.0 (i.e. a float) for PWM outputs.
However both `NativePin` and `MockPin` (I haven't looked at all the other `Pin` implementations yet) treat it as a boolean value. Indeed, `NativePin._get_state` says `return bool(...)`.
Is it worth modifying the `_get_state` implementations so that they match the documentation and always return a numeric value?

Feel free to close if you consider this irrelevant since the True/False and 1/0 values automatically coerce.

Also, it seems that some `Pin` implementations (e.g. NativePin) allow setting a `state` of e.g. 0.5 when PWM isn't in use (automatically converting it to 1 / True) whereas other implementations (e.g. RPiGPIOPin) throw a `PinInvalidState`. Could the documentation be tightened up on what the expected behaviour is please? :)  Or will this remain an "implementation-specific detail"?

waveform80:
Personally I'm not too bothered about the 0/1/False/True thing since as you note they automatically coerce (right down to the point that you can do things like indexing an array with a bool, e.g. ['No', 'Yes'][True]). That said, you're absolutely right that there's a certain inconsistency of treatment of other things like floats - and the docs don't really specify the expected behaviour. Yeah - let's come up with a good specification first and then tighten up the implementations to match it.

bennuttall:
Is this still relevant?

PR #319, opened 2016-05-10
Add tests for LEDBoard.active_high and LEDBarGraph.active_high

lurch:


lurch:
PR has been updated to also test LEDBoard.value too.
However in doing so I noticed a (possible) slight inconsistency, similar to #212. If we create an LEDBoard with `pwm=True` then only values of 0..1 are accepted, but if we create an LEDBoard with `pwm=False` (which is the default) then *any* values are allowed, as they automatically get coerced to bools. I.e.
```python
>>> from gpiozero.pins.mock import MockPin, MockPWMPin
>>> from gpiozero import LEDBoard
>>> pin1 = MockPin(2)
>>> pin2 = MockPin(3)
>>> pin3 = MockPin(4)
>>> pin4 = MockPWMPin(5)
>>> pin5 = MockPWMPin(6)
>>> pin6 = MockPWMPin(7)
>>> dboard = LEDBoard(pin1, pin2, pin3)
>>> pboard = LEDBoard(pin4, pin5, pin6, pwm=True)
>>> dboard.value = (0, -1, 2)
>>> dboard.value
LEDBoardValue(_0=False, _1=True, _2=True)
>>> pboard.value = (0, -1, 2)
  File "/home/andrew/tmp_working/python-gpiozero/gpiozero/devices.py", line 299, in __setattr__
    return super(CompositeDevice, self).__setattr__(name, value)
  File "/home/andrew/tmp_working/python-gpiozero/gpiozero/devices.py", line 146, in __setattr__
    return super(GPIOBase, self).__setattr__(name, value)
  File "/home/andrew/tmp_working/python-gpiozero/gpiozero/boards.py", line 81, in value
    device.value = v
  File "/home/andrew/tmp_working/python-gpiozero/gpiozero/devices.py", line 146, in __setattr__
    return super(GPIOBase, self).__setattr__(name, value)
  File "/home/andrew/tmp_working/python-gpiozero/gpiozero/output_devices.py", line 350, in value
    self._write(value)
  File "/home/andrew/tmp_working/python-gpiozero/gpiozero/output_devices.py", line 332, in _write
    raise OutputDeviceBadValue("PWM value must be between 0 and 1")
gpiozero.exc.OutputDeviceBadValue: PWM value must be between 0 and 1
```

lurch:
Alternatively, if we decide that "they automatically get coerced to bools" **is** the expected behaviour, then maybe I ought to modify the changes I made in #303 to also coerce-any-values-to-bools when `pwm=False` is being used; rather than explicitly only allowing 0 or 1 as it does now?  \*shrug\*

waveform80:
Hmmm ... good question. I'm rather tempted to keep the coerced to bools behaviour simply because I tend to think of "pure" digital devices (i.e. non-PWM) as having a bool value (rather than numeric) in which case anything "[truthy](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)" is acceptable as an input. However, PWM and analog devices necessarily have a numeric value; while we may default to float on output we should still accept any numeric input (Fraction, int, etc.) provided it's in range.

Any counter arguments?

lurch:
> Any counter arguments?

Nope, it's a perfectly valid argument.
And I guess modifying the "pure" digital device unit-tests (including `Motor`s constructed with `pwm=False`) to read and write True and False (instead of 0 and 1) would help make this more obvious?

waveform80:
Yeah - I must admit the tests using 0/1 for the digital devices were pure laziness on my part (trying to plough through stuff as fast as possible!).

Do you want me to hold off on merging this for the moment?

lurch:
> Do you want me to hold off on merging this for the moment?

Nah, this PR can be merged as-is. I'll just need to create another PR (following on from the now-merged #303) in which I 'undo' the modifications where I check that non-PWM values are only `0` or `1`.
Changing the 'digital' unit tests to True/False instead of 1/0 should be an entirely different PR.

One possible 'wrinkle' I've just thought of (while cleaning my teeth!), is what to return for devices that can be both 'digital' and 'PWM'...
If we have e.g. a UnidirectionalMotor (as suggested in #315) composed of a single PWMOutputDevice (i.e. constructed with the default `pwm=True`), then the result of the `value` property-getter will always be a float between 0 and 1. And the input to the `value` property-setter will also always be a float between 0 and 1.
Whereas if we have a UnidirectionalMotor constructed with `pwm=False` (i.e. composed of a single DigitalOutputDevice), then the input to the `value` property-setter is effectively a boolean value. However to maintain compatibility with the pwm-able UnidirectionalMotor, we should probably still return "a float between 0 and 1" from the `value` property-getter (i.e. `0.0` for 'off' and `1.0` for 'on'). Does it make sense for the value-getter to return a float, but the value-setter to (implicitly) operate on bools?

Maybe I'm just worrying about nothing, given the auto-conversion that Python does between bool/int/float ? Still, good to get the expected / defined behaviour fully hammered out :-)


(And I guess `Motor`s constructed with `pwm=False` will still need to operate with the float values (for both setting and getting) of `-1.0`, `0`, `1.0`, since a bool can't represent the three different states. So I guess setting a Motor with `pwm=False` to `-1.0` will drive it backwards, whereas setting a UnidirectionalMotor with `pwm=False` to `-1.0` will drive it forwards (remember that Motor is a subclass of UnidirectionalMotor). Hmmm, maybe I need to think about this more tomorrow when I'm not so tired!)

lurch:
...although even for `Motor`s with `pwm=False`, the optional argument to `forward()` and `backward()` _could_ still be coerced to bool, making `motor.forward(-2)` === `motor.forward(True)` === `motor.forward()`.

Still not sure what the most reasonable behaviour is, so happy to go along with whatever @waveform80 recommends. But whatever behaviour we *do* settle on, I think RGBLED, Motor, UnidirectionalMotor, LEDBoard, etc. should probably all work in the same way (in terms of value setters and getters). Currently [RGBLED](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L605) with `pwm=False` works differently to [LEDBoard](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/boards.py#L78) with `pwm=False`; the former only allows a tuple of `0` / `1` (or `False` / `True`) values but the latter accepts a tuple of any values, coercing each of them to bool. (Whereas when constructed with `pwm=True`, they both only accept a tuple of (values that can be coerced to) floats between 0 and 1 inclusive).

Apologies if I'm unnecessarily overcomplicating things :-/

waveform80:
> Apologies if I'm unnecessarily overcomplicating things :-/

No, they're good points - I'll need to think a bit about this but I'm currently hip-deep in a gdb backtrace in picamera that's segfaulted somewhere in libffi (god help me), so I'll have to leave it for the weekend. Anyway, I'll merge this now :)

PR #324, opened 2016-05-11
Always make PWMOutputDevice operate on floats.

lurch:
Before this patch, the following script:
```python
from gpiozero.pins.mock import MockPin, MockPWMPin
from gpiozero import DigitalOutputDevice, PWMOutputDevice
PinClass = MockPWMPin
for DeviceClass in (DigitalOutputDevice, PWMOutputDevice):
    print(DeviceClass)
    pins = []
    pins.append(DeviceClass(PinClass(2)))
    pins.append(DeviceClass(PinClass(3), initial_value=False))
    pins.append(DeviceClass(PinClass(4), initial_value=0))
    pins.append(DeviceClass(PinClass(5), initial_value=0.0))
    pins.append(DeviceClass(PinClass(6), active_high=False))
    pins.append(DeviceClass(PinClass(7), active_high=False, initial_value=False))
    pins.append(DeviceClass(PinClass(8), active_high=False, initial_value=0))
    pins.append(DeviceClass(PinClass(9), active_high=False, initial_value=0.0))
    print([x.value for x in pins])
    for value in (False, 0, 0.0, True, 1, 1.0):
        pins[0].value = value
        pins[4].value = value
        print([x.value for x in (pins[0], pins[4])])
    for pin in pins:
        pin.close()
    MockPin.clear_pins()
```
results in:
```
<class 'gpiozero.output_devices.DigitalOutputDevice'>
[False, False, False, False, False, False, False, False]
[False, False]
[False, False]
[False, False]
[True, True]
[True, True]
[True, True]
<class 'gpiozero.output_devices.PWMOutputDevice'>
[False, False, False, False, 0.0, 0.0, 0.0, 0.0]
[False, 0.0]
[False, 0.0]
[False, 0.0]
[1.0, 1.0]
[1.0, 1.0]
[1.0, 1.0]
```

whereas after this patch, the same script results in:
```
<class 'gpiozero.output_devices.DigitalOutputDevice'>
[False, False, False, False, False, False, False, False]
[False, False]
[False, False]
[False, False]
[True, True]
[True, True]
[True, True]
<class 'gpiozero.output_devices.PWMOutputDevice'>
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
[0.0, 0.0]
[0.0, 0.0]
[0.0, 0.0]
[1.0, 1.0]
[1.0, 1.0]
[1.0, 1.0]
```
which IMHO is much more consistent. This is sort-of a move in the same direction as #212 

While I was at it, I also added better 'encapsulation' of the active_high properties of both OutputDevice and PWMOutputDevice, and I made a small tweak to the init-method of PWMOutputDevice.

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -43,16 +43,15 @@ class OutputDevice(SourceMixin, GPIODevice):
         self.active_high = active_high
         if initial_value is None:
             self.pin.function = 'output'
-        elif initial_value:
-            self.pin.output_with_state(self._active_state)
         else:
-            self.pin.output_with_state(self._inactive_state)
+            self.pin.output_with_state(self._value_to_state(initial_value))
+
+    def _value_to_state(self, value):
+        return bool(self._active_state if value else self._inactive_state)
 
     def _write(self, value):
-        if not self.active_high:
-            value = not value
         try:
-            self.pin.state = bool(value)
+            self.pin.state = self._value_to_state(value)
         except AttributeError:
             self._check_open()
             raise
@@ -318,23 +317,16 @@ class PWMOutputDevice(OutputDevice):
             pass
         super(PWMOutputDevice, self).close()
 
-    def _read(self):
-        self._check_open()
-        if self.active_high:
-            return self.pin.state
-        else:
-            return 1 - self.pin.state
+    def _state_to_value(self, state):
+        return float(state if self.active_high else 1 - state)
+
+    def _value_to_state(self, value):
+        return float(value if self.active_high else 1 - value)
 
     def _write(self, value):
-        if not self.active_high:
-            value = 1 - value
         if not 0 <= value <= 1:
             raise OutputDeviceBadValue("PWM value must be between 0 and 1")
-        try:
-            self.pin.state = value
-        except AttributeError:
-            self._check_open()
-            raise
+        super(PWMOutputDevice, self)._write(value)
 
     @property
     def value(self):
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -376,9 +376,12 @@ class GPIODevice(Device):
         self._active_state = True
         self._inactive_state = False
 
+    def _state_to_value(self, state):
+        return bool(state == self._active_state)
+
     def _read(self):
         try:
-            return self.pin.state == self._active_state
+            return self._state_to_value(self.pin.state)
         except (AttributeError, TypeError):
             self._check_open()
             raise


gpiozero/output_devices.py
```diff
@@ -300,7 +299,10 @@ def __init__(self, pin=None, active_high=True, initial_value=0, frequency=100):
         self._controller = None
         if not 0 <= initial_value <= 1:
             raise OutputDeviceBadValue("initial_value must be between 0 and 1")
-        super(PWMOutputDevice, self).__init__(pin, active_high)
+        if initial_value in (0, 1):
+            super(PWMOutputDevice, self).__init__(pin, active_high, initial_value=initial_value)
```
waveform80: This is a little redundant; when `pin.frequency` is set below that starts the pin implementation's PWM engine (whatever it may be). These all default to a duty cycle of "off" so if the initial-value is 1, the pin'll just be driven back to 0 before we set value on line 309 (hence the comment about needing a method for setting frequency and value together).

waveform80:
Other than the minor note on initial-value in PWMOutputDevice - this is much cleaner, very nice! I'm happy to merge it if you want or I can hold off if you want to look at it a little more.

gpiozero/output_devices.py
lurch: Okay. So when the Pin classes *do* get fixed to set the frequency-and-value together (e.g. a `pwm_with_state` function, analogous to `output_with_state`), then I guess there'll also need to be some mechanism added to stop the parent OutputDevice class either calling `output_with_state` or setting the `function` property?

bennuttall: > Okay. So when the Pin classes do get fixed to set the frequency-and-value together

What do you mean?

lurch:
PR updated :)

gpiozero/output_devices.py
waveform80: @bennuttall At the moment the pin classes have some "utility" methods like `output_with_state` to facilitate configuring pins in a given state as fast as possible (only possible with some underlying implementations but helps prevent "power blips" on component construction - in some use-cases this is important). However, there currently isn't an equivalent facility for the PWM side of things - so setting pin frequency always results in the pin switching off (however briefly) at the moment. I need to add another utility method for this.

waveform80: @lurch precisely - and that's also why I haven't added one yet - I vaguely recall trying it a while back and realizing it wasn't a 30 second job so I left it for the time being (on the assumption that, given we only had software PWM at the time anyway blips were par for the course anyway :)

waveform80:
Lovely stuff.

gpiozero/output_devices.py
lurch: > What do you mean?

The way it works currently, is if you create a PWMOutputDevice with an intial_value of e.g. 1.0, then it'll first call the super() init-method in OutputDevice, which has a default initial_value of False. So the pin will get set to (digital) output-mode, with a value of 0. Then back in the PWMOutputDevice init-method we set the frequency property to the required value, which 'turns on' PWM mode, and again sets the value to 0.0. And finally in the PWMOutputDevice init-method we set the desired value, which eventually gets our pin operating at the correct value.

Ideally, the pin wouldn't enter 'digital' output mode, and wouldn't have a PWM-value of 0, but would immediately switch from 'input' to outputting the desired PWM-value.
(Yes, for all practical limitations there's no *actual* difference, but maybe there's a theoretical PWM device which should *never* have it's output driven to 0, no matter for how short a time-period? Possibly I'm just being overly-fussy.)

lurch: > I vaguely recall trying it a while back and realizing it wasn't a 30 second job so I left it for the time being

Doing a quick grep in the sourcecode, it seems the only child-classes of OutputDevice are DigitalOutputDevice, PWMOutputDevice and SPISoftwareInterface.
So I guess one approach might be to make OutputDevice an abstract base class (see http://gpiozero.readthedocs.io/en/latest/api_output.html#base-classes) which isn't directly instantiable (which will mean tweaking a whole bunch of tests), and then move `output_with_state` to only be in the DigitalOutputDevice constructor?  I may have a play later this evening...
(Hmmm, I wonder if `_active_state` and `_inactive_state` could to be moved to only be in DigitalOutputDevice (and DigitalInputDevice) as well?)

PR #331, opened 2016-05-26
Add ADC.voltage property, fix #329

### LINKED ISSUE #329: Add ADC.voltage property?
While looking at the example code shown on https://www.kickstarter.com/projects/raspitv/raspio-analog-zero-read-8-sensors-at-once-on-raspb I wondered if it might be worth adding an optional `max_voltage` property to the `AnalogInputDevice` constructor (defaulting it to `3.3`) which would then allow us to add a `voltage` property, which would simply return `self.value * self._max_voltage`, and then the example code would then become:

``` python
from gpiozero import MCP3008
adc = MCP3008(channel=0)
print("measured voltage is: ", adc.voltage)
```

i.e. it'd work very similar to the `distance` / `max_distance` properties of the current `DistanceSensor` class.  And allowing `max_voltage` to be set (rather than fixing it at 3.3V) would allow you to adjust it as necessary (to get the 'true value') if using the MCP3008 with a voltage-divider to bring a higher-voltage down into the 3.3V range input of the MCP3008. (Or if using a lower voltage than 3V3 on the Vref input pin)

Good idea / bad idea? 
Or perhaps this should be added to `MCP3xxx` rather than `AnalogInputDevice` given that (strictly speaking) a generic analog-input-device might not necessarily be reading voltage? (is that possible / likely?)

Ping @raspitv  ;-)

bennuttall:


lurch:
Should `max_voltage` also be exposed as a property?

bennuttall:
Do you mean a read-only property or read/write?

I considered this, but thought it unlikely people would set up a device, then change its max_voltage.

I think read-only makes sense.

bennuttall:
In which case I'll update this PR.

lurch:
Yes, a read-only property, just like `bits`.  And while you're at it, I guess it wouldn't hurt to check inside `__init__` that `max_voltage > 0` ?
And does the `max_voltage` parameter to `__init__` need documenting?

bennuttall:
I thought of that too, but wondered if there was a more sensible lower limit? Can it be any positive number? I'm only aware of 3V3 and 5V used in these cases.

bennuttall:
Done. Where should the new init param be documented?

lurch:
re: max_voltage

> Can it be any positive number?

Yes, as I mentioned in #329 you might be using a voltage-divider (e.g. [like this](http://raspi.tv/2015/raspio-duino-as-a-lipo-monitor) ) and so by specifying the appropriate max_voltage, you can get adc.voltage reporting the __real__ voltage that you're measuring, and not just the voltage being measured by the ADC. IYSWIM.

waveform80:
For documentation: take a look at `docs/api_spi.rst`. Specifically, you're modifying the base class AnalogInputDevice for which all members are included (see bottom of that file) so this should appear automagically given you've provided a doc-string for it. If you want to include it in the sub-classes though, you'll want to add it to the various `:members:` lists in the classes above (as in other cases, we limit the methods and properties for concrete classes for the sake of brevity).

waveform80:
BTW ... shouldn't this parameter be added to all the sub-classes? Otherwise ... it's just going to assume 3.3 for all of them.

lurch:
> BTW ... shouldn't this parameter be added to all the sub-classes? Otherwise ... it's just going to assume 3.3 for all of them.

I haven't tested it, but I assumed that specifying it as a keyword-only-arg meant it'd transparently get passed up-the-chain via the `**spi_args` ?

waveform80:
Hmmm, it will indeed be passed up by `**spi_args` however there's a couple of issues with that:

1. It's not an SPI-related arg (okay, purely semantic issue - doesn't really affect operation).
2. It means max_voltage won't appear in any of the documented constructors despite the fact it can be included.

PR #333, opened 2016-05-26
Add LedBorg as a subclass of RGBLED

lurch:
Note that this won't work properly until #327 has been merged.


lurch:
PR updated to add unit-test.

lurch:
...and now with updated output_device_hierarchy images :-)

PR #338, opened 2016-05-28
Change LEDBoard.leds to return a static tuple

### LINKED ISSUE #337: Slicing LEDBoard
It seems the LEDBoard `.leds` property is now a generator, rather than a list. Great for iterating over but I have to cast it to a list to be able to slice it:

```python
from gpiozero import PiLiter

lite = PiLiter()

for led in list(lite.leds)[::2]:
    led.on()
```

Is there a better way?

lurch:
fixes #337

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -116,19 +116,22 @@ class LEDCollection(CompositeOutputDevice):
                 LEDClass(pin_or_collection, active_high, initial_value)
                 for name, pin_or_collection in kwargs.items()
                 })
+        leds = []
+        for item in self:
+            if isinstance(item, LEDCollection):
+                for subitem in item.leds:
+                    leds.append(subitem)
+            else:
+                leds.append(item)
+        self._leds = tuple(leds)
 
     @property
     def leds(self):
         """
-        A flat iterator over all LEDs contained in this collection (and all
+        A flat tuple of all LEDs contained in this collection (and all
         sub-collections).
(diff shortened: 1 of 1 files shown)


lurch:
On a side-note: looking at the guts of CompositeDevice I guess `self._named` really ought to be a frozendict - is it worth pulling something like the MIT-licensed https://pypi.python.org/pypi/frozendict into compat.py and using that to 'freeze' `self._named` ?
(`self._all` is already frozen due to it being a tuple rather than a list)

waveform80:
Hmm ... it *should* be a frozendict (or something similar), but while frozendict is on PyPI, it's not packaged for Raspbian, and I'm also not entirely sure that adding a dependency is actually worth it to fix this (I generally consider adding a dependency to a package a big deal: if it's for novel functionality, or really critical, or security based, then sure - if it's for what amounts to a pretty small bug fix I generally either skip it or duplicate the bare minimum of functionality needed - constructing a read-only mapping is pretty trivial).

lurch:
Yeah, I wasn't suggesting adding it as a dependency, but instead copy-n-pasting the relevant parts into `compat.py` (which we're free to do, since it's MIT-licensed). In other words exactly your "duplicate the bare minimum of functionality needed" - sorry if I didn't word that very clearly in my previous comment.

waveform80:
Ah, in that case - by all means, go for it :)

PR #339, opened 2016-05-28
Add extra init-method params

lurch:
adds `initial_value` to PiLiter, TrafficLights, PiTraffic & SnowPi and
adds `pwm` to PiTraffic

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -372,8 +372,8 @@ class LEDBarGraph(LEDCollection):
 
     :param float initial_value:
         The initial :attr:`value` of the graph given as a float between -1 and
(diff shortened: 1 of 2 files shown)


lurch:
See https://github.com/RPi-Distro/python-gpiozero/commit/a116e9ded01b5f8bd34891a2dd40450748a18d12#commitcomment-17654093

The only slight 'wrinkle' is that all the `initial_value` params are documented as `bool`, but as shown by `test_snow_pi_initial_value_pwm` if a device is constructed with `pwm=True` then the `initial_value` can be a float between 0 and 1. But I wasn't sure if / how to add this to the docs without overcomplicating it. See also the discussions in #319

PR #341, opened 2016-05-28
remove usage of CompositeDevice.all since it's marked as deprecated

lurch:
https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/devices.py#L321

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -52,7 +52,7 @@ class CompositeOutputDevice(SourceMixin, CompositeDevice):
         """
         Turn all the output devices on.
         """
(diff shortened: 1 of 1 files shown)


bennuttall:
Should we not leave it for a while to save BC, and remove later (say, v2.0?) @waveform80?

lurch:
Yes, I agree. I didn't remove the function itself, I only updated other parts of gpiozero to no longer call it ;-)

bennuttall:
Oh I see. Yeah that's cool.

waveform80:
Yup, that's perfect :)

PR #352, opened 2016-06-03
Update sourcetools examples

lurch:
Remove blue.source_delay, as it's actually red.source_delay that controls the speed of the fading

### DIFF
--- a/gpiozero/tools.py
+++ b/gpiozero/tools.py
@@ -518,7 +518,6 @@ def sin_values(period=360):
         red = PWMLED(2)
         blue = PWMLED(3)
         red.source_delay = 0.01
-        blue.source_delay = 0.01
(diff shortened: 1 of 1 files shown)


bennuttall:
Is that not intentional?

waveform80:
Hmm, sort of. As it happens both are redundant as the default delay is actually 1/100th of a second, so they're really just there for show. Still, it's wise to specify both as the blue thread's delay could be *longer* than the red thread's at which point one LED would appear to fade "choppily" ... if that's a word.

lurch:
Fair enough. In spirit of DRY, what if we instead changed it to `blue.source_delay = red.source_delay` ?

waveform80:
Sounds good to me!

lurch:
PR updated :)

PR #355, opened 2016-06-05
Add extra product URLs to the boards.py docstrings

lurch:


bennuttall:
So this would work in the docs pages but is that the way they should be presented in docstrings?

I've no problem with it - just checking it's the *right way*.

We should get "What would @waveform80 do?" bracelets...

lurch:
Dunno - I was just copying from the other examples in that file. See e.g. http://gpiozero.readthedocs.io/en/latest/api_boards.html#snowpi

Although to make sure that these links don't go stale, maybe we could add an automatic-link-checker to the Travis build, just like is done on https://github.com/raspberrypi/documentation ?

bennuttall:
Ah ok, I was hoping there'd be a case we could refer to. Settled.

waveform80:
Yeah, doc-strings (when accessed via the help() function or via ipython's notebook interface) are generally just rendered as raw reST (which at least is pretty readable). For this reason, in doc-strings I tend to cluster all links at the bottom of the doc-string as if they were footnotes (similar to `elinks -dump` output).

This tends to be pretty readable at the console - you can read through the text and if you want to look up any links you know to find them all at the bottom, and the (usually lengthy) URLs don't interfere with your reading (if I know markdown is likely to be read raw, I likewise switch to its footnote link-style as it's much easier to  read the prose with).

bennuttall:
I didn't know they rendered as RST in IPython Notebook. That's handy.

Issue #353, opened 2016-06-03
Better pigpio support


dglaude:
I believe the pigpio support should respect pigpio environmental variable.

Current definition of PiGPIOPin function default to host='localhost' and port=8888.

PIGPIO Python library way of working is to check the environment variable PIGPIO_ADDR/PIGPIO_PORT

If any of those is set, it is also a clear indication that PIGPIO is available... and how to access those PINs

I also believe pigpio should automatically be used if local GPIO are not found and any of those environmental value is found.

The goal is to have the same code working whatever if the pigpiod is running locally or remotely.
Actually transparent use of pigpiod is the ultimate goal, this mean without having to specify PiGPIOPin.

See below for an experiment:

Please note that this code is working when changing the led and button definition the following way:
```
amber = LED(PiGPIOPin(22,'slave.local'))
button = Button(PiGPIOPin(7,'slave.local'), pull_up=False)
```

The experiment setup
- All command typed from a Mac
- PiZero connected with USB in Ethernet Gadget mode (hostname is 'slave')
- pigpiod started as root on the PiZero
- FishDish connected to the PiZero

**dglaude$** env | grep PIGPIO
PIGPIO_ADDR=slave.local
**dglaude$** cat gpiozero.pigpio.FishDish.py 
```
from gpiozero.pins.pigpiod import PiGPIOPin
from gpiozero import LED, Button
from signal import pause

amber = LED(PiGPIOPin(22))
button = Button(PiGPIOPin(7), pull_up=False)

button.when_pressed = amber.on
button.when_released = amber.off

pause()
```
**dglaude$** python gpiozero.pigpio.FishDish.py 
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
Can't connect to pigpio at localhost(8888)

Did you start the pigpio daemon? E.g. sudo pigpiod

Did you specify the correct Pi host/port in the environment
variables PIGPIO_ADDR/PIGPIO_PORT?
E.g. export PIGPIO_ADDR=soft, export PIGPIO_PORT=8888

Did you specify the correct Pi host/port in the
pigpio.pi() function? E.g. pigpio.pi('soft', 8888))
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
Traceback (most recent call last):
[...]

PR #356, opened 2016-06-05
Fix a typo (see https://github.com/RPi-Distro/python-gpiozero/issues/…

dglaude:
see #353

### DIFF
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -43,7 +43,7 @@ except ImportError:
         pin_factory = RPIOPin
     except ImportError:
         try:
-            from .pins.pigipod import PiGPIOPin
+            from .pins.pigpiod import PiGPIOPin
             pin_factory = PiGPIOPin
         except ImportError:
             from .pins.native import NativePin


dglaude:
Now contain a solution (with static if elif else) to solve issue #359:
default pin_factory based on environmental value

gpiozero/devices.py
```diff
@@ -34,20 +35,48 @@
...
 # supports platforms like PyPy
 from .pins import _pins_shutdown
+
+pin_factory = None
+env_default = os.getenv('GPIOZERO_PIN_FACTORY', None)
+
 try:
-    from .pins.rpigpio import RPiGPIOPin
-    pin_factory = RPiGPIOPin
+  if env_default is not None:
```
lurch: Can you please fix the indents to match the rest of the code? (i.e. 4 spaces per indent, rather than the two you've used here)

gpiozero/devices.py
```diff
@@ -34,20 +35,48 @@
...
+      from .pins.pigpiod import PiGPIOPin
+      pin_factory = PiGPIOPin
+    elif env_default == 'NativePin':
+      from .pins.native import NativePin
+      pin_factory = NativePin
+    elif env_default == 'MockPin':
+      from .pins.mock import MockPin
+      pin_factory = MockPin
+    else:
+      print ("Valid value for GPIOZERO_PIN_FACTORY are: 'RPiGPIOPin' 'RPIOPin' 'PiGPIOPin' 'NativePin' 'MockPin'")
```
lurch: IMHO "API code" like GpioZero shouldn't be calling `print` directly. Maybe it should be using https://docs.python.org/2/library/warnings.html instead?

bennuttall: Exception?

gpiozero/pins/pigpiod.py
```diff
@@ -98,7 +99,7 @@ class PiGPIOPin(Pin):
     GPIO_PULL_UP_NAMES = {v: k for (k, v) in GPIO_PULL_UPS.items()}
     GPIO_EDGES_NAMES = {v: k for (k, v) in GPIO_EDGES.items()}
 
-    def __new__(cls, number, host='localhost', port=8888):
+    def __new__(cls, number, host = os.getenv("PIGPIO_ADDR", 'localhost'), port = os.getenv("PIGPIO_PORT", 8888) ):
```
lurch: To match the style of the rest of the code, there shouldn't be spaces either side of the `=` characters, i.e.
```python
def __new__(cls, number, host=os.getenv("PIGPIO_ADDR", 'localhost'), port=os.getenv("PIGPIO_PORT", 8888)):
```

@waveform80 I guess theoretically these env-vars could change during the lifetime of a program, so should these default values be 'got' at module-import time, instead of during object-creation time? (I'm not sure what standard / pythonic behaviour is in circumstances like this.)

lurch:
I haven't actually tried testing your code yet, but just added some inline comments :)

gpiozero/devices.py
lurch: Nah, Exception implies fatal-error, whereas if the user has specified an invalid value for `GPIOZERO_PIN_FACTORY` IMHO it should just carry on anyway and use the 'old' fallback mechanism to find the 'best' GPIO implementation (which is what the code here does).

gpiozero/pins/pigpiod.py
waveform80: Actually those will be evaluated at module import time; function and method defaults are evaluated at declaration time (effectively module import time) rather than execution time. This is one of the reasons that things like this fail in interesting ways:

```python
def my_func(some_param=[]):
    some_param.append(1)
    return some_param

print(my_func())
print(my_func())
print(my_func())
```

lurch: Ahhhh. Kinda like backticks in Bash scripts then ;-)

waveform80:
Looks like this is superseded by #396? I'll close it for now, but let me know if I've missed something.

PR #367, opened 2016-06-08
Add optional enable init param to motor, close #366

### LINKED ISSUE #366: Motor with enable control
Motors on the new MotoZero board require 3 pins to control them. Forward pin, backward pin and an enable pin. They can be controlled with GPIO Zero by using `Motor` for the forward and backward pins, and `OutputDevice` for the enable pin like so:

``` python
>>> from gpiozero import Motor, OutputDevice
>>> motor = Motor(24, 27)
>>> motor_enable = OutputDevice(5)
>>> motor.forward()  # nothing happens
>>> motor_enable.on()  # the motor goes forward
>>> motor_enable.off()  # the motor stops
>>> motor.backward()  # nothing happens
>>> motor_enable.on()  # the motor goes backward
>>> motor.value = 0.5  # the motor goes forward at half speed
>>> motor.stop()  # the motor stops
```

My instinct is to want to remove the enabling step, and just let `Motor` turn on the enable pin on init, and never change its state, allowing `forward()`, `backward()`, `stop()` and `value` to be used to control the state of the motor.

However, Keith has [vocalised](https://twitter.com/PiTutorials/status/740489152253288448) his views that the enable pin is particularly useful:

> Nice benefit of using enable pin is that it can be PWMed to control the speed, independent of dir.

and

> Enable pin 0 = motor off, Enable pin 1 = Motor full speed. Enable pin PWM 0.5 = Motor 1/2 speed

But I don't think this makes it any easier than gpiozero already does:

``` python
>>> motor = Motor(17, 18)
>>> motor.forward()
>>> motor.value = 0.5
>>> motor.backward()
>>> motor.value = -0.5
>>> motor.reverse()
```

I propose we add an optional third pin to `Motor` and set it high on init. This will allow it to easily fit into `Motor` and `Robot`. Usage would look like:

``` python
>>> from gpiozero import Motor
>>> motor = Motor(24, 27, 5)
>>> motor.forward()  # motor goes forward
```

Any thoughts?

bennuttall:
(_Not complete yet, needs checking over and documenting_)

This PR adds a third (optional) pin parameter to `Motor`'s init. If provided, a `DigitalOutputDevice` is created (initially set high, and untouched elsewhere) and added to the list of devices.

As per #366.

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -797,6 +797,10 @@ class Motor(SourceMixin, CompositeDevice):
         The GPIO pin that the backward input of the motor driver chip is
         connected to.
 
+    :param int enable:
+        (Optional) The GPIO pin that enables the motor. Required for some motor
+        controller boards. Defaults to ``None``.
+
     :param bool pwm:
         If ``True`` (the default), construct :class:`PWMOutputDevice`
         instances for the motor controller pins, allowing both direction and
@@ -804,13 +808,20 @@ class Motor(SourceMixin, CompositeDevice):
         :class:`DigitalOutputDevice` instances, allowing only direction
         control.
     """
-    def __init__(self, forward=None, backward=None, pwm=True):
+    def __init__(self, forward=None, backward=None, enable=None, pwm=True):
         if not all([forward, backward]):
             raise GPIOPinMissing(
                 'forward and backward pins must be provided'
             )
         PinClass = PWMOutputDevice if pwm else DigitalOutputDevice
-        super(Motor, self).__init__(
+        if enable:
+            super(Motor, self).__init__(
+                forward_device=PinClass(forward),
+                backward_device=PinClass(backward),
+                enable_device=DigitalOutputDevice(enable, initial_value=True),
+                _order=('forward_device', 'backward_device', 'enable_device'))
+        else:
+            super(Motor, self).__init__(
                 forward_device=PinClass(forward),
                 backward_device=PinClass(backward),
                 _order=('forward_device', 'backward_device'))
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -814,7 +814,7 @@ class Motor(SourceMixin, CompositeDevice):
                 'forward and backward pins must be provided'
             )
         PinClass = PWMOutputDevice if pwm else DigitalOutputDevice
-        if enable:
+        if enable is not None:
             super(Motor, self).__init__(
                 forward_device=PinClass(forward),
                 backward_device=PinClass(backward),


gpiozero/output_devices.py
```diff
@@ -797,20 +797,31 @@ class Motor(SourceMixin, CompositeDevice):
...
+
     :param bool pwm:
         If ``True`` (the default), construct :class:`PWMOutputDevice`
         instances for the motor controller pins, allowing both direction and
         variable speed control. If ``False``, construct
         :class:`DigitalOutputDevice` instances, allowing only direction
         control.
     """
-    def __init__(self, forward=None, backward=None, pwm=True):
+    def __init__(self, forward=None, backward=None, enable=None, pwm=True):
```
lurch: To maintain backwards-compatibility, shouldn't the `enable` arg be added *after* the `pwm` arg?

gpiozero/output_devices.py
```diff
@@ -797,20 +797,31 @@ class Motor(SourceMixin, CompositeDevice):
...
         if not all([forward, backward]):
             raise GPIOPinMissing(
                 'forward and backward pins must be provided'
             )
         PinClass = PWMOutputDevice if pwm else DigitalOutputDevice
-        super(Motor, self).__init__(
+        if enable:
```
lurch: I think this would be better as `if enable is not None:` ?
(since in theory there's nothing stopping you using GPIO 0)

bennuttall:
We've broken BC in this way before! (e.g. when we added `fade_in_time` / `fade_out_time`)

`pwm` is usually passed as a kwarg.

gpiozero/output_devices.py
bennuttall: Ooh good point :P

gpiozero/output_devices.py
```diff
@@ -797,20 +797,31 @@ class Motor(SourceMixin, CompositeDevice):
...
+            super(Motor, self).__init__(
+                forward_device=PinClass(forward),
+                backward_device=PinClass(backward),
+                enable_device=DigitalOutputDevice(enable, initial_value=True),
+                _order=('forward_device', 'backward_device', 'enable_device'))
+        else:
+            super(Motor, self).__init__(
                 forward_device=PinClass(forward),
                 backward_device=PinClass(backward),
                 _order=('forward_device', 'backward_device'))
```
lurch: I guess an alternative (untested) way of doing this (with slightly less code-repetition) might be:
```python
devices = OrderedDict((
    ('forward_device', PinClass(forward)),
    ('backward_device', PinClass(backward)),
))
if enable is not None:
    devices['enable_device'] = DigitalOutputDevice(enable, initial_value=True)
super(Motor, self).__init__(_order=devices.keys(), **devices)
```

lurch:
> We've broken BC in this way before! (e.g. when we added fade_in_time / fade_out_time)

I can't remember if that was before or after the release of v1.0 ?

bennuttall:
It was after. We also broke it with `button.pin` which used to be an int, now is a `Pin` object.

bennuttall:
@waveform80 any thoughts on this PR?

lurch:
Hmmm, I wonder if people would also want `initial_enable` as a constructor kwarg?
And maybe there should be `enable` and `disable` methods (with `disable` being analogous to http://gpiozero.readthedocs.io/en/latest/api_output.html#gpiozero.Servo.detach ), which would be no-ops if the Motor doesn't have an `enable_device` ?
And then maybe there should also be an `enabled` property for completion, which would always return True for Motors without an `enable_device` ? Something like:
```python
@property
def enabled(self):
    return True if 'enable_device' not in self._named else self.enable_device.value
```

Hmmm part two: is it safe to assume that the enable_device will _always_ be `active_high=True` ? (perhaps we should just assume it is, until somebody provides a counter-example)

lurch:
OffTopic: I wonder if it'd be worth adding a `has_named_device` method (or something similar) to `CompositeDevice`, so that outside / child classes don't need to poke directly at `self._named` ? (and don't need to fuss with `getattr`)

bennuttall:
Tried a rebase but no dice...

bennuttall:
Now at #677

PR #371, opened 2016-06-10
Add lit_count property to LEDBarGraph

lurch:
Makes it easier to get / set the exact number of illuminated LEDs... kind of in the same spirit as `DistanceSensor.distance` and `ADC.voltage`.

(I originally called this `raw_value`, but decided it didn't "feel right", so changed it to `lit_count`)


lurch:
Just out of curiosity, is there anything that's preventing this PR from being merged?

bennuttall:
I don't see why not. @waveform80?

waveform80:
Nope, looks great to me! :)

PR #372, opened 2016-06-10
Add 'yellow' as an alias of 'amber' for TrafficLights

### LINKED ISSUE #345: Alias "amber" to "yellow" for American users
When teaching with gpiozero at Picademy USA, I discovered that they commonly refer to the middle light as "yellow", rather than "amber".

I suggest we add an alias for this.

@mrichardson23

lurch:
Fixes #345 

Not sure if this was the best way of doing this, but as the unit-test shows, it seems to work ;-)

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -536,11 +536,11 @@ class PiLiterBarGraph(LEDBarGraph):
 
 class TrafficLights(LEDBoard):
     """
-    Extends :class:`LEDBoard` for devices containing red, amber, and green
+    Extends :class:`LEDBoard` for devices containing red, yellow, and green
     LEDs.
 
     The following example initializes a device connected to GPIO pins 2, 3,
-    and 4, then lights the amber LED attached to GPIO 3::
+    and 4, then lights the amber (yellow) LED attached to GPIO 3::
 
         from gpiozero import TrafficLights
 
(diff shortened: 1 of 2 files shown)


lurch:
I noticed that `TrafficLights.value` (and `repr`) still said 'amber', even when constructed with a 'yellow' kwarg in `__init__`. The second commit fixes that too.

bennuttall:
I'd like to test this out

lurch:
Rebased...

waveform80:
Heh - I applaud the use of the OrderedDict. Beautiful way of building the device names for everything. 

Not so sure I like the `__getattr__` stuff - I tend to reserve that sort of thing for when I need to construct properties entirely on the fly (without knowing their names in advance). In this case, I'd prefer to construct the properties with fixed names. I'll see if I can grab your branch and tweak it a bit.

waveform80:
Hah! My own code which prevents CompositeDevice from stomping over pre-existing method / property names prevents manual construction of those properties (seeing as properties have to be class based, and can't just exist on an instance). All that means `__getattr__` is pretty much the only way of implementing this.

TL;DR: I'm convinced - this is about the cleanest way this can be implemented.

lurch:
> Heh - I applaud the use of the OrderedDict. Beautiful way of building the device names for everything.

Yeah, I thought it was quite elegant too. Kinda makes you wish that Python's `**kwargs` could be interpreted as an `OrderedDict` instead of a `dict` ;-)  (which would obviate the need for `_order`)

waveform80:
Yeah - ordering generally implies a loss of performance because it imposes an access order to certain operations (this is a major reason relational databases assume no order to tables unless one is imposed - well that, and it fits with the relational algebra's definition of a relation as a set of tuples).

Interestingly, pypy's later releases (I forget which off the top of my head) use a dict implementation which is ordered anyway (in other words, under pypy, OrderedDict = dict). They even claimed it was faster than their previous implementation. I'm sure they're right but I'm nonetheless intrigued as to how they managed it (haven't had the time to dig into the details). My gut instinct is there must be some trade off somewhere (more memory usage?)

PR #373, opened 2016-06-10
Allow TrafficLights, RGBLED and Motor to be constructed using GPIO 0

lurch:
Depending on which gets merged first (I don't mind), I suspect either this or #372 will need to be rebased...


bennuttall:
Yup. #372 needs a rebase

lurch:
Hah, I've just realised the irony of this patch adding support for GPIO 0 to a project named gpiozero...

bennuttall:
Yeah that made me chuckle :P

Bug: you can't use GPIO0 with GPIO Zero.

Issue #330, opened 2016-05-26
Add support for more motors


bennuttall:
As raised in #300, there's a suggestion to add more varieties of motor, such as the L298N.

> I believe to 'properly' support the L298N we'd need a Motor class with (Digital_forwards, Digital_backwards, PWM_speed) style controls.
> ~ @lurch 

Would this be another generic motor class? What would it be called?

Are there other types of motors not covered by this?

lurch:
Yup, there's also #276 and #314 and #315 
And I made some "comments about Motor classes" here: https://github.com/RPi-Distro/python-gpiozero/pull/277#issuecomment-211608633

goloplo:
Yes, there is L293D DC motor driver.
Keep up the good work.

lurch:
I've just had a look through the L293D datasheet, and AFAICT it'll be handled just fine by the current `Motor` class (assuming you hard-wire the EN pin high).

lurch:
And #366

PR #386, opened 2016-06-27
Add support for Phase/Enable motor board

MrHarcombe:
This is a re-request of my previous PR #276, now based on a branch of my fork to hopefully make life a little easier (for me, as much as anyone else).

It adds support for motors controlled by a motor board using phase/enable (ie one PWM pin for motor power, and another pin for direction) such as the Pololu DRV8835.


gpiozero/output_devices.py
```diff
@@ -897,3 +897,92 @@ def stop(self):
...
+            direction_device = OutputDevice(direction),
+            _order = ('power_device', 'direction_device'))
+
+    @property
+    def value(self):
+        """
+        Represents the speed of the motor as a floating point value between -1
+        (full speed backward) and 1 (full speed forward).
+        """
+        return self.power_device.value if self.direction_device.is_active() else -self.power_device.value
```
bennuttall: Should this be `self.direction_device.is_active` not `self.direction_device.is_active()`?

lurch:
Yup, PR is looking in a much healthier state now ;-)
(it takes a while to get the hang of git and branches when you first start using it, but after a while it 'clicks' - basically never commit to (your) master, always do everything in a branch).

I believe your PhaseEnableRobot ought to be updated to better-match the existing Robot implementation though...
i.e. replace
```python
    def __init__(self):
        super(PhaseEnableRobot, self).__init__()
        self._left = PhaseEnableMotor(12, 5)
        self._right = PhaseEnableMotor(13, 6)
```
with
```python
    def __init__(self):
        super(PhaseEnableRobot, self).__init__(
            left_motor=PhaseEnableMotor(12, 5),
            right_motor=PhaseEnableMotor(13, 6),
            _order=('left_motor', 'right_motor'))
```
and then change all references to `self._left` with `self.left_motor` etc. etc.

gpiozero/output_devices.py
lurch: Yeah, `is_active` is a property rather than a method, so you don't want parentheses.
(decorators are another thing it takes a while to get your head around at first)

bennuttall:
Also add the new classes to `__init__.py`

MrHarcombe:
Shall do - thanks, guys!

MrHarcombe:
Think I'm there with the motor/robot code... but I'm having problems writing test cases.

When I run `make test` the first time my PhaseEnableRobot fails, I get...
`cls = <class 'gpiozero.pins.mock.MockPWMPin'>, number = 5

    def __new__(cls, number):
        if not (0 <= number < 54):
            raise ValueError('invalid pin %d specified (must be 0..53)' % number)
        try:
            old_pin = cls._PINS[number]
        except KeyError:
            self = super(MockPin, cls).__new__(cls)
            cls._PINS[number] = self
            self._number = number
            self._function = 'input'
            self._state = False
            self._pull = 'floating'
            self._bounce = None
            self._edges = 'both'
            self._when_changed = None
            self.clear_states()
            return self
        if old_pin.__class__ != cls:
>           raise ValueError('pin %d is already in use as a %s' % (number, old_pin.__class__.__name__))
E           ValueError: pin 5 is already in use as a MockPin

gpiozero/pins/mock.py:53: ValueError`

Plainly I'm doing something wrong... any thoughts?

lurch:
Looks like you've set up Pin5 as a MockPin, but when you run the test-code it's trying to create Pin5 as a MockPWMPin.
Try changing your testcode to instead create a MockPWMPin on Pin5?

...although looking at the code in your PR, I'd *expect* Pin5 to be a MockPin, as AFAICT it's a `direction_device` rather than a `power_device`. If you want to paste your failing testcode, I'll have a look at it on Sunday (I'm jam-packed busy between now and then).

MrHarcombe:
That's a useful steer, thanks, @lurch - I'll chase that through and see where it leads me.

MrHarcombe:
Well, I got something going... but only by accepting defeat. I *cannot* understand why the code is apparently creating a MockPWMPin for 5 and 6 when it creates the PhaseEnableRobot.

Any thoughts?

```
def test_phaseenable_robot():
    pins = [MockPWMPin(12), MockPWMPin(5), MockPWMPin(13), MockPWMPin(6)]
    with PhaseEnableRobot() as robot:
        assert (
            [device.pin for device in robot.left_motor] +
            [device.pin for device in robot.right_motor]) == pins
        assert robot.value == (0, 0)
        robot.forward()
        assert [pin.state for pin in pins] == [1, 1, 1, 1]
        assert robot.value == (1, 1)
        robot.backward()
        assert [pin.state for pin in pins] == [1, 0, 1, 0]
        assert robot.value == (-1, -1)
        robot.forward(0.5)
        assert [pin.state for pin in pins] == [0.5, 1, 0.5, 1]
        assert robot.value == (0.5, 0.5)
        robot.left()
        assert [pin.state for pin in pins] == [1, 0, 1, 1]
        assert robot.value == (-1, 1)
        robot.right()
        assert [pin.state for pin in pins] == [1, 1, 1, 0]
        assert robot.value == (1, -1)
        robot.reverse()
        assert [pin.state for pin in pins] == [1, 0, 1, 1]
        assert robot.value == (-1, 1)
        robot.stop()
        assert [pin.state for pin in pins] == [0, 0, 0, 1]
        assert robot.value == (0, 0)
        robot.value = (-1, -1)
        assert robot.value == (-1, -1)
        robot.value = (0.5, 1)
        assert robot.value == (0.5, 1)
        robot.value = (0, -0.5)
        assert robot.value == (0, -0.5)
```

lurch:
> I _cannot_ understand why the code is apparently creating a MockPWMPin for 5 and 6 when it creates the PhaseEnableRobot. Any thoughts?

Apologies for sending you on a wild goose chase... :-S 
https://github.com/RPi-Distro/python-gpiozero/blob/master/tests/test_boards.py#L18 means that each test either uses `MockPin` for all pins **or** `MockPWMPin` for all pins; there's no support for a single test using a mixture of `MockPin` and `MockPWMPin`. Of course "in real life" all GPIO pins are analogous to `MockPWMPin` (unless you're using the `NativePin` implementation, which doesn't support PWM).

I guess *in theory* the `gpiozero.devices.pin_factory` could be split into separate `gpiozero.devices.digital_pin_factory` and `gpiozero.devices.pwm_pin_factory` (with the low-level classes being updated as appropriate), but IMHO that'd be a lot of work for no noticeable benefit.

MrHarcombe:
Hah - no worries, I hadn't thought that through when I was digging into the code. *slaps forehead*
Glad I just got things working, then - I'll just leave this here, then... :)

lurch:
> I'll just leave this here, then... :)

I'm hoping to refactor the way that `Motor`s and `Robot`s are implemented, to make it easier for alternative-version classes to implement the same interfaces. So if you don't mind, I'll leave this PR unmerged until after that :-)  (and then provide guidance on how this PR should be updated)

MrHarcombe:
Not a problem - happy to be contributing in any shape/form :)

(going to be looking at a Pololu line sensor module over the summer;
there's only an Arduino library at the moment...)

waveform80:
@bennuttall looks good to me - my one possible objection (which I'm happy to quickly subst) is that I'd use "speed" instead of "power" in the constructor ("power" *could* be mis-construed as something strictly on/off, while "speed" implies a variable setting and is consistent with similar usage in the `Motor` class)?

bennuttall:
Agreed

bennuttall:
Actually, just reviewing this I don't like `speed` either because I think `speed=5` would indicate that the speed is set to level 5 (if I didn't know better I'd probably assume 50% speed).

I'd suggest `speed_pin` and `direction_pin` as we use `pin` for single-pin devices, but for multi-pin devices we tend to name those arguments without `pin`, e.g. `left`, `right`, `echo`, `trigger`.

bennuttall:
Also this seems related to #330 and #366.

waveform80:
It's sort of related - my feeling on enable pins for H-bridges is that it could be added as a either a simple sub-class of Motor or even an optional parameter. But I haven't got time to include that in 1.4.1 - maybe something to think about for 1.5.

Back on topic for this one: how about we just use "phase" and "enable" instead of "speed" and "direction". If this seems a little obscure in wording, I'd argue firstly it's reminiscent of the class name, and secondly that in several cases this is the wording silk-screened on the board: https://a.pololu-files.com/picture/0J4054.1200.jpg?77f6351fc4ce555a39069a8af67b52f9 (so from the user's perspective, they set "phase" to whatever they've connected the phase pin to and so on).

waveform80:
Just chatted to Ben and he's fine with phase/enable for pin naming within the motor class. One other change I'm going to make is that there seem to be a few phase/enable motor drivers knocking around so PhaseEnableRobot should probably be a generic class with configurable pins, and we should have something like PololuPhaseEnableRobot (maybe just PololuRobot? I'll have a look what else they ship) descend from that just fixing pins in its constructor (like RyanteckRobot does with Robot).

Nearly done; this'll be the last thing to go in then I'll start on 1.4.1 packaging.

waveform80:
Well, this wound up a bit more of a re-write than I thought but I think the result looks alright. Turns out Pololu do a whole range of Phase/Enable driver boards so we've now got PhaseEnableRobot that takes left/right pins (like Robot - I still like the idea of a generic Robot class that takes two motor entries but that'll have to wait for now and I think we can migrate to that without much pain in future). From that descends PololuDRV8835Robot which uses the specific pins for @MrHarcombe's board.

The one other change I made is that the forward/backward values of the phase pin didn't match the specification (see table "Drive/brake operation in default PHASE/ENABLE mode" under https://www.pololu.com/product/2753) so I've flipped them and adjusted the tests accordingly. Finally, doc entries are added for all new classes.

MrHarcombe:
Brilliant - thanks guys, my first ever contribution to an open source project! 👍

Issue #359, opened 2016-06-06
default pin_factory based on environmental value


dglaude:
Based on #353 discussion, here is the idea.

In pigpio you define in environment variables how to connect to the pigpiod of your Pi (if nothing is explicitly said in the code).

That same trick could/should be used into gpiozero and provide a way to externally choose the pin implementation. Currently it is hard wired starting from line 32 of devices.py. It could be softwired with an single environmental variable such as GPIOZERO_PIN_FACTORY that could be not set for current behaviour or set to RPiGPIOPin or RPIOPin or PiGPIOin or NativePin.

lurch:
If this gets implemented, it would also make @ryanteck 's RTk.GPIO even *more* plug'n'play (assuming he adds the promised GpioZero integration) with exisiting GpioZero scripts than he already thought ;-)

dglaude:
Here is the proposed "test" code to verify the behaviour based on environmental value.
Latter try to write some code that bypass the fall-back sequence if GPIOZERO_PIN_FACTORY is set.

```
import gpiozero.devices
print ("Selected pin factory is " , gpiozero.devices.pin_factory)

import os
print ("GPIOZERO_PIN_FACTORY is ", os.getenv("GPIOZERO_PIN_FACTORY", 'not set'))
print ("PIGPIO_ADDR is ", os.getenv("PIGPIO_ADDR", 'not set'))
print ("PIGPIO_PORT is ", os.getenv("PIGPIO_PORT", 'not set'))

from gpiozero import LED
red = LED(9)
red.blink()

from signal import pause
pause()
```

This it the output on my Mac:

> $ python factory.py 
> ('Selected pin factory is ', <class 'gpiozero.pins.pigpiod.PiGPIOPin'>)
> ('GPIOZERO_PIN_FACTORY is ', 'not set')
> ('PIGPIO_ADDR is ', '192.168.192.113')
> ('PIGPIO_PORT is ', 'not set')

This is the output on my Pi:

> $ python factory.py 
> ('Selected pin factory is ', <class 'gpiozero.pins.rpigpio.RPiGPIOPin'>)
> ('GPIOZERO_PIN_FACTORY is ', 'not set')
> ('PIGPIO_ADDR is ', 'not set')
> ('PIGPIO_PORT is ', 'not set')

I remove* RPiGPIOin with  `sudo apt-get remove python-rpi.gpio` this is the output on my Pi:

> $ python factory.py 
[...]

PR #388, opened 2016-06-30
add RPi.GPIO to dependencies

### LINKED ISSUE #387: [performance] Button - huge CPU hog
Whenever I instantiate the Button class, I get ca. 35% CPU load per instance:

1. run "sudo python3" in one console and open "top" or "htop" in another one for monitoring
2. find the python3 process in the top list
3. enter in python interactive mode:

`import gpiozero`
`button1 = gpiozero.Button(4)`

EFFECT: +30...35% CPU load by python3 after instantiation of Button
SHOULD BE: CPU load doesn't rise drastically

Now try creating some more Button instances for different GPIO pins:

`button2 = gpiozero.Button(24)`

EFFECT: the CPU load rises proportionally, causing the Raspberry to run out of resources if too many instances are created

Try to get rid of the instances:

`button1 = None`
`button2 = None`

EFFECT: CPU load doesn't drop
SHOULD BE: instances are deleted and the load drops

Exit the interpreter:

`exit()`

EFFECT: the CPU load finally comes back to normal.

Tested on the old RPi B rev 2 and B+, Raspbian stretch, Python 3.5.2

elegantandrogyne:
fix #387

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ __keywords__ = [
     'gpio',
 ]
 
-__requires__ = [
+__requires__ = ['RPi.GPIO>0.6.0',
 ]
 
 __extra_requires__ = {


bennuttall:
gpiozero doesn't depend on RPi.GPIO. If it's present, we use it. If it's not, there are fallback alternatives. For example, some people will want to use RPIO for its better PWM support.

And you can even install gpiozero on a PC, and use pigpio's remote GPIO functionality to create gpiozero objects referring to pins on a Pi over a network - so the GPIO library itself is not required.

lurch:
@bennuttall So, RPi.GPIO isn't a dependency in `setup.py`, but it *is* a dependency in `debian/control`...?
See also my comments in https://github.com/RPi-Distro/python-gpiozero/issues/359#issuecomment-225455765

bennuttall:
It should be removed from debian/control.

bennuttall:
(and added to recommends)

waveform80:
That's now done (#399)

Issue #389, opened 2016-07-01
gpiozero fails on a machine with the warranty bit set.


DougieLawson:
See forum thread https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=153124. I wrote this patch and fixed the problem for the OP on that thread.

```
--- data.py.orig        2016-07-01 17:02:31.434937243 +0100
+++ data.py     2016-07-01 17:06:03.166932201 +0100
@@ -493,9 +493,9 @@
         for line in f:
             if line.startswith('Revision'):
                 revision = line.split(':')[1].strip().lower()
-                overvolted = revision.startswith('1000')
+                overvolted = revision.startswith('100')
                 if overvolted:
-                    revision = revision[4:]
+                    revision = revision[3:]
                 return revision
         raise IOError('unable to locate Pi revision in /proc/cpuinfo')
```

DougieLawson:
Close. Temporary patch works until V1.3.

PR #390, opened 2016-07-01
gpiozero fails on a Revision: 000e system with the warranty bit set.

DougieLawson:
https://github.com/RPi-Distro/python-gpiozero/issues/389

Pull request to resolve issue #389

### DIFF
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -493,9 +493,9 @@ def _get_pi_revision():
         for line in f:
             if line.startswith('Revision'):
                 revision = line.split(':')[1].strip().lower()
-                overvolted = revision.startswith('1000')
+                overvolted = revision.startswith('100')
                 if overvolted:
-                    revision = revision[4:]
+                    revision = revision[3:]
                 return revision
         raise IOError('unable to locate Pi revision in /proc/cpuinfo')
 


lurch:
That's interesting... http://elinux.org/RPi_HardwareHistory#Board_Revision_History implies that a blown warranty increases the Revision to 8 (hex) characters; whereas the forum thread you linked to implies that blowing the warranty bit increases the Revision to 7 characters. (I wonder if it's dependent on the firmware version?)

However in either case, this patch isn't actually needed, as commit 79a0fd2 (added since the release of gpiozero 1.2) changes it so that the revision is parsed and compared as an integer (i.e. `0xe`) instead of as a string (`'000e'`).

DougieLawson:
Read the forum thread. The guy with his warranty bit set is running V1.2 and gets a failure. So any changes in V1.2 simply don't work.

lurch:
> Read the forum thread.

I did ;-)

> So any changes in V1.2 simply don't work.

Sorry if my previous reply wasn't very clear.
Yes, this problem still exists in the released V1.2 of gpiozero, but it's already been (inadvertently) fixed in the github version of gpiozero; which means the next 'official release' of gpiozero (whenever that happens) will contain the fix for this problem.

DougieLawson:
OK. Close this. The guy with the problem has a working patch for V1.2 and some time in some future galaxy the newer version will become the mainstream version.

waveform80:
> but it's already been (inadvertently) fixed in the github version of gpiozero

It has? Sorry - was just reviewing this (ploughing through my enormous pile of unread notifications!) and I'm not sure I see where it's got inadvertently fixed. Any chance you can point me at the relevant bit @lurch?

waveform80:
(and incidentally apologies to anyone hit by this - it's my code that broke this originally - as noted above I'd based it entirely on the written spec without actually testing it on the hardware)

lurch:
"However in either case, this patch isn't actually needed, as commit 79a0fd2 (added since the release of gpiozero 1.2) changes it so that the revision is parsed and compared as an integer (i.e. 0xe) instead of as a string ('000e')."

waveform80:
Ahh, sorry - missed that completely. You're absolutely right. I must admit I'm still tempted to merge this because my original code is still "wrong", i.e. it's slicing too much off the front of the string. On off chance we ever need to use the output of `_get_pi_revision` somewhere else and it needs the leading zeros (yeah, I can't think of anything ... but what the hell :) it's worth including the fix anyway.

lurch:
I dunno... as I mentioned in another earlier comment:
"That's interesting... http://elinux.org/RPi_HardwareHistory#Board_Revision_History implies that a blown warranty increases the Revision to 8 (hex) characters; whereas the forum thread you linked to implies that blowing the warranty bit increases the Revision to 7 characters. (I wonder if it's dependent on the firmware version?)"

I've only got one Pi with a blown warranty-fuse myself, and that matches up with what the eLinux wiki says should happen, so I've still no idea how or why **nattybumppo829** in @DougieLawson 's [forum thread](https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=153124) has his warranty-bit in a different place :-S
Any ideas @ghollingworth ?

waveform80:
Ah, interesting - I'm not sure if any of my old Pi's have blown warranty bits. I'm sure one must because I've definitely turbo'd one of them before - unfortunately they're all tied up as media clients on various TVs at the moment. Might see if I can investigate further later.

lurch:
@waveform80 any particular reason why `_get_pi_revision` returns a `str` and not an `int` anyway?

waveform80:
Historical reasons basically - when I first wrote that I'd assumed "Beta" was a valid return value (based on the eLinux wiki) but since the chances of ever seeing that in the wild seem almost nil (and we don't really care about supporting the minute number of beta boards out there) it might as well be int these days.

Given it seems there's some discrepancy over whether it's 7 or 8 chars I'm tempted to modify the code to just return `revision[-4:]`, i.e. return the last 4 characters regardless of however many there are in the string (and perhaps move the hex-string conversion into `_get_pi_revision` too, so it does indeed return an int).

PR #395, opened 2016-07-26
Use PIGPIO_ADDR and PIGPIO_PORT environment variable if they exist to…

dglaude:
… configure PiGPIO, otherwise it use the default value 'localhost' and 8888. This is the same behaviour as the pigpio library for initialisation.

See #353 for more explanation.

This is the PiGPIO part of #356 and it follow the remark from @lurch about space around "=".
The environmental variable for selecting the pin factory will be in a separate pull request.
The typo has already been fixed by @bennuttall

### DIFF
--- a/gpiozero/pins/pigpiod.py
+++ b/gpiozero/pins/pigpiod.py
@@ -99,7 +99,7 @@ class PiGPIOPin(Pin):
     GPIO_PULL_UP_NAMES = {v: k for (k, v) in GPIO_PULL_UPS.items()}
     GPIO_EDGES_NAMES = {v: k for (k, v) in GPIO_EDGES.items()}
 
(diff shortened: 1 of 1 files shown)


lurch:
Sorry to be super-picky yet again, but for consistency with the rest of the gpiozero code, the strings should be enclosed in single-quotes `'` rather than double-quotes `"`.
And perhaps the port should be coerced to an integer? i.e. `port=int(os.getenv('PIGPIO_PORT', 8888))`

lurch:
...but thankyou for splitting it up into smaller parts, makes each part much easier to review in isolation :-)

dglaude:
The reason why all my changes started to merge into a single PR is that I do not master git. Now I will make one branch per PR and try to keep my master in sync. I hope it will give better result.

waveform80:
Yup, this looks good.

PR #401, opened 2016-08-14
Annotate recipe with LED sys/class names

waveform80:


lurch:
IMHO an explicit mention in the text of something like "/sys/class/leds/led0 is used to control the activity LED, and /sys/class/leds/led1 is used to control the power LED" would be better than an easily-ignored inline comment in the python code.

waveform80:
Yeah, good point - any chance you feel like bunging together a PR for this? Got about 6 branches on my clone at the moment (servos, distance sensor fixes, remote SPI work) that I'm trying to juggle :)

lurch:
Sure, I'll put something together later in the week :)

PR #402, opened 2016-08-14
Raise warning when spidev missing for hardware SPI

waveform80:


gpiozero/spi.py
```diff
@@ -396,24 +396,28 @@ def SPI(**spi_args):
...
-                    'software (error was: %s)' % str(e)))
+                    'failed to import spidev, falling back to software SPI'))
+        else:
+            try:
+                if shared:
+                    return SharedSPIHardwareInterface(
+                            port=0, device={8: 0, 7: 1}[spi_args['select_pin']])
+                else:
+                    return SPIHardwareInterface(
+                            port=0, device={8: 0, 7: 1}[spi_args['select_pin']])
```
lurch: Might it make sense to refactor this to:
```python
    hardware_spi_args = {port: 0, device: {8: 0, 7: 1}[spi_args['select_pin']]}
    if shared:
        return SharedSPIHardwareInterface(**hardware_spi_args)
    else:
        return SPIHardwareInterface(**hardware_spi_args)
```
or is that taking DRY too far? ;)

waveform80: No, that's nicer - just didn't occur to me!

Issue #295, opened 2016-04-21
Privacy breach: local documentation shouldn't include Travis build status


EdwardBetts:
As discussed in #284.

The documentation available at http://gpiozero.readthedocs.org/en/latest/ includes the Travis build status. This is useful.

The build status shouldn't be included in the documentation intended to be read offline, specifically in the Debian package of documentation.

waveform80:
Thanks for the reminder ticket - I'll see what I can do about this for the next release

waveform80:
ARGH! Bloody PyPI doesn't work with the damned thing either. Right, sorry about this but I'm going to have to revert the change - it just breaks too many things and I've got no time to fix it now. If someone wants to do a proper fix later (whatever that may entail), be my guest.

lurch:
Was this closed again accidentally? Should it be re-opened?

waveform80:
Bah, yeah that's GitHub reacting to the "Fix" in the title without seeing the "Revert" which git bunged before it.

waveform80:
Fixed in a2b2b816510cebdbd2e4cb56906b2fe27e8bbf7e

PR #403, opened 2016-08-14
Fix #320

### LINKED ISSUE #320: Broken link in documentation
I've just noticed that the "Contribute" link on http://gpiozero.readthedocs.io/en/latest/index.html#development gives a 404 error.

waveform80:
Yup, dirty horrid hack but the result is okay.


lurch:
Hmm, I've just noticed that the `recipes` link on http://gpiozero.readthedocs.io/en/v1.2.0/#development links to `http://gpiozero.readthedocs.org/en/latest/recipes.html`, and I guess this new `contribute` link will behave the same (once it makes it into 'stable'). Note the mismatch between `v1.2.0` and `latest`.

Is there any way the absolute URL could be converted back into a relative URL during the sphinx process, i.e. so that the `recipes` link mentioned above would then be just `recipes.html` ?  (which I guess would also make it work better for the offline docs)

waveform80:
I wish there were, but no - GitHub's reST renderer doesn't understand substitutions so there's no way I can make the link any more nuanced and have it work under both GitHub and RTD. On RTD alone I could do it (simple `:ref:` link).

lurch:
Would having separate index / README pages for RTD and Github be too much of a maintenance burden then? (or is the index of RTD automatically the Github README / vice-versa?)

waveform80:
The RTD index is only the same as README because I `.. include` it - in other words, it's my DRY sense in overdrive (I always try and remove redundancy so there's as few places to update as possible - otherwise sod's law I always forget one). Still, you're right - it's not such a big deal to have a separate index and README.

bennuttall:
Unfortunately GitHub doesn't pick up `CONTRIBUTING.RST`.

See the difference between [RPi-Distro/python-gpiozero](https://github.com/RPi-Distro/python-gpiozero/issues/new) and [RPi-Distro/python-sense-hat](https://github.com/RPi-Distro/python-sense-hat/issues/new).

From https://github.com/blog/1184-contributing-guidelines:

> As a maintainer, all you have to do is add a CONTRIBUTING file (or CONTRIBUTING.md if you're using Markdown) to the root of your repository.

waveform80:
Eurgh ... sometimes GitHub really annoys me. Okay, I'll revert it and figure something out (even if it is duplicating something)

lurch:
@waveform80 Did this actually get reverted? Was it supposed to have been? The root of this repo still has no CONTRIBUTING or CONTRIBUTING.md file.

waveform80:
Erm, good question - there's e9ff6fce3cb33fff4c3792a1ffc5eab6bbc210f0 which reverts the Fix #295 patch but I never got around to introducing a CONTRIBUTING file.

PR #405, opened 2016-08-17
Correct number of ComputeModule USB ports (part 2)

lurch:
Correct associated documentation


waveform80:
Well, I'll merge this when I get back home anyway - apparently I can't merge from mobile (add all ironic hashtags you can imagine ... 21st century problems etc :-)

PR #411, opened 2016-08-29
Convert recipe examples to includes

waveform80:
Makes it much easier to test things - no copying'n'pasting just run the
examples straight from the dir (after wiring stuff up)

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -23,34 +23,19 @@ components:
 
 .. image:: images/pin_layout.*
 
+
 LED
 ===
 
 .. image:: images/led.*
 
-Turn an :class:`LED` on and off repeatedly::
+Turn an :class:`LED` on and off repeatedly:
 
-    from gpiozero import LED
-    from time import sleep
+.. literalinclude:: examples/led_1.py
 
-    red = LED(17)
+Alternatively:
 
-    while True:
-        red.on()
-        sleep(1)
-        red.off()
-        sleep(1)
-
-Alternatively::
-
-    from gpiozero import LED
-    from signal import pause
-
-    red = LED(17)
-
-    red.blink()
-
-    pause()
+.. literalinclude:: examples/led_2.py
 
 .. note::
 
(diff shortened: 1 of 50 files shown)


docs/recipes.rst
```diff
@@ -930,54 +385,24 @@ Measure temperature with an ADC
...
-    led = RGBLED(2, 3, 4)
-    red_pot = MCP3008(0)
-    green_pot = MCP3008(1)
-    blue_pot = MCP3008(2)
-
-    led.source = zip(red_pot.values, green_pot.values, blue_pot.values)
-
-    pause()
+.. literalinclude:: examples/rgbled_pot_2.py
+    :emphasize-lines: 8
```
lurch: I think you emphasized the wrong line? http://gpiozero.readthedocs.io/en/latest/recipes.html#full-color-led-controlled-by-3-potentiometers

lurch:
Nice! But perhaps filenames like `all_on_traffichat.py` would have been more meaningful than `all_on_2.py` ? ;-)

docs/recipes.rst
waveform80: Doh - I did - unfortunately the offline build (which I normally give a quick eye-ball before pushing a PR) doesn't include the emphasis for some reason. Must figure out a way I can test this better.

waveform80:
Yeah, they would be - but my creativity only stretches so far at 11pm :) I'll have to leave it for now - but it ought to be fixed up at some point

docs/recipes.rst
lurch: Could it be as simple as the offline build using the wrong theme or something?

waveform80: Offline build's using the default theme - unfortunately I think it's unaware of the emphasis styles. But then the offline build has to remain compatible with whatever downlevel version of sphinx is supported in Raspbian at the moment (debuild ensures that venv's aren't used during package build so you can try and use a later sphinx, but it won't work - you'll wind up on the raspbian installed version)

lurch: Ahhh, so maybe there should be a `doc_with_rtd_styling:` target in the Makefile? ;-)  (for testing purposes)

waveform80: Heh - maybe - don't think I'll have time for that any-time soon though!

PR #412, opened 2016-08-29
Fix #289

### LINKED ISSUE #289: Add note about remote access
Note

The second example of the Keyboard Controlled Robot will only work if the keyboard/screen are attached directly to the RPi, so will not work with remote access mechanisms (SSH and VNC).

waveform80:
Tweak the second keyboard robot recipe so it detects keyboards and
doesn't throw exceptions on unknown keys. Also add a note to the recipe
that it won't work over remote connections like SSH.

### DIFF
--- a/docs/examples/robot_keyboard_2.py
+++ b/docs/examples/robot_keyboard_2.py
@@ -3,8 +3,23 @@ from evdev import InputDevice, list_devices, ecodes
 
 robot = Robot(left=(4, 14), right=(17, 18))
 
+# Get the list of available input devices
 devices = [InputDevice(device) for device in list_devices()]
-keyboard = devices[0]  # this may vary
+# Filter out everything that's not a keyboard. Keyboards are defined as any
+# device which has keys, and which specifically has keys 1..31 (roughly Esc,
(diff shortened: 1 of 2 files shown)


docs/examples/robot_keyboard_2.py
```diff
@@ -3,8 +3,23 @@
...
+# *not* have key 0 (reserved)
+must_have = {i for i in range(1, 32)}
+must_not_have = {0}
+devices = [
+    device
+    for device in devices
+    for keys in (set(device.capabilities().get(ecodes.EV_KEY, [])),)
+    if must_have.issubset(keys)
+    and must_not_have.isdisjoint(keys)
+]
```
lurch: Is it just me, or is `devices = [device for device in devices for ...` a bit unreadable? ;-)

waveform80: Hmm, is it the similarity of the words that's the issue? Would "dev" be better to full distinguish the result variable from the source variable "devices"?

lurch: Already replied to this elsewhere, sorry. For some reason it's taking ages for the github notification emails to come through :-(

PR #413, opened 2016-08-29
Fix #295

### LINKED ISSUE #295: Privacy breach: local documentation shouldn't include Travis build status

waveform80:
At least, I *think* this should fix this. It does locally, but we'll
have to see what happens when I push it to RTD!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,17 +2,19 @@
 gpiozero
 ========
 
-.. image:: https://badge.fury.io/py/gpiozero.svg
-    :target: https://badge.fury.io/py/gpiozero
-    :alt: Latest Version
+.. ifconfig:: html_theme == 'sphinx_rtd_theme'
 
-.. image:: https://travis-ci.org/RPi-Distro/python-gpiozero.svg?branch=master
-    :target: https://travis-ci.org/RPi-Distro/python-gpiozero
-    :alt: Build Tests
+    .. image:: https://badge.fury.io/py/gpiozero.svg
+        :target: https://badge.fury.io/py/gpiozero
+        :alt: Latest Version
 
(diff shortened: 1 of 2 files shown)


lurch:
Could something similar be used to re-do #403 in a better way?

lurch:
P.S. When you use "Fix #nnn" as your PR title, it's a shame that github doesn't auto-linkify it to the corresponding issue :-/

lurch:
OTOH this fix means that the build-status buttons have now disappeared from https://github.com/RPi-Distro/python-gpiozero :-(  So maybe we really do need separate readme / index files for github and the documentation? (or maybe you could have a git-commit-hook so that updating one automatically updates the other)

waveform80:
> Could something similar be used to re-do #403 in a better way?

Possibly - will have to check on that.

> P.S. When you use "Fix #nnn" as your PR title, it's a shame that github doesn't auto-linkify it to the corresponding issue :-/

No kidding - really annoying when Trac used to do exactly that (I still miss some things about Trac...)

> OTOH this fix means that the build-status buttons have now disappeared from https://github.com/RPi-Distro/python-gpiozero :-( So maybe we really do need separate readme / index files for github and the documentation? (or maybe you could have a git-commit-hook so that editing one automatically generates the other)

Yeah - saw that last night and said to myself "it's midnight ... declare victory for now". I'll try and get around to fixing it at some point, but it's kinda low down the list. Unfortunately GitHub's reST implementation doesn't know about the ifconfig extension (for that matter it doesn't properly degrade with unknown tags which is kinda crap). Anyway - any suggestions welcome, but for now I'm saying "good enough".

PR #415, opened 2016-08-30
Fix all the stuff you broke last night...

waveform80:
In particular the `pi_revision` thing in PiGPIOPin, all the stuff @lurch
picked up in `pins/data.py` (thank goodness *someone's* watching!), and
make all those links pointing to "Notes" point somewhere useful like
"Pin Numbering"...

### DIFF
--- a/docs/examples/robot_keyboard_2.py
+++ b/docs/examples/robot_keyboard_2.py
@@ -12,9 +12,9 @@ devices = [InputDevice(device) for device in list_devices()]
 must_have = {i for i in range(1, 32)}
 must_not_have = {0}
 devices = [
-    device
-    for device in devices
-    for keys in (set(device.capabilities().get(ecodes.EV_KEY, [])),)
+    dev
+    for dev in devices
+    for keys in (set(dev.capabilities().get(ecodes.EV_KEY, [])),)
     if must_have.issubset(keys)
     and must_not_have.isdisjoint(keys)
 ]
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -573,10 +573,12 @@ def _parse_pi_revision(revision):
             '3B': True,
             }.get(model, False)
         csi = {
-            'Zero': 0 if pcb_revision == '0.0' else 1,
+            'Zero': 0 if pcb_revision == '1.0' else 1,
             'CM':   2,
             }.get(model, 1)
-        dsi = csi
+        dsi = {
+            'Zero': 0,
+            }.get(model, csi)
         headers = {
             'A':  {'P1': REV2_P1, 'P5': REV2_P5},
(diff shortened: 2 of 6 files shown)


gpiozero/pins/data.py
```diff
@@ -573,10 +573,12 @@ def _parse_pi_revision(revision):
             '3B': True,
             }.get(model, False)
         csi = {
-            'Zero': 0 if pcb_revision == '0.0' else 1,
+            'Zero': 0 if pcb_revision == '1.0' else 1,
             'CM':   2,
             }.get(model, 1)
-        dsi = csi
+        dsi = {
+            'Zero': csi if pcb_revision == '1.0' else 0,
```
lurch: Huh? *Both* revisions of the Pi Zero have no DSI connectors...

waveform80: Urgh ... not thinking straight - bouncing between two branches now (a cherry picked one and the servo one)! Thanks for the catch :)

lurch:
Should the PWM stuff have been in a separate PR?

docs/examples/robot_keyboard_2.py
```diff
@@ -12,9 +12,9 @@
 must_have = {i for i in range(1, 32)}
 must_not_have = {0}
 devices = [
-    device
-    for device in devices
-    for keys in (set(device.capabilities().get(ecodes.EV_KEY, [])),)
+    dev
+    for dev in devices
+    for keys in (set(dev.capabilities().get(ecodes.EV_KEY, [])),)
```
lurch: Heh, I was more referring to the multiple usages of `devices` and nested list comprehensions - if I was writing this myself I might have done something like:
```keyboard_devices = filter(is_keyboard, evdev_devices)```
IYSWIM  \*shrug\*

waveform80:
Ahh, I was considering it, but I'm too squeezed for time right now so this is basically "fix everything so you've got a clean PR which only includes Servo et al" - as I'm anticipating that's going to generate a bit of discussion.

Incidentally, there's a Raspbian freeze coming up fairly shortly (@bennuttall tomorrow?) so in a vague effort to get GPIO Zero (and some other stuff) into the next Raspbian image I'm going to try and get a 1.3 release out rather shortly. As a result, the Servo stuff is going to be pared down to the bare minimum that I think is *definitely* fine (on the assumption that adding stuff is always easier than removing stuff as the latter is backwards incompatible) - I'm also contemplating merging ButtonBoard this evening if you're happy for that to go in as it is?

lurch:
Yeah, feel free to merge in whatever you think's suitable - afraid I don't have much time for working on gpiozero ATM :-/

docs/examples/robot_keyboard_2.py
waveform80: Ah, that would be neater ... maybe next time after a few users have posted "WTF?!" tickets next to this recipe :)

waveform80:
I know the feeling - still, very much appreciate you taking the time to pick up on all my bugs :)

PR #416, opened 2016-08-30
Fix #248

### LINKED ISSUE #248: Add Servo
We should add a Servo class, or more likely, a series of Servo classes (e.g. `Servo180`, `Servo360`...). I'm not sure what the variety of different servos is like, but perhaps there needs to be much configuration ability on init.

The way servos work is with PWM, setting a specific duty cycle to turn the servo to a specific angle. The one I have, a [Tower Pro SG90](http://www.micropik.com/PDF/SG90Servo.pdf), moves proportionately between 0 degrees and about 170 degrees when the value is set between 0.033 and 0.253 at a frequency of 100, but any value outside that does nothing.

I started playing with one using `PWMOutputDevice` and got it turning to certain angles by sending various signals via the `value` property.

When thinking about implementing a class for this, I realised I probably didn't want to extend `PWMOutputDevice` as is, because I didn't need `on()`, `off()`, etc. so I guess we should introduce a `DigitalPWMOutputDevice` class for this, and all existing classes which currently extend `PWMOutputDevice` would switch to `DigitalPWMOutputDevice`, and `Servo` would extend the new `PWMOutputDevice`. Perhaps "digital" is the wrong word in this case...

I'm not certain what the servo API should look like, but I guess ideally, you could set it by angle, and maybe setting `value` between 0 and 1 would move proportionally between the minimum and maximum positions.

waveform80:
Add Servo and AngularServo implementation along with docs and tests.
This is a deliberately minimal implementation designed to be added to as
we agree on new extensions (better than making an all-singing,
all-dancing version in which I get things wrong and then wind up making
backward incompatible changes to get it right :)


gpiozero/output_devices.py
```diff
@@ -898,3 +898,313 @@ def stop(self):
...
+            # NOTE: Why round(n, 12) here instead of 14? Angle ranges can be
+            # much larger than -1..1 so we need a little more rounding to
+            # smooth off the rough corners!
+            return round(
+                self._angular_range *
+                ((result - self._min_value) / self._value_range) +
+                self._min_angle, 12)
+
+    @angle.setter
+    def angle(self, value):
```
lurch: Given the overlap between `value` in the base class and `angle` in this class, perhaps this function would be more readable with a renamed parameter i.e. `def angle(self, value):` ?

gpiozero/output_devices.py
```diff
@@ -898,3 +898,313 @@ def stop(self):
...
+        i.e. that no control signal is being sent. Typically this means the
+        servo's position remains unchanged, but that it can be moved by hand.
+        """
+        result = self._get_value()
+        if result is None:
+            return result
+        else:
+            # NOTE: This round() only exists to ensure we don't confuse people
+            # by returning 2.220446049250313e-16 as the default initial value
+            # instead of 0. The reason _get_value and _set_value are split
```
lurch: Ummm, there isn't a `_set_value` ?

waveform80: Ahh ... there was but it wound up being redundant so I reverted that ... obviously forgot to revert that comment though! Oh well.

gpiozero/output_devices.py
```diff
@@ -898,3 +898,313 @@ def stop(self):
...
+                self._angular_range *
+                ((result - self._min_value) / self._value_range) +
+                self._min_angle, 12)
+
+    @angle.setter
+    def angle(self, value):
+        if value is None:
+            self.value = None
+        else:
+            self.value = (
```
lurch: Might be worth checking that the angle is between `_min_angle` and `_max_angle` and throwing an appropriate exception, as otherwise setting an invalid angle will result in a confusing "Servo value must be between -1 and 1, or None" error.
"Servo angle must be between _X_ and _Y_, or None" would be more appropriate IMHO.

gpiozero/output_devices.py
```diff
@@ -898,3 +898,313 @@ def stop(self):
...
+    @property
+    def max_pulse_width(self):
+        """
+        The control pulse width corresponding to the servo's maximum position,
+        measured in seconds.
+        """
+        return (self._dc_range * self.frame_width) + self.min_pulse_width
+
+    @property
+    def pulse_width(self):
```
lurch: If we're having a pulse_width getter, should there also be a pulse_width setter?

gpiozero/output_devices.py
```diff
@@ -898,3 +898,313 @@ def stop(self):
...
+            frame_width=20/1000):
+        if min_pulse_width >= max_pulse_width:
+            raise ValueError('min_pulse_width must be less than max_pulse_width')
+        if max_pulse_width >= frame_width:
+            raise ValueError('max_pulse_width must be less than frame_width')
+        self._frame_width = frame_width
+        self._min_dc = min_pulse_width / frame_width
+        self._dc_range = (max_pulse_width - min_pulse_width) / frame_width
+        self._min_value = -1
+        self._value_range = 2
```
lurch: Looks like these last two variables never change? Do they need to be instance variables? Perhaps instead you could add some static functions which convert a -1 -> +1 value into a 0 -> +1 value, and vice-versa?

gpiozero/output_devices.py
waveform80: Yeah, that'd be nicer

gpiozero/output_devices.py
waveform80: Yeah, they're left over from an earlier variant where I'd made min_value configurable as 0 or -1 on construction. I left that stuff lying around in case we want to add that back in the future, but you're right that some static functions for those fixed ranges would make stuff even simpler.

lurch: > in case we want to add that back in the future

Fair enough :)

PR #426, opened 2016-09-03
Include documentation in sdist pypi release

EdwardBetts:
I'm thinking about uploading gpiozero to Debian and I'd like to include the documentation.

Debian packages tend to be based on the pypi release, so I've changed the manifest to included it when setup.py sdist is run.

Other uses of the pypi release will probably find it helpful to have the documentation and examples.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,4 @@
 include README.rst
 recursive-include tests *.py
+recursive-include docs Makefile *.py *.rst
+recursive-include docs/images *


lurch:
I'm afraid I don't know anything about packaging (either for Debian or pypi), but how does this tie in with https://github.com/RPi-Distro/python-gpiozero/tree/master/debian ? (which I believe is used to build the Raspbian packages)

waveform80:
I'd be reticent to include our docs in the PyPI package for the simple fact they're already significantly larger than the source code and will only grow once we get around to adding more wiring diagrams (this is the primary reason they're split out in our Raspbian packages). For reference, the library package is about 70kb; the docs are 0.5Mb (I'd guesstimate they'll be about a meg or more once we've completed all the necessary wiring diagrams).

That probably doesn't sound like a lot to us, but there's still plenty out there with <3G levels of connectivity.

bennuttall:
Thanks @EdwardBetts, would be great to see the package in Debian. My first thought is would it be possible to follow suit and provide separate packages for source and docs like @waveform80 said, as we do in Raspbian? If so it should be easy enough to take what we have.

gpiozero is installed by default in Raspbian, where space is very limited (users install on SD cards and we're already too big for a 4GB card) so we tend to be very conservative about what we choose to go in. So we wouldn't include docs in one Raspbian package, so they can be an optional extra. So if you insisted on including them, that would have to be part of the Debian delta.

Related: are you able to help get the new [sense_emu](https://github.com/RPi-Distro/python-sense-emu) package into Debian?

EdwardBetts:
Don't worry about pypi, I can build the Debian package from the github release tarballs. Yes, I'll put the documentation in a separate binary package.

I'll take a look at sense_emu and I'll probably upload it to Debian.

bennuttall:
Awesome, thanks. cc @davidhoness

waveform80:
Many thanks indeed! Incidentally I must admit that whilst I tend to make .debs for all the stuff I do (picamera, sense_emu, etc.) my deb packaging skills probably leave much to be desired. Any things that're blatantly wrong, or just stuff you'd change, do let me know.

PR #281, opened 2016-04-14
Tweak the ADC classes to make the docs neater and more accurate

lurch:
If I've done anything the "wrong way" in this PR, I'm happy to make any changes requested before it gets merged.

As with #277 I'll leave it up to @waveform80 to regenerate the images and docs ;)

PR #430, opened 2016-09-08
Fix #421

### LINKED ISSUE #421: SPI tests
@waveform80 pointed out that we're missing SPI tests, which would have prevented the bug found after last night's surprise release! Let's get some tests added.

waveform80:
Added SPI tests, simplified the shared SPI software bus implementation, and fixed several errors in our implementation of various MCP3xxx chips (the protocol on the x1 and x2 series was wrong)


waveform80:
Phew ... pypy tests were a little tricky to get working (difference in object lifetimes - shared stuff is nasty) but it all looks good now.

gpiozero/spi_devices.py
```diff
@@ -128,10 +167,9 @@ def channel(self):
     @property
     def differential(self):
         """
-        If ``True``, the device is operated in pseudo-differential mode. In
-        this mode one channel (specified by the channel attribute) is read
-        relative to the value of a second channel (implied by the chip's
-        design).
+        If ``True``, the device is operated in differential mode. In this mode
+        one channel (specified by the channel attribute) is read relative to
+        the value of a second channel (implied by the chip's design).
```
lurch: I'm not sure about the removal of 'pseudo' here? AIUI the MCP30xx and MCP32xx use pseudo-differential mode (each channel's value must still be between GND and 3V3, both sharing the same GND), and it's only the MCP330xx chips that have proper differential mode (where it's only the voltage _difference_ between the two channels that is measured)?
But IANAEE - I Am Not An Electrical Engineer ;-)

Then again, to most users of GpioZero it probably doesn't make any difference. \*shrug\*

docs/images/spi_device_hierarchy.dot
```diff
@@ -29,6 +32,8 @@ digraph classes {
     MCP3202->MCP32xx;
     MCP3204->MCP32xx;
     MCP3208->MCP32xx;
+    MCP3002->MCP3xx2;
+    MCP3202->MCP3xx2;
```
lurch: Whooh, diamond inheritance! :)

gpiozero/pins/mock.py
```diff
@@ -275,3 +275,141 @@ def _set_frequency(self, value):
...
+
+class MockSPIDevice(object):
+    def __init__(
+            self, clock_pin, mosi_pin, miso_pin, select_pin=None,
+            clock_polarity=False, clock_phase=False, lsb_first=False,
+            bits_per_word=8, select_high=False):
+        self.clock_pin = MockSPIClockPin(clock_pin)
+        self.mosi_pin = None if mosi_pin is None else MockPin(mosi_pin)
+        self.miso_pin = None if miso_pin is None else MockPin(miso_pin)
+        self.select_pin = None if select_pin is None else MockSPISelectPin(select_pin)
```
lurch: Should the code be checking that (at least) clock_pin, mosi_pin and miso_pin are all non-None?

gpiozero/spi_devices.py
```diff
@@ -72,10 +111,10 @@ class AnalogInputDevice(SPIDevice):
     .. _analog to digital converters: https://en.wikipedia.org/wiki/Analog-to-digital_converter
     """
 
-    def __init__(self, bits=None, **spi_args):
-        if bits is None:
-            raise InputDeviceError('you must specify the bit resolution of the device')
+    def __init__(self, bits, **spi_args):
         self._bits = bits
+        self._min_value = -(2 ** bits)
+        self._range = 2 ** (bits + 1) - 1
```
lurch: I'm struggling to understand what you're doing here? For a non-differential ADC I'd expect a `_min_value` of `0`, and a `_range` of `2 ** bits` (e.g 0 to 1023); and for a differential ADC I'd expect a `_min_value` of `-(2 ** (bits - 1))` with the same range of `2 ** bits` (e.g. -512 to 511) ?
But you've obviously studied the data-sheets closer than me, so maybe I'm missing something?? :-S

gpiozero/spi_devices.py
```diff
@@ -92,9 +131,9 @@ def _read(self):
     def value(self):
         """
         The current value read from the device, scaled to a value between 0 and
-        1.
+        1 (or -1 to +1 for certain devices operating in differential mode).
```
lurch: If this can return **either** a 0->1 value or a -1 -> +1 value, does that imply that this class itself should 'know' if it's operating in differential mode, and offer that as a property?

gpiozero/pins/mock.py
```diff
@@ -275,3 +275,141 @@ def _set_frequency(self, value):
...
+        far. The :attr:`tx_buf` sequence gives the sequence of 1s and 0s to
+        transmit on the next clock pulses. All these attributes can be modified
+        within this method.
+
+        The :meth:`rx_word` and :meth:`tx_word` methods can also be used to
+        read and append to the buffers using integers instead of bool bits.
+        """
+        pass
+
+    def rx_word(self):
```
lurch: I'm afraid I don't fully comprehend this code, but should this `rx_word` method only be returning `bits_per_word` sized values?  (apologies if that's a naive question)
Or perhaps a list of such appropriately-sized values?

gpiozero/pins/mock.py
```diff
@@ -275,3 +275,141 @@ def _set_frequency(self, value):
...
+        return result
+
+    def tx_word(self, value, bits_per_word=None):
+        if bits_per_word is None:
+            bits_per_word = self.bits_per_word
+        bits = [0] * bits_per_word
+        for bit in range(bits_per_word):
+            bits[bit] = value & 1
+            value >>= 1
+        assert not value
```
lurch: So I guess this would assert if `value > 2 ** bits_per_word` ?

gpiozero/pins/mock.py
```diff
@@ -275,3 +275,141 @@ def _set_frequency(self, value):
...
+        if self in self.clock_pin.spi_devices:
+            self.clock_pin.spi_devices.remove(self)
+        if self.select_pin is not None:
+            self.select_pin.spi_device = None
+
+    def on_select(self):
+        if self.select_pin.state == self.select_high:
+            self.on_start()
+
+    def on_clock(self):
```
lurch: I wonder if it would make sense for there to be an `on_word` callback for when `self.rx_bit == self.bits_per_word` ?

gpiozero/pins/mock.py
```diff
@@ -275,3 +275,141 @@ def _set_frequency(self, value):
...
+        self.rx_bit = 0
+        self.rx_buf = []
+        self.tx_buf = []
+
+    def on_bit(self):
+        """
+        Override this in descendents to react to receiving a bit.
+
+        The :attr:`rx_bit` attribute gives the index of the bit received (this
+        is reset to 0 by default by :meth:`on_select`). The :attr:`rx_buf`
```
lurch: Should this be updated to say `reset to 0 by default by :meth:on_start` ? (which is what the code here is actually doing)

gpiozero/spi_devices.py
```diff
@@ -196,43 +279,44 @@ def __init__(self, channel=0, differential=False, **spi_args):
         super(MCP33xx, self).__init__(channel, 12, differential, **spi_args)
 
     def _read(self):
-        # MCP3304/02 protocol looks like the following:
+        if self.differential:
+            result = self._words_to_int(
+                self._spi.transfer(self._send())[-2:], self.bits + 1)
+            # Account for the sign bit
+            if result > 4095:
+                return -(8192 - result)
```
lurch: Would these numbers be more readable if written as explicit powers of two? ;)

gpiozero/spi_devices.py
```diff
@@ -351,23 +455,33 @@ class MCP3208(MCP32xx):
...
+        #
+        #     Byte        0        1
+        #     ==== ======== ========
+        #     Rx   xx0SRRRR RRRRRRRR
+        result = self._words_to_int(self._spi.read(2), 13)
+        # Account for the sign bit
+        if result > 4095:
+            return -(8192 - result)
+        else:
+            return result
```
lurch: In the spirit of DRY, I wonder if this `result > 4095` snippet could be shared in a common function with the base `MCP33xx._read` method?

tests/test_spi.py
```diff
@@ -0,0 +1,170 @@
...
+        with SPI(shared=True) as device:
+            assert isinstance(device, SharedSPIHardwareInterface)
+        with pytest.raises(ValueError):
+            SPI(port=1)
+        with pytest.raises(ValueError):
+            SPI(device=2)
+        with pytest.raises(ValueError):
+            SPI(port=0, clock_pin=12)
+        with pytest.raises(ValueError):
+            SPI(foo='bar')
```
lurch: No hardware tests using the `miso_pin` ?

tests/test_spi_devices.py
```diff
@@ -0,0 +1,339 @@
...
+
+def test_MCP3304():
+    mock = MockMCP3304(11, 10, 9, 8)
+    with pytest.raises(ValueError):
+        MCP3304(channel=9)
+    with MCP3304(channel=5) as pot:
+        single_mcp_test(mock, pot, 5, 12)
+    with MCP3304(channel=5, differential=True) as pot:
+        differential_mcp_test(mock, pot, 5, 4, 12, full=True)
+
```
lurch: Might be nice to have some MCP3008 tests using some non-default `**spi_args` ?

(Haven't read through all the testcode yet, will try to do so tomorrow. Now hitting the sack.)

lurch:
Lots of code, which is why I added so many (possibly stupid) comments.
All seems to look very comprehensive :) :+1: (in typical @waveform80 fashion!)

gpiozero/spi_devices.py
waveform80: Damn ... that's probably a copy'n'paste mistake. It is meant to read pseudo- on the MCP3xxx devices and not on the MCP33xx ones.

docs/images/spi_device_hierarchy.dot
waveform80: Yeah, 'fraid so - but it's the simplest way to implement the 3xx2 protocol on both the 30xx and 32xx branches. Turns out there's nothing complex in this particular case as both sides of the diamond override completely separate methods so there's no complex MRO to figure out.

gpiozero/pins/mock.py
waveform80: Erm ... it should probably check clock_pin is not none, but I've been doing a bit of reading about SPI. Turns out there's some chips that do away with the MOSI or MISO pin entirely. For example, the MCP3xx1 chips don't bother with a MOSI pin at all (because they don't read any data, they simply return it). There's other chips that don't bother with a select line (they assume they're the only thing in the system). So the only guaranteed mandatory pin is clock (strictly clock and one of MOSI or MISO). Should probably implement some slightly more complex test here...

gpiozero/spi_devices.py
waveform80: Ahh, not quite. For the pseudo-differential devices, they actually operate 0..2**bits just like in single mode. They're called pseudo-differential because they can't do <0 (where IN- is greater than IN+).

For the full-differential ADCs they operate in, e.g. 12-bits in single-ended mode, and 13-bits (2's complement) in full-different mode. So, for the sake of ease, we treat them as 12-bit whatever mode they're in and assume that their range is -2**bits .. (2**bits) - 1. When they're in single ended mode this is fine because the output will never be <0 (it'll just occupy the top half of the range). In full differential mode, this is also fine as it's the full range. Hence, the full numeric range of their output is 2 ** (bits + 1) - 1 values.

gpiozero/spi_devices.py
waveform80: Possibly ... At the moment I don't know whether the MCP ADCs are representative of ADCs generally and thus whether differential can be a simple bool or whether we might need something more complex in future. Hence, for now I've left this as "for future consideration".

gpiozero/pins/mock.py
waveform80: No, I should add some proper docs to rx_word (and tx_word). rx_word will return a single integer representing the value of all received bits in the buffer (however large that is). It could be limited to word-size, but then a lot of chips don't seem terribly bothered about their word boundaries (the MCP ADCs don't seem to care about word boundaries at all). So I'm using "word" in the sense of "I want to read what's in the buffer as a word ... but I don't really care what size it is".

gpiozero/pins/mock.py
waveform80: Yup, that's the idea - basically if you've fed it a particular value, expecting it to queue up all those bits in the tx buffer, but your value won't fit in the word size you've specified (or implied) then this'll assert.

gpiozero/pins/mock.py
waveform80: Heh ... I actually started out with that when I first wrote the tests. I then discovered that all our currently implemented ADCs don't give a damn about word boundaries and that it was much easier to work at the bit level so I removed `on_word` until I can find a use for it (I assume there's *some* SPI chips that obey their word boundaries ... we just haven't got any yet :).

gpiozero/pins/mock.py
waveform80: Doh, absolutely right - it originally was in `on_select` (when I wrote the docs) but then I had to take account of select-line polarity so it got moved and I forgot to update the docs.

gpiozero/spi_devices.py
waveform80: Erm ... possibly? It'd probably make it more obvious that it's **12 and **13 respectively (which ties into the bit-ness of the ADC in question). Yeah, good idea.

gpiozero/spi_devices.py
waveform80: Yeah, that's a good idea.

tests/test_spi.py
waveform80: Ooops - I meant to make that second mosi_pin test miso_pin instead.

tests/test_spi_devices.py
waveform80: Yeah, I changed up the channels a bit and then forgot to go perturb the pins. Oh well, more cleaning up :)

waveform80:
No stupid comments there - always good to know someone's watching! I'm currently hacking together the pretty-print pinout function (with colours :) but I'm going to be swamped with some non-Pi stuff this afternoon and weekend, so if you want to have a crack at tidying up any of the loose ends do feel free!

gpiozero/spi_devices.py
lurch: I was mainly asking from a sourcetools PoV, where you need to know what range of `values` you can expect...
With the code currently you'd need to scale the (potential) -1 -> +1 range down to 0 -> 1 if connecting it to e.g. a PWMLED, and that in turn means the LED would never be below 50% brightness if used with a non-differential ADC.

Given the audience that GpioZero is aimed at, I wonder if any of the differential ADC code will actually *ever* get used??

Hmm, that's just given me another idea... I wonder if there could be some kind of `metavalue` property, that describes what type / range the value and values properties return, so that the sourcetools functions could check that things are 'hooked up correctly' and possibly auto-scale where appropriate? ;)  (Worth opening an issue for further discussion? Or is it just overcomplicating things)

gpiozero/spi_devices.py
lurch: Okay, so for full-differential devices the range is `2 ** (bits + 1) - 1`, but presumably for pseudo-differential ADCs (or non-differential ADCs) the range is still `2 ** bits - 1` ?

Or is the code simply over-simplifying, in order to keep code-reuse to a maximum? (and specialisation to a minimum)

gpiozero/pins/mock.py
lurch: I guess I was just curious about the mismatch of `tx_word` using `bits_per_word`, but `rx_word` not using it.

lurch:
I find reviewing the code-changes a great way to try to grok what it's doing ;-)  And it's nice to see that at least a couple of my comments were useful :)
Afraid I'm also busy this weekend, but _might_ be able to squeeze in a bit of time on Sunday.

lurch:
I hope the colours will be turn-off-able via a kwarg for people with dumb terminals, e.g. a serial console.

tests/test_spi_devices.py
```diff
@@ -0,0 +1,339 @@
...
+            result = clamp(result, -self.vref, self.vref)
+        else:
+            result = clamp(self.channels[channel], 0, self.vref)
+        result = scale(result, self.vref, self.bits)
+        if result < 0:
+            result += 8192
+        self.tx_word(result, self.bits + 3)
+
+
+class MockMCP3001(MockMCP3xx1):
```
lurch: All these MockMCP classes should probably be refactored to match what I did in #281

tests/test_spi_devices.py
```diff
@@ -0,0 +1,339 @@
...
+    vmin = -(2 ** bits)
+    vmax = -vmin - 1
+    vrange = vmax - vmin
+    return int(((v + 1) / 2.0) * vrange + vmin)
+
+
+class MockMCP3xxx(MockSPIDevice):
+    def __init__(
+            self, clock_pin, mosi_pin, miso_pin, select_pin=None,
+            channels=8, bits=10):
```
lurch: I'm not sure if it makes sense to provide default-values for the `channels` and `bits` args?
But it obviously makes sense to have them after the optional `select_pin` arg, so maybe you could default them to None, and then assert if they're still None, like you do in RGBLED? https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L556

tests/test_spi_devices.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+class MockMCP33xx(MockMCP3xxx):
+    def __init__(
+            self, clock_pin, mosi_pin, miso_pin, select_pin=None,
+            channels=8):
+        super(MockMCP33xx, self).__init__(
+            clock_pin, mosi_pin, miso_pin, select_pin, channels, 12)
+
+    def on_result(self, differential, channel):
```
lurch: Seems to be scope here for code-sharing with the base MockMCP3xxx.on_result

tests/test_spi_devices.py
```diff
@@ -0,0 +1,339 @@
...
+        assert pot.raw_value == -scale
+        assert isclose(pot.value, -1.0, abs_tol=tolerance)
+    else:
+        assert pot.raw_value == 0
+        assert isclose(pot.value, 0.0, abs_tol=tolerance)
+
+
+def test_MCP3001():
+    mock = MockMCP3001(11, 10, 9, 8)
+    with MCP3001() as pot:
```
lurch: I'm guessing that `pot` implies Potentiometer. IMHO using a single potentiometer to test a differential ADC wouldn't work, so maybe the differential ADC code should use a variable named something other than `pot`, so that the code is easier to read?

tests/test_spi_devices.py
```diff
@@ -0,0 +1,339 @@
...
+        differential_mcp_test(mock, pot, 0, 1, 10)
+
+def test_MCP3002():
+    mock = MockMCP3002(11, 10, 9, 8)
+    with pytest.raises(ValueError):
+        MCP3002(channel=5)
```
lurch: I think this should have been `channel=3` ?  (and similarly for the other test_MCPxxx2 functions)

gpiozero/pins/mock.py
waveform80: Yeah, tx_word has that limit mainly because I think it might be useful in future. At the moment I think all transmissions for the MCP ADCs use an explicitly specified (over-sized) bits_per_word parameter (because they're all using byte-sized words, but they all return larger-than-bytes values).

gpiozero/spi_devices.py
waveform80: Over-simplifying. The *effective* range for pseudo-differential and single-ended ADCs is indeed 0..(2**bits)-1 but given that's a perfect subset of the range -2**bits..(2**bits)-1 we can just treat it as the same range and assume they'll never give us a negative value to normalize (which they won't because there's no 2's complement handling in them).

gpiozero/spi_devices.py
waveform80: Ahhh ... good point. And a property which advertises the range of `value` ... that's a very interesting idea. That needs a little thought as to how it handles edge cases like `None` (in `Servo`) but it's definitely worth exploring. I very much like the idea of the `source` property throwing a warning when connecting to a `values` which has a larger range than it can handle.

gpiozero/spi_devices.py
lurch: Okay, thanks for clearing it up for me :)

PR #431, opened 2016-09-09
Fix #354

### LINKED ISSUE #354: pinout function
Just playing around with `pi_info()` and wondered about adding a pinout function to show the pinout for the current Pi. For example:

```python
def pinout():
    info = pi_info()
    pins = info.headers['P1']
    for i in range(1, len(pins), 2):
        print(pins[i].function, pins[i+1].function)
```
```
3V3 5V
GPIO0 5V
GPIO1 GND
GPIO4 GPIO14
GND GPIO15
GPIO17 GPIO18
GPIO21 GND
GPIO22 GPIO23
3V3 GPIO24
GPIO10 GND
GPIO9 GPIO25
GPIO11 GPIO8
GND GPIO7
```

(yes, I'm on a Pi 1 rev1)

It just needs padding out. And I'm sure @waveform80 would have a better way of pairing them.

Any thoughts?

waveform80:
Actually, "fix" doesn't really do this justice. This is closer to "nukes
from orbit" ...

### DIFF
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -7,6 +7,11 @@ from __future__ import (
 str = type('')
 
 import io
+import os
+import sys
+from textwrap import dedent
+from itertools import cycle
+from operator import attrgetter
 from collections import namedtuple
 
 from ..exc import PinUnknownPi, PinMultiplePins, PinNoPins
@@ -65,6 +70,110 @@ GPIO42 = 'GPIO42'
 GPIO43 = 'GPIO43'
 GPIO44 = 'GPIO44'
 GPIO45 = 'GPIO45'
+
+# Board layout ASCII art
+
+REV1_BOARD = """\
+{style:white on green}+------------------{style:black on white}| |{style:white on green}--{style:on cyan}| |{style:on green}------+{style:reset}
+{style:white on green}| {P1:{style} col2}{style:white on green} P1 {style:black on yellow}|C|{style:white on green}  {style:on cyan}|A|{style:on green}      |{style:reset}
+{style:white on green}| {P1:{style} col1}{style:white on green}    {style:black on yellow}+-+{style:white on green}  {style:on cyan}+-+{style:on green}      |{style:reset}
+{style:white on green}|                                |{style:reset}
+{style:white on green}|                {style:on black}+---+{style:on green}          {style:black on white}+===={style:reset}
+{style:white on green}|                {style:on black}|SoC|{style:on green}          {style:black on white}| USB{style:reset}
+{style:white on green}|   {style:on black}|D|{style:on green} {style:bold}Pi Model{style:normal} {style:on black}+---+{style:on green}          {style:black on white}+===={style:reset}
+{style:white on green}|   {style:on black}|S|{style:on green} {style:bold}{model:2s} V{pcb_revision:3s}{style:normal}                  |{style:reset}
+{style:white on green}|   {style:on black}|I|{style:on green}                  {style:on black}|C|{style:black on white}+======{style:reset}
+{style:white on green}|                        {style:on black}|S|{style:black on white}|   Net{style:reset}
+{style:white on green}|                        {style:on black}|I|{style:black on white}+======{style:reset}
+{style:black on white}=pwr{style:on green}             {style:on white}|HDMI|{style:white on green}          |{style:reset}
+{style:white on green}+----------------{style:black on white}|    |{style:white on green}----------+{style:reset}"""
+
+REV2_BOARD = """\
+{style:white on green}+------------------{style:black on white}| |{style:white on green}--{style:on cyan}| |{style:on green}------+{style:reset}
+{style:white on green}| {P1:{style} col2}{style:white on green} P1 {style:black on yellow}|C|{style:white on green}  {style:on cyan}|A|{style:on green}      |{style:reset}
+{style:white on green}| {P1:{style} col1}{style:white on green}    {style:black on yellow}+-+{style:white on green}  {style:on cyan}+-+{style:on green}      |{style:reset}
+{style:white on green}|    {P5:{style} col1}{style:white on green}                        |{style:reset}
+{style:white on green}| P5 {P5:{style} col2}{style:white on green}        {style:on black}+---+{style:on green}          {style:black on white}+===={style:reset}
+{style:white on green}|                {style:on black}|SoC|{style:on green}          {style:black on white}| USB{style:reset}
+{style:white on green}|   {style:on black}|D|{style:on green} {style:bold}Pi Model{style:normal} {style:on black}+---+{style:on green}          {style:black on white}+===={style:reset}
+{style:white on green}|   {style:on black}|S|{style:on green} {style:bold}{model:2s} V{pcb_revision:3s}{style:normal}                  |{style:reset}
+{style:white on green}|   {style:on black}|I|{style:on green}                  {style:on black}|C|{style:black on white}+======{style:reset}
+{style:white on green}|                        {style:on black}|S|{style:black on white}|   Net{style:reset}
+{style:white on green}|                        {style:on black}|I|{style:black on white}+======{style:reset}
+{style:black on white}=pwr{style:on green}             {style:on white}|HDMI|{style:white on green}          |{style:reset}
+{style:white on green}+----------------{style:black on white}|    |{style:white on green}----------+{style:reset}"""
+
+A_BOARD = """\
+{style:white on green}+------------------{style:black on white}| |{style:white on green}--{style:on cyan}| |{style:on green}------+{style:reset}
+{style:white on green}| {P1:{style} col2}{style:white on green} P1 {style:black on yellow}|C|{style:white on green}  {style:on cyan}|A|{style:on green}      |{style:reset}
+{style:white on green}| {P1:{style} col1}{style:white on green}    {style:black on yellow}+-+{style:white on green}  {style:on cyan}+-+{style:on green}      |{style:reset}
+{style:white on green}|    {P5:{style} col1}{style:white on green}                        |{style:reset}
+{style:white on green}| P5 {P5:{style} col2}{style:white on green}        {style:on black}+---+{style:on green}          {style:black on white}+===={style:reset}
+{style:white on green}|                {style:on black}|SoC|{style:on green}          {style:black on white}| USB{style:reset}
+{style:white on green}|   {style:on black}|D|{style:on green} {style:bold}Pi Model{style:normal} {style:on black}+---+{style:on green}          {style:black on white}+===={style:reset}
+{style:white on green}|   {style:on black}|S|{style:on green} {style:bold}{model:2s} V{pcb_revision:3s}{style:normal}                  |{style:reset}
+{style:white on green}|   {style:on black}|I|{style:on green}                  {style:on black}|C|{style:on green}     |{style:reset}
+{style:white on green}|                        {style:on black}|S|{style:on green}     |{style:reset}
+{style:white on green}|                        {style:on black}|I|{style:on green}     |{style:reset}
+{style:black on white}=pwr{style:on green}             {style:on white}|HDMI|{style:white on green}          |{style:reset}
+{style:white on green}+----------------{style:black on white}|    |{style:white on green}----------+{style:reset}"""
+
+BPLUS_BOARD = """\
+{style:white on green},--------------------------------.{style:reset}
+{style:white on green}| {P1:{style} col2}{style:white on green} P1     {style:black on white}+===={style:reset}
+{style:white on green}| {P1:{style} col1}{style:white on green}        {style:black on white}| USB{style:reset}
+{style:white on green}|                             {style:black on white}+===={style:reset}
+{style:white on green}|      {style:bold}Pi Model {model:2s} V{pcb_revision:3s}{style:normal}          |{style:reset}
+{style:white on green}|      {style:on black}+----+{style:on green}                 {style:black on white}+===={style:reset}
+{style:white on green}| {style:on black}|D|{style:on green}  {style:on black}|SoC |{style:on green}                 {style:black on white}| USB{style:reset}
+{style:white on green}| {style:on black}|S|{style:on green}  {style:on black}|    |{style:on green}                 {style:black on white}+===={style:reset}
+{style:white on green}| {style:on black}|I|{style:on green}  {style:on black}+----+{style:on green}                    |{style:reset}
+{style:white on green}|                   {style:on black}|C|{style:on green}     {style:black on white}+======{style:reset}
+{style:white on green}|                   {style:on black}|S|{style:on green}     {style:black on white}|   Net{style:reset}
+{style:white on green}| {style:black on white}pwr{style:white on green}        {style:black on white}|HDMI|{style:white on green} {style:on black}|I||A|{style:on green}  {style:black on white}+======{style:reset}
+{style:white on green}`-{style:black on white}| |{style:white on green}--------{style:black on white}|    |{style:white on green}----{style:on black}|V|{style:on green}-------'{style:reset}"""
+
+APLUS_BOARD = """\
+{style:white on green},--------------------------.{style:reset}
+{style:white on green}| {P1:{style} col2}{style:white on green} P1  |{style:reset}
+{style:white on green}| {P1:{style} col1}{style:white on green}     |{style:reset}
+{style:white on green}|                          |{style:reset}
+{style:white on green}|      {style:bold}Pi Model {model:2s} V{pcb_revision:3s}{style:normal}    |{style:reset}
+{style:white on green}|      {style:on black}+----+{style:on green}           {style:black on white}+===={style:reset}
+{style:white on green}| {style:on black}|D|{style:on green}  {style:on black}|SoC |{style:on green}           {style:black on white}| USB{style:reset}
+{style:white on green}| {style:on black}|S|{style:on green}  {style:on black}|    |{style:on green}           {style:black on white}+===={style:reset}
+{style:white on green}| {style:on black}|I|{style:on green}  {style:on black}+----+{style:on green}              |{style:reset}
+{style:white on green}|                   {style:on black}|C|{style:on green}    |{style:reset}
+{style:white on green}|                   {style:on black}|S|{style:on green}    |{style:reset}
+{style:white on green}| {style:black on white}pwr{style:white on green}        {style:black on white}|HDMI|{style:white on green} {style:on black}|I||A|{style:on green} |{style:reset}
+{style:white on green}`-{style:black on white}| |{style:white on green}--------{style:black on white}|    |{style:white on green}----{style:on black}|V|{style:on green}-'{style:reset}"""
+
+ZERO12_BOARD = """\
+{style:white on green},-------------------------.{style:reset}
+{style:white on green}| {P1:{style} col2}{style:white on green} P1 |{style:reset}
+{style:white on green}| {P1:{style} col1}{style:white on green}    |{style:reset}
+{style:black on white}---+{style:white on green}       {style:on black}+----+{style:on green} {style:bold}PiZero{style:normal}  |{style:reset}
+{style:black on white} sd|{style:white on green}       {style:on black}|SoC |{style:on green}  {style:bold}V{pcb_revision:3s}{style:normal}   |{style:reset}
+{style:black on white}---+|hdmi|{style:white on green} {style:on black}+----+{style:on green} {style:black on white}usb{style:on green} {style:black on white}pwr{style:white on green} |{style:reset}
+{style:white on green}`---{style:black on white}|    |{style:white on green}--------{style:black on white}| |{style:white on green}-{style:black on white}| |{style:white on green}-'{style:reset}"""
+
+ZERO13_BOARD = """\
+{style:white on green}.-------------------------.{style:reset}
+{style:white on green}| {P1:{style} col2}{style:white on green} P1 |{style:reset}
+{style:white on green}| {P1:{style} col1}{style:white on green}   {style:black on white}|c{style:reset}
+{style:black on white}---+{style:white on green}       {style:on black}+----+{style:on green} {style:bold}PiZero{style:normal} {style:black on white}|s{style:reset}
+{style:black on white} sd|{style:white on green}       {style:on black}|SoC |{style:on green}  {style:bold}V{pcb_revision:3s}{style:normal}  {style:black on white}|i{style:reset}
+{style:black on white}---+|hdmi|{style:white on green} {style:on black}+----+{style:on green} {style:black on white}usb{style:on green} {style:on white}pwr{style:white on green} |{style:reset}
+{style:white on green}`---{style:black on white}|    |{style:white on green}--------{style:black on white}| |{style:white on green}-{style:black on white}| |{style:white on green}-'{style:reset}"""
+
+CM_BOARD = """\
+{style:white on green}+-----------------------------------------------------------------------------------------------------------------------+{style:reset}
+{style:white on green}| Raspberry Pi Compute Module                                                                                           |{style:reset}
+{style:white on green}|                                                                                                                       |{style:reset}
+{style:white on green}| You were expecting more detail? Sorry, the Compute Module's a bit hard to do right now!                               |{style:reset}
+{style:white on green}|                                                                                                                       |{style:reset}
+{style:white on green}|                                                                                                                       |{style:reset}
+{style:white on green}||||||||||||||||||||-||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||{style:reset}"""
 
 # Pin maps for various board revisions and headers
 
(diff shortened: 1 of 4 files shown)


lurch:
Few quick things I've spotted:

1. The labels on the Zero boards should be upper-cased to match the other boards

1. When `pprint`-ing the boards with both P1 and P5, the P1 and P5 pinouts get printed in an arbitrary order - I guess a quick `sort` in the right place would fix this.

1. When `pprint`-ing on a terminal with a white background and black foreground, some of the colours in the ASCII-art aren't displayed correctly.

lurch:
And just out of curiosity (and unrelated to the changes in this PR) it looks like the pin-names on the CM header have had their underscores replaced with spaces (e.g. 'HDMI CK N'), when compared to the CM schematics (e.g. 'HDMI_CK_N'). Was that intentional? Is it desired?

lurch:
...under Python2.7
```
>>> print('{0:col2}'.format(pi_info('e').headers['P1']))
  File "gpiozero/pins/data.py", line 592, in __format__
    elif format_spec.startswith('col') and format_spec[3:].isnumeric():
AttributeError: 'str' object has no attribute 'isnumeric'
```

waveform80:
> 1. The labels on the Zero boards should be upper-cased to match the other boards

Ah, that was my (apparently too) subtle method of trying to indicate that they're mini-HDMI and micro-USB ports rather than the full-sized variants on the other boards.

> 2. When pprint-ing the boards with both P1 and P5, the P1 and P5 pinouts get printed in an arbitrary order - I guess a quick sort in the right place would fix this.

Ah, good catch - I'll get that fixed in a mo.

> 3. When pprint-ing on a terminal with a white background and black foreground, some of the colours in the ASCII-art aren't displayed correctly.

Not much I can do about that: the terminal colors aren't fixed (e.g. you can generally configure color 1, "red" to be anything you like, and so on). When you set some terminals to white background all they do is flip their internal color definitions so white is black and vice versa. You do get more control (I want an actual red dammit!) if you assume a 256-color capable terminal but that's generally a risky assumption (some are but don't advertise it, some advertise it and aren't). Basically terminal colors come down to: stuff will work if you're using defaults but if you fiddle with the settings all bets are off.

I should probably add some formatting options to force non-colored output even on platforms that are color capable (i.e. anything except Windows). Would be useful anyway.

waveform80:
> And just out of curiosity (and unrelated to the changes in this PR) it looks like the pin-names on the CM header have had their underscores replaced with spaces (e.g. 'HDMI CK N'), when compared to the CM schematics (e.g. 'HDMI_CK_N'). Was that intentional? Is it desired?

I originally did that when copying them in (I think I did a search'n'replace on the copy'n'pasted stuff) on the assumption the underscores were just representing spaces anyway. They could probably do with expansion though ("HDMI CK N" to "HDMI CLOCK-", "HDMI CK P" to "HDMI CLOCK+" or something similar).

waveform80:
> ...under Python2.7
> ```
> >>> print('{0:col2}'.format(pi_info('e').headers['P1']))
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "gpiozero/pins/data.py", line 592, in __format__
>     elif format_spec.startswith('col') and format_spec[3:].isnumeric():
> AttributeError: 'str' object has no attribute 'isnumeric'
> ```

Erm...

```
(gpio-zero) pi@raspberrypi ~/gpio-zero $ python
Python 2.7.9 (default, Mar  8 2015, 00:52:26) 
[GCC 4.9.2] on linux2
>>> from gpiozero import *                                                                                                    
>>> pi_info().pprint()
,--------------------------------.
| oooooooooooooooooooo P1     +====
| 1ooooooooooooooooooo        | USB
|                             +====
|        Pi Model 3B V1.2        |
|      +----+                 +====
| |D|  |SoC |                 | USB
| |S|  |    |                 +====
| |I|  +----+                    |
|                   |C|     +======
|                   |S|     |   Net
| PWR        |HDMI| |I||A|  +======
`-| |--------|    |----|V|-------'


P1:
   3V3  (1) (2)  5V    
 GPIO2  (3) (4)  5V    
 GPIO3  (5) (6)  GND   
 GPIO4  (7) (8)  GPIO14
   GND  (9) (10) GPIO15
GPIO17 (11) (12) GPIO18
GPIO27 (13) (14) GND   
GPIO22 (15) (16) GPIO23
   3V3 (17) (18) GPIO24
GPIO10 (19) (20) GND   
 GPIO9 (21) (22) GPIO25
GPIO11 (23) (24) GPIO8 
   GND (25) (26) GPIO7 
 GPIO0 (27) (28) GPIO1 
 GPIO5 (29) (30) GND   
 GPIO6 (31) (32) GPIO12
GPIO13 (33) (34) GND   
GPIO19 (35) (36) GPIO16
GPIO26 (37) (38) GPIO20
   GND (39) (40) GPIO21
>>> 
```

Hmm, I wonder if 2.7.something introduced isnumeric. Let me try on an old wheezy box ...

```
dave@gonzo ~ $ python
Python 2.7.3 (default, Jun 22 2016, 03:14:32) 
[GCC 4.6.3] on linux2
>>> '1'.isnumeric()
AttributeError: 'str' object has no attribute 'isnumeric'
```

Damn. Okay, I'll figure out something for that too...

lurch:
>> it looks like the pin-names on the CM header have had their underscores replaced with spaces (e.g. 'HDMI CK N'), when compared to the CM schematics (e.g. 'HDMI_CK_N'). Was that intentional? Is it desired?

> I originally did that when copying them in (I think I did a search'n'replace on the copy'n'pasted stuff) on the assumption the underscores were just representing spaces anyway. They could probably do with expansion though ("HDMI CK N" to "HDMI CLOCK-", "HDMI CK P" to "HDMI CLOCK+" or something similar).

My preference would be for the pin-names in GpioZero to match the pin-names in the schematic, i.e. 'HDMI_CK_N'. What thinks @bennuttall ?

> Hmm, I wonder if 2.7.something introduced isnumeric. Let me try on an old wheezy box ...

It's not a 2.7.x thing, it's a unicode thing: http://stackoverflow.com/questions/23183868/why-isnt-isnumeric-working

> I should probably add some formatting options to force non-colored output

Agreed. And maybe also an option for only printing the pin-labels and skipping the ascii-art?
And I wonder if it'd be worth customising the output of `str(pi_info(2))` to not display the ascii-art code?

>> When pprint-ing on a terminal with a white background and black foreground, some of the colours in the ASCII-art aren't displayed correctly.

> Not much I can do about that

Actually, I think you can. If you run e.g. `xterm -fg black -bg white` and then try e.g. `pi_info('a22082').pprint()`, you'll see that the 'AV' label gets printed correctly, and the 'I' of CSI gets printed correctly, but the 'CS' of CSI doesn't get displayed. I think the fix is for your ASCII-art to always explicitly set the foreground colour whenever you change the background colour, rather than assuming the foreground colour is white. IYSWIM!

lurch:
> Ah, that was my (apparently too) subtle method of trying to indicate that they're mini-HDMI and micro-USB ports rather than the full-sized variants on the other boards.

In that case, should `PWR` be lower-case on all boards?

lurch:
...and I've just noticed that `PI_REVISIONS` doesn't include 0xa01040 from http://elinux.org/RPi_HardwareHistory ?

waveform80:
> It's not a 2.7.x thing, it's a unicode thing: http://stackoverflow.com/questions/23183868/why-isnt-isnumeric-working

Ah, and of course I was testing in an 2.x environment where the default string had been coerced to unicode via the future import. Eurgh. Still, easy enough workaround - just use `isdigit` instead (isnumeric is a bit "wide" for our purposes anyway) which exists on both str and unicode in 2.x.

> Agreed. And maybe also an option for only printing the pin-labels and skipping the ascii-art?

Not sure I understand? You mean print just the pin-out diagrams? There's already an option for that: use pprint on just the header you're interested in:

```python
>>> pi_info('0002').headers['P1'].pprint()
```

So, `pprint` can be called for the board as a whole ("pretty-print everything for me!") or for a specific header ("show me the layout for header P5"). And there's some customization one can perform too:

```python
>>> print('{0:full}'.format(pi_info('0002'))) # same as pi_info('0002').pprint()
>>> print('{0:board}'.format(pi_info('0002'))) # show me just the board ASCII art
>>> print('{0:headers}'.format(pi_info('0002'))) # show me just the headers
>>> print('{0:full}'.format(pi_info('0002').headers['P1'])) # same as pi_info('0002').headers['P1'].pprint()
>>> print('{0:col1}'.format(pi_info('0002').headers['P1'])) # render just the first column of pins for P1
```

I'm just adding in some format specs for mono as well (I'll add those as params to the `pprint` methods).

> And I wonder if it'd be worth customising the output of str(pi_info(2)) to not display the ascii-art code?

Good idea - it's way too verbose otherwise. I might use the "standard" `pprint` on the output too as it's a bit too jumbled to make sense of.

> In that case, should PWR be lower-case on all boards?

Good point!

waveform80:
> ...and I've just noticed that PI_REVISIONS doesn't include 0xa01040 from http://elinux.org/RPi_HardwareHistory ?

Looks like Phil's added that one since I last looked: http://elinux.org/index.php?title=RPi_HardwareHistory&type=revision&diff=414791&oldid=412436

lurch:
Yeah, `print('{0:headers}'.format(pi_info('0002')))` (print just the headers) is what I was looking for. It's a bit of a mouthful though - is it worth adding a simple wrapper function? Or a `headers_only=True` or similar param to the `pprint` function?

waveform80:
I think I'll leave out `0xa01040` for now and let the parsing routine handle it (it gets everything basically right but obviously the release date is a guess based on the other 2B models).

> Actually, I think you can. If you run e.g. xterm -fg black -bg white and then try e.g. pi_info('a22082').pprint(), you'll see that the 'AV' label gets printed correctly, and the 'I' of CSI gets printed correctly, but the 'CS' of CSI doesn't get displayed. I think the fix is for your ASCII-art to always explicitly set the foreground colour whenever you change the background colour, rather than assuming the foreground colour is white. IYSWIM!

Ahh, the "default" colors (ANSI 39 and 49), which aren't necessarily defined as 7 and 0. Hmmm, things are a tad more complex than the old BBS days. I'll see what I can do!

waveform80:
Okay, that's a fairly extensive re-work. I've divided stuff about as cleanly as I can:

* The style stuff (including term-detection) is all stuffed in a single class
* repr(pi_info()) is now minimal - both headers and board are "..."
* non-color output can be requested from any pprint method (and via the format specs if you so desire)
* colors should work with inverse terminals (well, xterm at least - I make no guarantees about all terms - it's too much of a minefield :)
* headers output is ordered by header name
* board specs are included in the full output (the pprint is meant to cover the whole structure so I figured it made sense to include stuff like memory, soc, etc.)
* I've chopped out everything except old revisions from the lookup table (why redundantly define stuff when the algorithm in the new revision parser can handle it)
* pwr is lowercase everywhere :)
* I haven't added a specified method for all-headers output - there's already a pprint for individual headers which is probably sufficient for the vast majority of use-cases, and for the few remaining use-cases either the format-spec should suffice (wordy though it is), or the full output will do

lurch:
I wonder if the PiZero 1.3 should have its 'CSI' label in lower-case, given that it's a smaller connector than on the other Pis?

gpiozero/pins/data.py
```diff
@@ -422,13 +760,183 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
 
     .. attribute:: headers
 
-        A dictionary which maps header labels to dictionaries which map
-        physical pin numbers to :class:`PinInfo` tuples. For example, to obtain
-        information about pin 12 on header P1 you would query
-        ``headers['P1'][12]``.
+        A dictionary which maps header labels to :class:`HeaderInfo` tuples.
+        For example, to obtain about header P1 you would query
```
lurch: 'obtain information' ?

gpiozero/pins/data.py
```diff
@@ -298,7 +489,144 @@ class PinInfo(namedtuple('PinInfo', (
...
+        lines = [
+            ' '.join(
+                '{cell.style}{cell.content:{cell.align}{width}s}{style:reset}'.format(
+                    cell=cell, width=width, style=style)
+                for cell, width, align in zip(line, col_lens, cycle('><')))
+            for line in lines
+            ]
+        return '\n'.join(lines)
+
+    def _format_row(self, row, style):
```
lurch: Should `_format_row` and `_format_col` raise an exception if passed an invalid row or column?
`print('{0:col4}'.format(pi_info('0002').headers['P1']))` _appears_ to work but obviously produces invalid output.

gpiozero/pins/data.py
```diff
@@ -317,17 +645,27 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
...
     a tuple, it is strongly recommended that you use the following named
-    attributes to access the data contained within.
+    attributes to access the data contained within. The object can be used
+    in format strings with various custom format specifications::
+
+        from gpiozero import *
+
+        print('{0:full}'.format(pi_info()))
+        print('{0:board}'.format(pi_info()))
+        print('{0:headers}'.format(pi_info()))
```
lurch: Should probably also include a usage-example of '{0:specs}' ?

lurch:
> non-color output can be requested from any pprint method (and via the format specs if you so desire)

Could you document how to switch off colour using format-specs?

> colors should work with inverse terminals

Yep, works for me :)

> I've chopped out everything except old revisions from the lookup table (why redundantly define stuff when the algorithm in the new revision parser can handle it)

I agree that would make sense (I'd always wondered why the same info was encoded in two separate places), but that change doesn't seem to be present in the current PR?

P.S. I've just noticed that you even show different corners in the ascii-art for the square-edged and rounded-corners boards. Nice touch ;)

lurch:
You might be able to squeeze "Zero v1.2" and "Zero v1.3" into the board ascii-art? e.g.
```
.-------------------------.
| oooooooooooooooooooo P1 |
| 1ooooooooooooooooooo   |C
---+       +----+ PiZero |S
 sd|       |SoC |  v1.3  |I
---+|hdmi| +----+ usb pwr |
`---|    |--------| |-| |-'
```

gpiozero/pins/data.py
waveform80: Ooops

gpiozero/pins/data.py
waveform80: Yeah, they should - most of this is a bit rushed :)

gpiozero/pins/data.py
waveform80: Ah, knew I'd forgotten something!

waveform80:
> I wonder if the PiZero 1.3 should have its 'CSI' label in lower-case, given that it's a smaller connector than on the other Pis?

Yeah, good idea

> Could you document how to switch off colour using format-specs?

Will do

> I agree that would make sense (I'd always wondered why the same info was encoded in two separate places), but that change doesn't seem to be present in the current PR?

Huh ... that's weird. Where the hell did I make that change then? I remember making it *somewhere* ... reminds me of that post on bash.org "I've lost a machine ... I can ping it, but I can't *find* it"

> P.S. I've just noticed that you even show different corners in the ascii-art for the square-edged and rounded-corners boards. Nice touch ;)

As I was telling Ben, I'm an old hand at ASCII art :) I'm amazed no-one's made something like TheDraw for modern systems (it'd be great to use the Unicode box drawing chars, with some auto-mapping to convert them to similar ASCII chars on Windows' desperately backwards console).

> You might be able to squeeze "Zero v1.2" and "Zero v1.3" into the board ascii-art? e.g.

Oooh - that's nice!

waveform80:
Okay, that's all the bugs fixed and enough tests to get codecov to shut up. Good enough to merge?

gpiozero/pins/data.py
lurch: It's just occurred to me that both _format_row and _format_col could possibly share a _format_pin function? Not important though.

waveform80: ah, good point - that was a quickie copy/paste block

lurch:
LGTM! I don't think there's much left to nuke... ;-)

(unless you wanted to add a unit-test for the '0002' and '900092' Pis having only a P1 header, and the '0004' Pi having both a P1 and P5 header?)

Actually, perhaps the new 'pprint' tests should use 'a21041' (instead of '900092') so that they're more consistent with e.g. `test_physical_pins`?

gpiozero/pins/data.py
```diff
@@ -242,37 +351,113 @@
...
+    0x9:      ('A',    '2.0', '2013Q1', 'BCM2835', 'Qisda',     256,  'SD',      1,  0,  False, False, 1,  1,  {'P1': REV2_P1, 'P5': REV2_P5}, A_BOARD,      ),
+    0xd:      ('B',    '2.0', '2012Q4', 'BCM2835', 'Egoman',    512,  'SD',      2,  1,  False, False, 1,  1,  {'P1': REV2_P1, 'P5': REV2_P5}, REV2_BOARD,   ),
+    0xe:      ('B',    '2.0', '2012Q4', 'BCM2835', 'Sony',      512,  'SD',      2,  1,  False, False, 1,  1,  {'P1': REV2_P1, 'P5': REV2_P5}, REV2_BOARD,   ),
+    0xf:      ('B',    '2.0', '2012Q4', 'BCM2835', 'Egoman',    512,  'SD',      2,  1,  False, False, 1,  1,  {'P1': REV2_P1, 'P5': REV2_P5}, REV2_BOARD,   ),
+    0x10:     ('B+',   '1.2', '2014Q3', 'BCM2835', 'Sony',      512,  'MicroSD', 4,  1,  False, False, 1,  1,  {'P1': PLUS_P1},                BPLUS_BOARD,  ),
+    0x11:     ('CM',   '1.2', '2014Q2', 'BCM2835', 'Sony',      512,  'eMMC',    1,  0,  False, False, 2,  2,  {'SODIMM': CM_SODIMM},          CM_BOARD,     ),
+    0x12:     ('A+',   '1.2', '2014Q4', 'BCM2835', 'Sony',      256,  'MicroSD', 1,  0,  False, False, 1,  1,  {'P1': PLUS_P1},                APLUS_BOARD,  ),
```
lurch: I think this (0x12) should actually be A+ rev 1.1? (which would agree with both eLinux, and the actual A+ boards that I have)

waveform80:
> (unless you wanted to add a unit-test for the '0002' and '900092' Pis having only a P1 header, and the '0004' Pi having both a P1 and P5 header?)

Yeah, why not.

> Actually, perhaps the new 'pprint' tests should use 'a21041' (instead of '900092') so that they're more consistent with e.g. test_physical_pins?

Well, they're all the same headers these days - nice to have a bit of variety (I mainly used 900092 as I could remember it and all its details without looking stuff up (my tests tend to be based on whatever's freshest in my head, in this case the Zero's having just fiddled with the board art for them :)

gpiozero/pins/data.py
```diff
@@ -242,37 +351,113 @@
...
+    0x13:     ('B+',   '1.2', '2015Q1', 'BCM2835', 'Egoman',    512,  'MicroSD', 4,  1,  False, False, 1,  1,  {'P1': PLUS_P1},                BPLUS_BOARD,  ),
+    0x14:     ('CM',   '1.1', '2014Q2', 'BCM2835', 'Embest',    512,  'eMMC',    1,  0,  False, False, 2,  2,  {'SODIMM': CM_SODIMM},          CM_BOARD,     ),
+    0x15:     ('A+',   '1.1', '2014Q4', 'BCM2835', 'Sony',      256,  'MicroSD', 1,  0,  False, False, 1,  1,  {'P1': PLUS_P1},                APLUS_BOARD,  ),
```
lurch: And according to eLinux, the 0x15 A+ is made by Embest, not Sony.

(and apparently the 0xf Model B is made by Qisda rather than Egoman, if that's worth fixing?)

gpiozero/pins/data.py
```diff
@@ -422,13 +782,185 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
...
+                    'A': 'SD',
+                    'B': 'SD',
+                    'CM': 'eMMC',
+                    }.get(model, 'MicroSD')
+                usb = {
+                    'A':    1,
+                    'A+':   1,
+                    'Zero': 1,
+                    'B':    2,
+                    'CM':   0,
```
lurch: This should be `'CM':   1,`

lurch:
So, just out of curiosity (and because of insomnia), I wrote a little function to compare the `pi_info()` results for two different Pi revisions:
```python
from gpiozero import pi_info

def rev_diff(rev1, rev2):
    info1 = pi_info(rev1)._asdict()
    info2 = pi_info(rev2)._asdict()
    for key in info1.keys():
        if key not in ('revision',):
            if info1[key] != info2[key]:
                if key in ('board', 'headers'):
                    print('{0} is different'.format(key))
                else:
                    print('{0}: {1} != {2}'.format(key, info1[key], info2[key]))
```
and then I tried to see how close I could get a (fake) new-style revision to match up with the old-style revision codes. Which is how I spotted the CM USB error :)  Assuming the errors I've pointed out earlier (highlighted below with `# error`) get fixed, these are the closest matches I came up with:
(obviously some of the release dates don't match, and the new-style revision scheme doesn't have an encoding for Qisda, so I chose Egoman instead)
```python
>>> rev_diff(0x2,0x810011)
>>> rev_diff(0x3,0x810011)
released: 2012Q3 != 2012Q1
>>> rev_diff(0x4,0x800012)
released: 2012Q3 != 2012Q4
>>> rev_diff(0x5,0x810012)
manufacturer: Qisda != Egoman
>>> rev_diff(0x6,0x810012)
>>> rev_diff(0x7,0x810002)
>>> rev_diff(0x8,0x800002)
>>> rev_diff(0x9,0x810002)
manufacturer: Qisda != Egoman
>>> rev_diff(0xd,0x910012)
>>> rev_diff(0xe,0x900012)
>>> rev_diff(0xf,0x910012) # should report manufacturer: Qisda != Egoman
>>> rev_diff(0x10,0x900032)
>>> rev_diff(0x11,0x900061)
pcb_revision: 1.2 != 1.1 # error
usb: 1 != 0 # error
>>> rev_diff(0x12,0x800021)
pcb_revision: 1.2 != 1.1 # error
>>> rev_diff(0x13,0x910032)
released: 2015Q1 != 2014Q3
>>> rev_diff(0x14,0x920061)
usb: 1 != 0 # error
>>> rev_diff(0x15,0x820021)
manufacturer: Sony != Embest # error
```

gpiozero/pins/data.py
```diff
@@ -422,13 +782,185 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
...
+                    'Zero': 0,
+                    }.get(model, csi)
+                headers = {
+                    'A':  {'P1': REV2_P1, 'P5': REV2_P5},
+                    'B':  {'P1': REV2_P1, 'P5': REV2_P5} if pcb_revision == '2.0' else {'P1': REV1_P1},
+                    'CM': {'SODIMM': CM_SODIMM},
+                    }.get(model, {'P1': PLUS_P1})
+                board = {
+                    'A':    A_BOARD,
+                    'B':    REV1_BOARD if pcb_revision == '1.0' else REV2_BOARD,
```
lurch: Perhaps this should be modified to better-match the 'headers' if-test just above? i.e.
```python
'B':    REV2_BOARD if pcb_revision == '2.0' else REV1_BOARD,
```

gpiozero/pins/data.py
waveform80: Yup, looks like some of those changes were made after I last scraped the table. I'll get it updated to match.

gpiozero/pins/data.py
waveform80: May as well while I'm at it.

gpiozero/pins/data.py
waveform80: Ah, good point - I was wondering the other day what on earth USB DM was and whether I'd done something wrong when copy'n'pasting the diagram. Nope - it's DM on the diagram too - so they've used N for NEGATIVE everywhere else but on the USB pins it's M for MINUS? (and therefore presumably P for PLUS instead of POSITIVE :)

gpiozero/pins/data.py
waveform80: Yeah, consistency is always nice and to my mind the "else" case should be the general one (i.e. we test for the exceptional rather than the general).

waveform80:
> So, just out of curiosity (and because of insomnia), I wrote a little function to compare the pi_info() results for two different Pi revisions:

Ah, nice idea. I'm not terribly bothered by release date differences - that's all to be expected and I assume many are guesses anyway. But I'll get the rest of the diffs fixed up.

gpiozero/pins/data.py
lurch: I dunno the reasoning, but it's also USB DM  / DP in the Model B schematics, so at least it's consistent ;-)  (and _N / _P for all the other differential signals)

gpiozero/pins/data.py
```diff
@@ -422,13 +782,185 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
...
+                    'Zero': 0,
+                    }.get(model, csi)
+                headers = {
+                    'A':  {'P1': REV2_P1, 'P5': REV2_P5},
+                    'B':  {'P1': REV2_P1} if pcb_revision == '1.0' else {'P1': REV1_P1, 'P5': REV2_P5},
+                    'CM': {'SODIMM': CM_SODIMM},
+                    }.get(model, {'P1': PLUS_P1})
+                board = {
+                    'A':    A_BOARD,
+                    'B':    REV2_BOARD if pcb_revision == '1.0' else REV2_BOARD,
```
lurch: Ooops! Should be `REV2_BOARD if pcb_revision == '1.0'`

gpiozero/pins/data.py
```diff
@@ -422,13 +782,185 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
...
+            # MSB -----------------------> LSB
+            # uuuuuuuuFMMMCCCCPPPPTTTTTTTTRRRR
+            #
+            # uuuuuuuu - Unused
+            # F        - New flag (1=valid new-style revision, 0=old-style)
+            # MMM      - Memory size (0=256, 1=512, 2=1024)
+            # CCCC     - Manufacturer (0=Sony, 1=Egoman, 2=Embest)
+            # PPPP     - Processor (0=2835, 1=2836, 2=2837)
+            # TTTTTTTT - Type (0=A, 1=B, 2=A+, 3=B+, 4=2B, 5=Alpha (??), 6=CM, 8=3B, 9=Zero)
+            # RRRR     - Revision (0, 1, or 2)
```
lurch: As mentioned earlier, this now goes "up to 3" for the newest PiZero, so perhaps change the comment to ` Revision (0, 1, 2, etc.)` ?

gpiozero/pins/data.py
```diff
@@ -422,13 +782,185 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
...
+                csi = {
+                    'Zero': 0 if pcb_revision == '1.2' else 1,
+                    'CM':   2,
+                    }.get(model, 1)
+                dsi = {
+                    'Zero': 0,
+                    }.get(model, csi)
+                headers = {
+                    'A':  {'P1': REV2_P1, 'P5': REV2_P5},
+                    'B':  {'P1': REV2_P1} if pcb_revision == '1.0' else {'P1': REV1_P1, 'P5': REV2_P5},
```
lurch: ...and this should be:
```
'B':  {'P1': REV1_P1} if pcb_revision == '1.0' else {'P1': REV2_P1, 'P5': REV2_P5},
```
Apologies for not spotting this last time round!

I'm beginning to wonder if it'd be worth having a board -> headers-dict lookup table?

waveform80: Damn, quite possibly - unfortunately I haven't got time to fiddle much more at the moment - got a ton of stuff to do before I head off to PyconUK on Thurs. I'll get this committed in a mo, then you can PR fixes for everything I've broken ;)

gpiozero/pins/data.py
```diff
@@ -422,13 +782,185 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
...
+                csi = {
+                    'Zero': 0 if pcb_revision == '1.2' else 1,
+                    'CM':   2,
+                    }.get(model, 1)
+                dsi = {
+                    'Zero': 0,
+                    }.get(model, csi)
+                headers = {
+                    'A':  {'P1': REV2_P1, 'P5': REV2_P5},
+                    'B':  {'P1': REV1_P1} if pcb_revision == '1.0' else {'P1': REV1_P1, 'P5': REV2_P5},
```
lurch: The bit at the very end of this line should read `{'P1': REV2_P1, 'P5': REV2_P5},`  (you've got REV1_P1 instead of REV2_P1)

waveform80: Good point - I shouldn't rush this stuff in the middle of other things!

lurch:
Hurrah, all seems fine now :-)
```python
>>> rev_diff(0x2,0x810011)
>>> rev_diff(0x3,0x810011)
released: 2012Q3 != 2012Q1
>>> rev_diff(0x4,0x800012)
released: 2012Q3 != 2012Q4
>>> rev_diff(0x5,0x810012)
manufacturer: Qisda != Egoman
>>> rev_diff(0x6,0x810012)
>>> rev_diff(0x7,0x810002)
>>> rev_diff(0x8,0x800002)
>>> rev_diff(0x9,0x810002)
manufacturer: Qisda != Egoman
>>> rev_diff(0xd,0x910012)
>>> rev_diff(0xe,0x900012)
>>> rev_diff(0xf,0x910012)
manufacturer: Qisda != Egoman
>>> rev_diff(0x10,0x900032)
>>> rev_diff(0x11,0x900061)
>>> rev_diff(0x12,0x800021)
>>> rev_diff(0x13,0x910032)
released: 2015Q1 != 2014Q3
>>> rev_diff(0x14,0x920061)
>>> rev_diff(0x15,0x820021)
```

waveform80:
Yay!

Issue #397, opened 2016-08-03
Classes which support blink() should have a blinking property


martinohanlon:
e.g. when an LED is blinking you cant tell if its blinking through a public property.

As a workaround you can check whether _blink_thread != None.

lurch:
Nice idea :-)

Issue #228, opened 2016-03-03
RGBLED.cycle()


bennuttall:
Suggestion:

We add a `cycle` method to `RGBLED` which cycles through the colours. Parameters similar to `blink` (maybe `cycle_time`, `n`, `background`, `start_color`)

lurch:
Been having a think about this, and I guess in a way it's slightly analogous to #149 
So what about a method signature like:
`def cycle(on_time=1, fade_time=0, colors=((1,0,0), (0,1,0), (0,0,1)), n=None, background=True)`
?

Calling `rgbled.cycle()` would then display a repeated sequence of "red on for 1 second, green on for one second, blue on for one second" and `rgbled.cycle(0, 2)` would display a repeated sequence of "2 seconds smoothly fading (re-using the `lerp` code from [here](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L644) ) from red to green, 2 seconds smoothly fading from green to blue, 2 seconds smoothly fading from blue to red".
And you could also do `rgbled.cycle(0, 1, ((1,0,0), (1,1,0), (0,1,0), (0,1,1), (0,0,1), (1,0,1)) )` to fade through the full HSV range.
(So perhaps that'd be a better default-value for `colors`? Meaninig that `rgbled.cycle()` with no args would then 'flick' through "red, yellow, green, blue, cyan, magenta").

And given that `cycle` with no-args is kinda-similar to `blink` with no-args, maybe we could also add a `fadecycle` (or `smoothcycle`) which simply calls `cycle` with different default-args; similar to how `pulse` simply calls `blink` with different default-args?


Alternatively, if we wanted a simpler interface (perhaps I'm over-complicating things again) we could go with Ben's suggestion of:
`def cycle(cycle_time=1, start_hue=0, n=None, background=True)`
and then always smoothly cycle through the full range of colours using something like https://en.wikipedia.org/wiki/HSL_and_HSV#/media/File:HSV-RGB-comparison.svg
[...]

Issue #149, opened 2016-01-18
blink sequence method


shmick:
I was using the led blink method and was wondering if it could be possible to add a blink sequence method, where multiple on / off times could be defined?

lurch:
https://github.com/lurch/python-gpiozero/commit/ae366f3043eb452614e51c3f1916c41e3e40d72f is one possible way of doing it (which I won't be submitting as a PR, because it's only a rough first stab).

You'd create your LED object as normal, and then the `sequence` expected by `blink_sequence` is a list of tuples of `(on_time, off_time)`, e.g. you could do something like:
```python
from gpiozero import LED
from signal import pause

led = LED(17)
led.blink_sequence([(0.2, 0.3), (1.2, 0.5), (0.4, 1.5)])
pause()
```

And then the standard `led.blink()` is identical to `led.blink_sequence([(1, 1)])`

bennuttall:
This might be a case of API boat, but how about passing in a `sequence` arg to `blink()` that takes a list/tuple of numbers, on/off in sequence, then repeated:

```python
led = LED(17)

sequence = (0.2, 0.3, 1.2, 0.5, 0.4, 1.5)  # or list of 2-tuples like lurch suggested
led.blink(sequence=sequence)
```

Though I guess that conflicts with `on_time` and `off_time` if you specify those as well. They could be ignored in that case. I'm not sure really.

I think most sequences would be calculated anyway (i.e. `on_time` doubles as `off_time` halves) so that could be done more easily in a loop, though that would have to be in the foreground.

bennuttall:
I'm leaning on the side of voting that this is API bloat. There are already four (optional) parameters for `blink`. Those four parameters provide the flexibility to do clever things with `blink`. Besides `on_time` and `off_time` there's:

- `n` => Specify the number of times to cycle (e.g. `led.blink(n=10)` would blink 10 times, then stop)
- `background` => Specify whether the blink loop should take place in the background or the foreground

Used in combination, this can allow a repetitive sequence:

```python
while True:
    led.blink(on_time=2, off_time=1, n=10, background=False)
    sleep(1)  # because it doesn't do the final off_time
    led.blink(on_time=1, off_time=2, n=10, background=False)
    sleep(2)  # because it doesn't do the final off_time
```

This would go on for 2 seconds, off for 1, a total of 10 times, then swap to on 1, off 2 for 10 times, then repeat.

[...]

PR #449, opened 2016-09-19
Disable source when directly controlling DigitalOutputDevice, close #200

### LINKED ISSUE #200: Can't directly-control an OutputDevice whose source has been set
While thinking about #189 I came up with the following example:

``` python
from gpiozero import LED
from signal import pause
red = LED(21)
green = LED(20)
red.source = green.values
red.blink()
pause()
```

You _might_ expect the code above to blink the red LED, but it doesn't.
I guess there's two ways of looking at this: a) this is 'undefined behaviour' and you shouldn't try directly-setting an LED that has its source property set (in which case perhaps a warning or exception should be thrown) or b) directly-setting an LED's value should 'nullify' the source property.
I guess perhaps b) follows the 'principle of least surprise' (maybe?) and 'cancelling' the `_source_thread` when a value is directly set, would be very similar to the way that the `_blink_thread` is currently cancelled when a value is directly set. But I'm not sure how/if this would work in terms of Mixins.

Also, should (nonsensical) code like that below throw some kind of exception?

``` python
from gpiozero import LED
red = LED(21)
green = LED(20)
red.source = green.values
green.source = red.values
```

(attempting to do either `red.close()` or `green.close()` after this throws a `AttributeError: 'NoneType' object has no attribute 'state'`)

bennuttall:
A fix for #200

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -185,8 +185,7 @@ class DigitalOutputDevice(OutputDevice):
             self._blink_thread = None
 
     def _stop_source(self):
-        if self.source is not None:
-            self.source = None
+        self.source = None
 
     def _stop_blink(self):
         if self._controller:
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -142,14 +142,17 @@ class DigitalOutputDevice(OutputDevice):
 
     def close(self):
         self._stop_blink()
+        self._stop_source()
         super(DigitalOutputDevice, self).close()
 
     def on(self):
         self._stop_blink()
+        self._stop_source()
         self._write(True)
 
     def off(self):
         self._stop_blink()
+        self._stop_source()
         self._write(False)
 
     def blink(self, on_time=1, off_time=1, n=None, background=True):
@@ -172,6 +175,7 @@ class DigitalOutputDevice(OutputDevice):
             this method never returning).
         """
         self._stop_blink()
+        self._stop_source()
         self._blink_thread = GPIOThread(
             target=self._blink_device, args=(on_time, off_time, n)
         )
@@ -179,6 +183,10 @@ class DigitalOutputDevice(OutputDevice):
         if not background:
             self._blink_thread.join()
             self._blink_thread = None
+
+    def _stop_source(self):
+        if self.source is not None:
+            self.source = None
 
     def _stop_blink(self):
         if self._controller:
@@ -1207,4 +1215,3 @@ class AngularServo(Servo):
                 self._value_range *
                 ((value - self._min_angle) / self._angular_range) +
                 self._min_value)
-


gpiozero/output_devices.py
```diff
@@ -180,6 +184,10 @@ def blink(self, on_time=1, off_time=1, n=None, background=True):
             self._blink_thread.join()
             self._blink_thread = None
 
+    def _stop_source(self):
+        if self.source is not None:
```
bennuttall: I guess this if statement is redundant. If it's None already, setting it to None will do no harm.

lurch:
Given that _stop_source is used everywhere that _stop_blink is used, would it have made sense to just include it inside _stop_blink?  (and _stop_source seems missing from @value.setter ? )

Perhaps it's just me, but "stop" feels like the wrong verb to use with "source" :-/  "disconnect source" maybe?

Why is this only on DigitalOutputDevice - should it be part of SourceMixin?

Just food for thought... ;-)

waveform80:
> Given that _stop_source is used everywhere that _stop_blink is used, would it have made sense to just include it inside _stop_blink? (and _stop_source seems missing from @value.setter ? )
(EDIT: or maybe _stop_blink and _stop_source could be bundled together into a _stop_threads function?)

Ah, we tried that but it doesn't work: specifically, setting value must stop blinking but must *not* disconnect the source. However, source explicitly sets value while connected, so there's one case where _stop_blink must be called, but _stop_source must not.

> Perhaps it's just me, but "stop" feels like the wrong verb to use with "source" :-/ Maybe "disconnect source" instead?

Hmm, consistency versus semantics. Not sure which I prefer :)

> Why is this only on DigitalOutputDevice - should it be part of SourceMixin?

Looking at it, it should definitely be moved up to OutputDevice (where SourceMixin first enters that hierarchy) ... and arguably _stop_source (or whatever we call it) should be part of SourceMixin, but given Ben's comment that really it ought to be a single `self.source = None` line anyway I'm not convinced we even need a method for it - just a load of `self.source = None` lines scattered about the place. Hmm, haven't checked boards.py either - plenty of output devices there.

lurch:
> setting value must stop blinking but must not disconnect the source

Are you _sure_ that's the behaviour we want? Observe:
```python
>>> import gpiozero.devices
>>> from gpiozero.pins.mock import MockPin
>>> gpiozero.devices.pin_factory = MockPin
>>> from gpiozero import Button, LED
>>> b = Button(5)
>>> l = LED(6)
>>> l.source = b.values
>>> b.value
False
>>> l.value
False
>>> l.value = True
>>> l.value
False
```

A user might expect that setting `l.value` would work, and that it would disconnect the source.
Perhaps the SourceThread needs to be modified to use `_write()` rather than `value` ?

bennuttall:
> Looking at it, it should definitely be moved up to OutputDevice (where SourceMixin first enters that hierarchy)

That makes sense.

>  ... and arguably _stop_source (or whatever we call it) should be part of SourceMixin, but given Ben's comment that really it ought to be a single self.source = None line anyway I'm not convinced we even need a method for it - just a load of self.source = None lines scattered about the place. Hmm, haven't checked boards.py either - plenty of output devices there.

Well, it started out as a 3 or 4 line method doing 2 things (stopping source and stopping blink), so a method made sense. Then we removed the second thing and instead repeated it in on/off/toggle/blink and even removed the if statement so it became unnecessary. I don't think it matters either way, unless there's a chance we might want to bung something else into the method in future, then at least it only needs changing in one place.

bennuttall:
> Are you sure that's the behaviour we want?

Ok I think we have two different things here:

1. The user explicitly setting the value
2. The value being set internally by a source thread or blink thread, or on/off/toggle methods.

We can't disable source just because the value has been set, in case it was set by the source thread.

We should disable the blink or source thread when `on()` is called, but I think we're saying we should also disable it when `.value` is set by the user.

So do we differentiate between the two cases:

```python
led.on()  # should stop blink and source threads?
```

```python
led.value = 1  # should not stop blink and source threads?
```

?

If not, I guess we would need to create an internal `._set_value()` method which the internal methods used, and the `.value` being set by the user would terminate the source.

*Note: I just realised I think `_terminate_blink()` and `_terminate_source()` might be the right method names...*

lurch:
Thinking about this issue, and looking at https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L191 (which already uses `self._write(new_value)` instead of `self.value = new_value`), I'm beginning to wonder if the "right approach" might be to swap things around, so that blink is implementable as source/values, and then calling `led.blink()` would just do `self.source = blink_values(...)` or similar.
It'd then mean there'd just be a single `source_thread`, and not a separate thread _just_ for blinking.

?

If I get time on Thursday evening I'll see if I can knock up a prototype.  (*Hmmm, perhaps there'd be problems with the sub-blinkability on CompositeDevices like SnowPi?*)

bennuttall:
Good point! That could work.

...except that would collide with #397 where we suggested digital output devices could have an `.is_blinking` property... though that could be set separately ... maybe?

lurch:
`.is_driven_by_source` ? ;-)

waveform80:
> Thinking about this issue, and looking at https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L191 (which already uses self._write(new_value) instead of self.value = new_value), I'm beginning to wonder if the "right approach" might be to swap things around, so that blink is implementable as source/values, and then calling led.blink() would just do self.source = blink_values(...) or similar.
> It'd then mean there'd just be a single source_thread, and not a separate thread just for blinking.

Very interesting thought. That would solve the issue with setting source (because it'd automatically override whatever blink set as the source). It'd also keep all the other methods pretty much as they are (just instead of calling _stop_blink, which would disappear, they'd simply set source = None).

The one thing that springs to mind is that in order to guarantee blinks at particular frequencies we'd have to implicitly set source_delay to some reasonable value. Still, I don't have a particular issue with that:

1. it's a bit strange to mix paradigms like source and on/off or blink, hence I won't have a great deal of sympathy for people that complain it "silently" adjusts source_delay to do what they asked it to do
2. the default source_delay of 1/100th of a second is probably sufficient for the vast majority of cases
3. so, this'll only affect people that both adjust source_delay *and* then call blink (I suspect that's a pretty slim population)

Yeah, I like this idea.

> Good point! That could work.
>
> ...except that would collide with #397 where we suggested digital output devices could have an .is_blinking property... though that could be set separately ... maybe?

I think what `is_blinking` is really trying to solve is more or less what @lurch alludes to: "is the LED currently being controlled by some background thread?". In which case we could either stick with the name `is_blinking` (which probably covers the majority of use-cases) or come up with something a little more descriptive. Either way, the implementation could simply be "is the source thread alive?"

lurch:
> so, this'll only affect people that both adjust source_delay and then call blink (I suspect that's a pretty slim population)

Wouldn't the blink function set both source_delay _and_ source?

I'll have to remind myself of the difference between http://gpiozero.readthedocs.io/en/latest/api_generic.html#gpiozero.SourceMixin.source_delay and http://gpiozero.readthedocs.io/en/latest/api_tools.html#gpiozero.tools.post_delayed :)

waveform80:
> Wouldn't the blink function set both source_delay and source?

Yes, precisely. Sorry, I wasn't clear: I meant people might be surprised that it adjusts source_delay to an appropriate value (but that I'm fine with that :).

bennuttall:
What would source/values based implementation of `blink()` look like?

It *could* manipulate `source_delay` but then would it need to set it back to what it was before, to make sure there are no surprises? And would that be possible? Alternatively, it could pass in a series of 1s and 0s to ensure the blink would work correctly at the current source delay ... but that could be tricky. And what about pulse? Oh god...

I think we might be going down a rabbit hole now (if not before)...

waveform80:
Actually it's pretty easy - blink already constructs an iterable of values but it assumes differing pauses between on and off points (on_time and off_time) while source always has the same pause between (source_delay). So that'd need adjusting slightly, but it shouldn't be too hard. As for pulse: much the same. Again, it constructs a list of values suitable for source - just need to get the timing right. As for number of repeats, itertools.cycle will deal with infinites and a simple variant can deal with finite repeats (you could just multiply the source array, but for long repeats that'll be wasteful of memory).

waveform80:
The only tricky bit is the thing @lurch mentioned about composite devices controlling their subordinates ... I haven't thought about that yet (sorry, only so much brain-bandwidth I can devote to side discussions while refactoring pins!)

lurch:
Exactly - the source_delay would just need to be some common factor of both on_time and off_time.
And unifying `source`, `blink` and `pulse` would also make #149 and #228 really simple.
(Unifying pulse, blink, blink_sequence and cycle is something I experimented with on a private branch months ago, but never got around to polishing and PRing. IIRC the solution I came up with at the time involved adding another mixin)

waveform80:
> Exactly - the source_delay would just need to be some common factor of both on_time and off_time.

Yup - some variant on GCF (we should always pick the greatest common factor simply for efficiency - the bigger the source_delay the more time the background thread spends sleeping and not nicking time from everyone else).

lurch:
> we should always pick the greatest common factor simply for efficiency

That's exactly what I was thinking / planning :)

normanr:
So I'd like to be able to fade between two values with a PWMLED.  From 0 to 1, or 1 to 0, which you have to do manually with the current API.

I ended up making a function that works similar to _blink_device, but instead of _write(value), I set source to a generator which yields the value and waits for delay.

eg:
```
def fader(from_value, to_value, fade_time=1, fps=25):
    if fade_time > 0:
        value_size = to_value - from_value
        for i in range(int(fps * fade_time)):
            yield i * (1 / fps) / fade_time * value_size + from_value
            threading.current_thread().stopping.wait(1 / fps)
    yield to_value
```

usage:
```
    led.source_delay = 0
    led.source = fader(0, 1)
```

lurch:
Hmmm, kinda like a https://gpiozero.readthedocs.io/en/stable/api_tools.html#gpiozero.tools.ramping_values which stops at one end of the ramp?

Perhaps it'd be worth adding a `fade_to(value, fade_time=1, background=True)` method to `PWMLED` which would fade from the LED's current value to `value` over `fade_time` seconds ? (a bit like a much-simplified version of `pulse`)

bennuttall:
Can't you just do this by iterating over the result of a list comprehension?

```python
led.source_delay = 0.01
led.source = [i/100 for i in range(100)] # <- up or down -> reversed(i/100 for i in range(100))
```

lurch:
IMHO
```python
led.fade_to(1)
```
looks a lot simpler than
```python
led.source_delay = 0.01
led.source = [i/100 for i in range(100)]
```
and isn't the goal of GpioZero to be easy-to-use? ;-)

bennuttall:
I wasn't really arguing against that (except for more api bloat, it's a fine suggestion), more against the idea that it's really complex with the current API.

@normanr what do you think to the idea of using a list comprehension? Do you think we should add a `fade_to` method?

Tbh `fade_to` would be quite useful in RGBLED.

normanr:
In the context of this bug I was thinking that blink, pulse, fade_to could/should all be implemented as event source generator functions.  This way the blink/pulse threads go away and just need to set events to None in value setter (for example).  The public APIs don't need to change at all.

My example code above should actually be split into value generation and timing as two functions as their own.  This way all features are compositions of event sources, eg:

blink: delay_after((on, off))
pulse: delay_afyer(chain(fade in, fade out))
(both can have cycle added for repetition, before the delay part)

The actual logic (of blink or pulse) doesn't change very much just the final loop becomes a yield/wait generator function instead of the set value loop.

normanr:
Also the fade/pulse API should be flexible enough that it can be used for RGB as well as single color, by using different interpolate functions.

Eg fade only generates 0 to 1, but then the interpolater can 'scale' that between two different colors.  Or if you want a logarithmic fade instead of a linear one, you'd add a logarithmic scaler.

normanr:
btw, it doesn't change the number of threads: the blink_thread would just get replaced with the source thread.

bennuttall:
Yes but the source thread already exists anyway.

PR #453, opened 2016-09-19
Use Python syntax highlighting

thijstriemstra:


lurch:
Just a quick heads-up that `README.rst` is used by Github, ReadTheDocs, Sphinx *and* PyPI. The last time @waveform80 tried changing it (#295 and #403 ) one or more of those parts didn't work properly so he had to revert things. I dunno python / rst documentation well enough to know if this change is 'okay' in all those places ;-) @waveform80 ?

But if it is 'okay', would it be worth modifying _all_ the inline-code snippets in the docs in this way, and not just the ones in the README?

thijstriemstra:
I've changed it to the docutils `code` [directive](http://docutils.sourceforge.net/docs/ref/rst/directives.html#code) (available since docutils 0.9) that should also work on pypi afaik.

thijstriemstra:
For Sphinx you can use the [highlight_language](http://www.sphinx-doc.org/en/stable/config.html#confval-highlight_language) project setting.

thijstriemstra:
Someone should also change to the https readthedocs url in the Github project description.

docs/conf.py
```diff
@@ -52,12 +53,13 @@ def __getattr__(cls, name):
 #source_encoding = 'utf-8-sig'
 master_doc = 'index'
 project = _setup.__project__.title()
-copyright = '2015 %s' % _setup.__author__
+copyright = '2015-{0} {1}'.format(datetime.now().year, _setup.__author__)
```
lurch: The rest of this file uses %-style string interpolation, so IMHO this should be switched back to that rather than changing it to use `.format`.

lurch:
I'm not able to edit the URL in the Github project description, but I did just change a couple of URLs in the README to use https :-)

waveform80:
> Just a quick heads-up that README.rst is used by Github, ReadTheDocs, Sphinx and PyPI. The last time @waveform80 tried changing it (#295 and #403 ) one or more of those parts didn't work properly so he had to revert things. I dunno python / rst documentation well enough to know if this change is 'okay' in all those places ;-) @waveform80 ?

I'd be quite surprised if all of those didn't support `.. code` - it's part of the base reST set of directives (unlike the conditional directive I was attempting to use which is a loadable extension). Then again, I have found base directives that GitHub's renderer in particular didn't support (I'm not sure they're using docutils). Oh well, let's try it and see!

bennuttall:
Note that this PR incorporates multiple changes - some certainly look reasonable and I'd be happy to merge, others I'd leave to @waveform80.

bennuttall:
> I'm not able to edit the URL in the Github project description

Done

PR #454, opened 2016-09-20
Remove mailto: from the front of email address

lurch:
It gets converted into a mailto link automatically


lurch:
I've submitted this as a PR instead of pushing it directly to master, in case this change isn't desired and/or it doesn't work in all the places that README.rst ends up getting used.

PR #458, opened 2016-09-20
Refactor GPIOQueue.fill()

lurch:
squash the two while loops down into one.

The unit-tests still all pass, and IMHO this is a little bit neater. But @waveform80 ought to double-check that I've not done anything stupid here ;-)

### DIFF
--- a/gpiozero/mixins.py
+++ b/gpiozero/mixins.py
@@ -493,15 +493,11 @@ class GPIOQueue(GPIOThread):
 
     def fill(self):
         try:
-            while (not self.stopping.wait(self.sample_wait) and
-                    len(self.queue) < self.queue.maxlen):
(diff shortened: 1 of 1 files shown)


waveform80:
Looks good to me; my only reason for using two loops originally is that filling the queue is a one-off operation so this permitted the body of the second loop (which will form the vast majority of the runtime) to be a bit simpler. But that's almost certainly premature optimization. Might be worth seeing just how long it takes to check an Event to see whether it's set - it *might* be an expensive operation (if it involves thread locks). But if not, this looks good to merge.

waveform80:
Yup, it's fast: no threading lock for checking whether it's set, and timings are about 1.3µs-1.5µs on a Pi3 under both Py2 and Py3.

PR #460, opened 2016-09-21
Add a new alternating_values SourceTool

lurch:
Another way of implementing this would be:
```python
def alternating_values(initial_value=False):
    return cycle([initial_value == True, initial_value == False])
```
but I think the version in this PR is a bit more readable ;-)

### DIFF
--- a/docs/api_tools.rst
+++ b/docs/api_tools.rst
@@ -69,6 +69,8 @@ Combining sources
 Artificial sources
 ==================
 
+.. autofunction:: alternating_values
+
 .. autofunction:: cos_values
 
 .. autofunction:: random_values
--- a/gpiozero/tools.py
+++ b/gpiozero/tools.py
@@ -576,3 +576,26 @@ def cos_values(period=360):
     angles = (2 * pi * i / period for i in range(period))
     for a in cycle(angles):
         yield cos(a)
+
+
+def alternating_values(initial_value=False):
+    """
+    Provides an infinite source of values alternating between ``True`` and
+    ``False``, starting wth *initial_value* (which defaults to ``False``). For
+    example, to produce a flashing LED::
+
+        from gpiozero import LED
+        from gpiozero.tools import alternating_values
+        from signal import pause
+
+        red = LED(2)
+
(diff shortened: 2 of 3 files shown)


bennuttall:
That's cool. I don't know which I consider neater. Both nice :)

bennuttall:
I also suggested to Dave adding a booleanised `random_values` function i.e. randomly on or off for a non-pwm device. But that's easy enough to achieve by processing `random_values`.

lurch:
That's exactly what http://gpiozero.readthedocs.io/en/latest/api_tools.html#gpiozero.tools.booleanized is for ;-)

bennuttall:
That's what I meant!

lurch:
...which means a very inefficient **third** way of implementing this PR would be:
```python
def alternating_values(initial_value=False):
    it = booleanized(cos_values(2), 0, 1)
    if not initial_value:
        it = negated(it)
    return it
```

;-)

waveform80:
Oh, that's nice. As I think I mentioned at PyConUK - source/values is by far my favourite bit of the library!

lurch:
I'm thinking of adding `ramping_values` next, which will be like a 'linear' version of `cos_values` (i.e. a triangle wave).

waveform80:
Ahh - I was wondering when someone (or myself) would get around to a triangle wave! I wonder if there should be aliases like square_values and triangle_values for these things?

lurch:
> I wonder if there should be aliases like square_values and triangle_values for these things?

I'm not sure... I guess `triangle_values` _could_ be aliased to `ramping_values`, but I think if we wanted a `square_values` function then it'd need a `period` parameter (which `ramping_values` will have).

Do you think I should make `ramping_values` output 0 -> 1 (like `alternating_values` effectively does), or -1 -> +1 (like `cos_values` does) ?  My inclination is the former. (Obviously either variant can be `scaled` to the other one afterwards)

waveform80:
Ah, good point about the `period` parameter. And yes, my gut says 0->1 as well. Scaling's easy enough to apply.

PR #462, opened 2016-09-22
Add release date for new BCM2837-based Pi 2B

bennuttall:
I think this is all that's needed to make `pi_info()` accurate


waveform80:
Er, shouldn't that be against 2B instead of B+?

bennuttall:
I have no idea how I managed that.

lurch:
Are you sure this is correct? The _current_ 2B (2015Q1) has a revision of 0xa01041 which decodes to a pcb_revision of 1.1

waveform80:
Yup, so the current one (rev 1.1) will have a release date of 2015Q1, the new (rev 1.2) will have 2016Q3

lurch:
Doh! I misread the 'if' logic as the other way around. Ignore me... ;-)

lurch:
Although... given that http://elinux.org/RPi_HardwareHistory mentions an a01040, should the 'if' be expanded to `if pcb_revision == '1.0' or pcb_revision == '1.1'` ?

waveform80:
Yeah, not a bad idea

lurch:
That's cool - I didn't know github let you edit other people's PRs. Hope you don't mind Ben!
IMHO we should get verification of the revision code on an actual 2837-based Pi 2B before this gets merged.

waveform80:
> That's cool - I didn't know github let you edit other people's PRs.

It's a recent thing. There's a new check-box when creating a PR labelled something like "allow maintainers to edit PR" which permits it (it's checked by default).

bennuttall:
It's `a22042`

lurch:
> It's a22042

Cheers. I've just updated http://elinux.org/RPi_HardwareHistory

Issue #464, opened 2016-09-23
Device = 1 on SPI not working


Grumpy-Mike:
using MCP3008(channel=0, device=1) or MCP3008(channel=0, device=0) results in the same pin being used for chip select, namely GPIO8

Even if this did work, how would you have two MCP3008s on the SPI bus one using GPIO8 and the other using GPIO7 as the chip address?

Python 2 - gpio zero versions 1.0.0 and 1.3.1

lurch:
Looking at what https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/spi.py#L26 is doing, I wonder if this might be a bug in spidev?

@waveform80 on a slight side-note (given that this is the first time I've properly looked at the SPI code), I think it would have made more sense for SPISoftwareInterface to descend from CompositeDevice (since an SPI interface is composed of multiple pins) rather than from OutputDevice?

@Grumpy-Mike Isn't [this](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus#Independent_slave_configuration) how it's supposed to work?

joan2937:
It appears to work correctly for me.
```py
#!/usr/bin/env python

import gpiozero

d1=gpiozero.MCP3008(channel=0, device=0)
d2=gpiozero.MCP3008(channel=0, device=1)

v1 = d1.value
v2 = d2.value

print(v1, v2)

d1.close()
d2.close()
```
CE0(8)/CE1(7) are used for d1/d2.

I'm using the version of gpiozero bundled with jessie (1.3.1).

Grumpy-Mike:
True that does. What does not work is
`#!/usr/bin/env python
from gpiozero import MCP3008

MCP3008(channel=0,device=1)

print"Twin A/D board test"
while True:
     a2d0 = MCP3008(0)  
     print" ",a2d0.raw_value`
Which makes this a problem with the documentation which suggests that it would work.

lurch:
I'm not sure I follow? By doing `MCP3008(channel=0,device=1)` you're creating an ADC on CS1, which instantly gets garbage-collected (because you're not storing the returned object anywhere). And then when you later do `a2d0 = MCP3008(0)` you're creating an ADC on CS0, because that's the default when no `device` is explicitly specified.

I wonder if this is more what you're trying to do?
```python
#!/usr/bin/env python
from gpiozero import MCP3008

[...]

Issue #459, opened 2016-09-21
Pins re-factor for remote SPI support


waveform80:
This is turning out to be a pretty hefty PR, so I figured before I get too deep into it I should note down some of the goals of this to make sure everyone's happy with them. So, here's the goals:
- At the moment our software SPI support (regardless of pin implementation) constructs pins and bit-bangs on them. This is all well and good for all the strictly local implementations, but it's no good for remote pins in pigpiod because the latency involved ruins the SPI bandwidth (to the point where stuff starts malfunctioning).
- Now, pigpiod has built-in SPI bit-banging support so instead of constructing a bunch of PiGPIOPin instances, we _should_ be constructing something that looks like SPISoftwareInterface (or SPIHardwareInterface) but which uses pigpiod's various SPI methods.
- This makes SPI pin implementation specific which means it needs to become part of the pin implementation itself.
- Something very similar to this is going to be required in the near future for I2C support (again, pigpiod supports this remotely).

Here's how I'm going about implementing this:
-  The pin implementations are already too complex (messing around with `__new__`) and I don't fancy adding yet more class methods. The sensible thing to do here is to make an actual Factory instance with methods for returning a pin, an SPI interface, and probably in future an I2C interface (maybe 1-wire too eventually).
- This means that pin_factory will change from being a Pin sub-class to being an actual Factory instance.
- I'll also be introducing an abstract base SPI class. SPISoftwareInterface and SPIHardwareInterface will descend from this, and instances of these will be returned by the factories associated with local pin implementations (i.e. everything except pigpiod).
- Each pin implementation will wind up implementing at least _two_ things: a Pin subclass and a Factory subclass, and optionally an SPI sub-class (as mentioned above, local pin factories will all use SPISoftwareInterface wrapping their respective pin subclasses, or SPIHardwareInterface).

Finally, this is what this means for end users:
- Because pin_factory is constructed up-front on library import, this implies that initialization of the pin factory will happen at import time rather than at time of first pin construction. This doesn't sound too important but it actually has a big impact on non-pi users as it means the NativeFactory tries to initialize straight away (and usually fails because neither /dev/gpiomem nor /dev/mem is accessible).
- This means on a PC you have to either setup the pigpio client properly or explicitly specify the mock factory to import gpiozero successfully; trying to import something that will only work on a Pi will fail at import time rather than the first time you try and use a pin.
[...]

PR #465, opened 2016-09-23
Slight refactoring of the SPI pin numbers

lurch:
While looking at the SPI code (due to #464 ) it really niggled me that the numbers `8` and `7` were hardcoded in so many places. So I decided to do a little bit of refactoring... (which should also make it easier to support SPI1 if/when drivers for that become available).

However in light of all the changes going on for #459 this should probably be treated more as a "suggestion PR" than a "merge PR".


waveform80:
Ah, this is a nice idea. Once #468 is reviewed and merged it'd be good to revisit this; I'm afraid I may have multiplied the places that it's needed!

Issue #378, opened 2016-06-14
Button bounce_time


lurch:
Given that the intention of GpioZero is to make components as easy to use as possible, and all the Button examples in the docs just do e.g. `button = Button(2)` I wonder if it might make sense for `bounce_time` in the Button init-method to default to something other than `None` ?

I suspect that up until now this hasn't really been an issue since most of the examples just use a Button to turn on and off an LED, in which case you won't notice the switch-bounce. But if you do something like printing a string to the screen when the button was pressed or released, a 'noisy' switch will trigger multiple press or release events for each _real_ press or release.

And something I just remembered while reading some old [blog-comments](http://raspi.tv/2013/how-to-use-interrupts-with-python-on-the-raspberry-pi-and-rpi-gpio-part-3) is that IIRC the way the `bounce_time` works in RPi.GPIO is that if you have a bounce-time of e.g. 100 milliseconds, and then register a falling-edge callback, pressing the button will trigger the callback as expected. But if you then hold the button down for a second (so that the bounce-time has elapsed) and then release the button, the switch-bounce on the release event can cause an additional falling-edge callback to be triggered. If this is still the case (I can't be sure, I'll have to check later) it would be great to see this handled 'properly' inside GpioZero so that only **real** button-presses trigger the button_pressed callback, no matter how long the switch is held down for before releasing it.

waveform80:
Interesting - I think I've encountered something like this in RPi.GPIO in the past, but I've no idea if it's still an issue (or if I was mistaking it for something else). Worth investigating though.

PR #468, opened 2016-09-25
Fix #459 - properly support remote SPI with pigpio

### LINKED ISSUE #459: Pins re-factor for remote SPI support

waveform80:
Sorry! Dave's messing around with the pin implementations again.
Hopefully the last time. The pin_factory is now really a factory object
which can be asked to produce individual pins or pin-based interfaces
like SPI (which can be supported properly via pigpio).


waveform80:
Wow ... passing on everything but pypy. I was honestly expecting more of a horror show here. And the pypy stuff just looks like the new reservation system whinging because of longer object lifespan on pypy, which I was half expecting. Let's see if I can get this passing quickly...

waveform80:
Yay! Passing on all variants. I'm not too bothered about the patch failure; given this is pins based stuff (which Travis can't directly test), and that it's expanding the implementation (so I'm adding a fair bit of stuff that can't be tested away from a Pi), it's not bad.

waveform80:
Okay, the patches are separated. This PR should now be just the remote SPI (and minor auxilliary stuff). I'll open a separate PR for the pin numbering stuff tomorrow.

waveform80:
Sorry, hadn't noticed this needed rebasing - done now, along with the dependent pin-numbering PR. Anybody got any time to look at reviewing this? I know it's horrid but we ought to get this in for 1.4

gpiozero/pins/__init__.py
```diff
@@ -6,39 +8,133 @@
...
+    SPIFixedSelect,
+    SPIFixedWordSize,
     )
 
 
-PINS_CLEANUP = []
-def _pins_shutdown():
-    for routine in PINS_CLEANUP:
-        routine()
+class Factory(object):
```
lurch: It looks weird, with this class not even providing an empty `__init__` method? Seems to imply that sub-classes won't need to provide an `__init__` either? Or is this a common Python idiom?

waveform80: Yup, that's pretty normal - if no `__init__` is needed, no need to specify one (the default `object.__init__` takes no parameters and does nothing so if that's acceptable there's no need to bother overriding it).

docs/api_pins.rst
```diff
@@ -11,68 +11,68 @@ are concerned with. However, some users may wish to take advantage of the
...
++=========+===============================================+===========================================+
+| rpigpio | :class:`gpiozero.pins.rpigpio.RPiGPIOFactory` | :class:`gpiozero.pins.rpigpio.RPiGPIOPin` |
++---------+-----------------------------------------------+-------------------------------------------+
+| rpio    | :class:`gpiozero.pins.rpio.RPIOFactory`       | :class:`gpiozero.pins.rpio.RPIOPin`       |
++---------+-----------------------------------------------+-------------------------------------------+
+| pigpio  | :class:`gpiozero.pins.pigpiod.PiGPIOFactory`  | :class:`gpiozero.pins.pigpiod.PiGPIOPin`  |
++---------+-----------------------------------------------+-------------------------------------------+
+| native  | :class:`gpiozero.pins.native.NativeFactory`   | :class:`gpiozero.pins.native.NativePin`   |
++---------+-----------------------------------------------+-------------------------------------------+
+
```
lurch: Do `MockFactory` and `MockPWMFactory` need to be documented anywhere? Or are these strictly "for internal testing only" and therefore shouldn't be publicly documented?
If only there was a developer documentation page.... ;-)

setup.py
```diff
@@ -68,12 +68,12 @@
...
-        'RPIOPin    = gpiozero.pins.rpio:RPIOPin',
-        'NativePin  = gpiozero.pins.native:NativePin',
-        'MockPin    = gpiozero.pins.mock:MockPin',
-        'MockPWMPin = gpiozero.pins.mock:MockPWMPin',
+        'pigpio  = gpiozero.pins.pigpiod:PiGPIOFactory',
+        'rpigpio = gpiozero.pins.rpigpio:RPiGPIOFactory',
+        'rpio    = gpiozero.pins.rpio:RPIOFactory',
+        'native  = gpiozero.pins.native:NativeFactory',
+        'mock    = gpiozero.pins.mock:MockFactory',
+        'mockpwm = gpiozero.pins.mock:MockPWMFactory',
```
lurch: Is MockPWMFactory missing from the mock class?

gpiozero/pins/pigpiod.py
```diff
@@ -99,133 +157,96 @@ class PiGPIOPin(Pin):
...
-        # naughty and check an internal variable instead...
-        if self._connection.sl.s is not None:
+        if self.factory.connection:
             self.frequency = None
             self.when_changed = None
             self.function = 'input'
-            self.pull = 'up' if self._pi_info.pulled_up('GPIO%d' % self.number) else 'floating'
+            self.pull = 'up' if self.factory.pi_info.pulled_up('GPIO%d' % self.number) else 'floating'
+
+    def _get_address(self):
```
lurch: This appears to be identical to the parent-class (PiPin) version, so could be removed from here?

gpiozero/pins/pi.py
```diff
@@ -0,0 +1,214 @@
...
+
+class PiPin(Pin):
+    """
+    Abstract base class representing a multi-function GPIO pin attached to a
+    Raspberry Pi.
+    """
+    def __init__(self, factory, number):
+        super(PiPin, self).__init__()
+        try:
+            factory.pi_info.physical_pin('GPIO%d' % number)
```
lurch: Given that there's **_so_** many occurrences of `'GPIO%d'`scattered throughout the code, would a `_get_str_name` type method be useful?

gpiozero/pins/mock.py
```diff
@@ -413,3 +385,34 @@ def tx_word(self, value, bits_per_word=None):
             bits = reversed(bits)
         self.tx_buf.extend(bits)
 
+
+class MockFactory(LocalPiFactory):
+    def __init__(self, revision='a21041', pin_class=MockPin):
```
lurch: I wonder if these factory-parameters should also be settable through env-vars, like the pigpio factory params are? (to make testing from the command-line easier)

gpiozero/pins/mock.py
```diff
@@ -413,3 +385,34 @@ def tx_word(self, value, bits_per_word=None):
...
+        self._revision = revision
+        self.pin_class = pin_class
+
+    def _get_address(self):
+        return ('mock',)
+
+    def _get_revision(self):
+        return self._revision
+
+    def reset(self):
```
lurch: I wonder if this should be named `_reset`, given that it's a method that no other factories have?

lurch:
Seems there's a few files that have had `import weakref` added, but then not actually use it?

lurch:
Just a few selected comments for now, rather than a full 'review', but it's a start... ;-)

gpiozero/pins/pigpiod.py
waveform80: Doh, you're absolutely right - must've missed that when I was re-factoring to pull all the common stuff into PiPin.

gpiozero/pins/pi.py
waveform80: Good idea ... half tempted to just use `repr(pin)` for this given that's what it currently returns (and it sort of fits; it's the canonical representation of the pin).

gpiozero/pins/pi.py
```diff
@@ -0,0 +1,214 @@
...
+
+    @property
+    def number(self):
+        return self._number
+
+    @property
+    def factory(self):
+        return self._factory
+
+    def _get_address(self):
```
waveform80: I wondered about this myself (which partially explains why there's a `__repr__` for the various SPI descendent classes although hardware vs software currently have different schemes). But I couldn't come up with a good use-case that the reservation/conflict system didn't (or couldn't in future) already cover so I figured "leave it out for now, we can always add it later".

gpiozero/pins/mock.py
waveform80: Well, I figured it's called externally (by `test_mock_pin.teardown_function`) so it's effectively part of the "public" API. It may not be part of the parent's public API (or any other descendant's) but then it's a fairly special descendant so I don't have any particular issue with it having a slightly "enhanced" API compared to the rest.

waveform80:
> Just a few selected comments for now, rather than a full 'review', but it's a start... ;-)

Much appreciated!

gpiozero/pins/spi.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        Writes data (a list of integer words where each word is assumed to have
+        :attr:`bits_per_word` bits or less) to the SPI interface, and reads an
+        equivalent number of words, returning them as a list of integers.
+        """
+        result = []
+        with self.lock:
+            shift = operator.lshift if lsb_first else operator.rshift
+            for write_word in data:
+                mask = 1 if lsb_first else 1 << (bits_per_word - 1)
```
waveform80: Good question. At the moment the answer is "yes" because every single SPI device we've implemented (the MCP3xxx series of ADCs) use multi-byte transfers. Whether that'll continue to be the case with future devices, I don't know, but it can't hurt so it's probably worthwhile doing (also worth noting that's one of the performance critical bits of code given it's a bit-banging SPI implementation, so any performance enhancements are worth considering).

waveform80:
> Seems there's a few files that have had import weakref added, but then not actually use it?

Ah, I'll see what I can find. I'll push an additional commit in a bit to avoid wiping out the existing comments (I'll have to rebase the pin-numbering branch at some point but that can wait until this is done).

gpiozero/pins/mock.py
waveform80: `revision` probably should be settable by env-vars (and it'd be trivial to add; though I'm not sure if any tests currently depend on the default?).

Whether `pin_class` should be ... erm ... not sure. It's a bit of a hodge-podge at the moment anyway given the pin class can be specified in numerous places (the factory constructor, after construction by simply re-writing the `pin_class` field, or even when calling the `pin()` method where the class is an optional parameter). Is there much difference between writing, say, `$ MOCK_PIN_CLASS=MockPWMPin python` and `>>> factory = MockFactory(pin_class=MockPWMPin)` or am I missing a use-case here?

docs/api_pins.rst
```diff
@@ -11,68 +11,68 @@ are concerned with. However, some users may wish to take advantage of the
...
+.. code-block:: console
 
-.. warning::
-
-    While the devices API is now considered stable and won't change in
-    backwards incompatible ways, the pins API is *not* yet considered stable.
-    It is potentially subject to change in future versions. We welcome any
-    comments from testers!
+    $ export GPIOZERO_PIN_FACTORY=pigpio
+    $ PIGPIO_ADDR=remote-pi python3 my_script.py
```
lurch: I wonder if doing this as a one-liner would be more useful?
`GPIOZERO_PIN_FACTORY=pigpio PIGPIO_ADDR=remote-pi python3 my_script.py`

Less danger of the user getting confused if they forget to un-set GPIOZERO_PIN_FACTORY ;-)

gpiozero/__init__.py
```diff
@@ -6,56 +6,19 @@
...
-    PinMultiplePins,
-    PinNoPins,
-    GPIOZeroWarning,
-    SPIWarning,
-    SPISoftwareFallback,
-    PinWarning,
-    PinNonPhysical,
-)
+# Yes, import * is naughty, but exc imports nothing else so there's no cross
+# contamination here ... and besides, have you *seen* the list lately?!
```
lurch: LOL :)

gpiozero/devices.py
```diff
@@ -229,13 +189,100 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
     """
     Represents a single device of any type; GPIO-based, SPI-based, I2C-based,
-    etc. This is the base class of the device hierarchy. It defines the
-    basic services applicable to all devices (specifically the :attr:`is_active`
+    etc. This is the base class of the device hierarchy. It defines the basic
+    services applicable to all devices (specifically the :attr:`is_active`
     property, the :attr:`value` property, and the :meth:`close` method).
     """
+    _pin_factory = None # instance of a Factory sub-class
+    _reservations = {} # maps pin addresses to lists of devices
```
lurch: I'm obviously not following what the code is doing, but why is `_reservations` stored on the Device and not the PinFactory?

gpiozero/devices.py
```diff
@@ -229,13 +189,100 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+        conflicts with alternate pin functions like SPI.
+        """
+        addresses = (
+            p.address if isinstance(p, Pin) else p
+            for p in pins_or_addresses
+            )
+        with self._res_lock:
+            for address in addresses:
+                try:
+                    conflictors = self._reservations[address]
```
lurch: Should this be `Device._reservations` to make it clear that it's not an instance-level attribute? (or am I still misunderstanding things?)

gpiozero/devices.py
```diff
@@ -229,13 +189,100 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+                    conflictors = []
+                    self._reservations[address] = conflictors
+                for device_ref in conflictors:
+                    device = device_ref()
+                    if device is not None and self._conflicts_with(device):
+                        raise GPIOPinInUse(
+                            'pin %s is already in use by %r' % (
+                                '/'.join(address), device)
+                        )
+                conflictors.append(weakref.ref(self))
```
lurch: IMHO this looks a bit odd - seems like you're only modifying a local variable.
I realise the code is actually equivalent (object references etc.), but I wonder if
```python
self._reservations[address].append(weakref.ref(self))
```
would make it clearer that this method *is* actually modifying instance-state. (IYSWIM?)

gpiozero/devices.py
```diff
@@ -441,3 +484,41 @@ def __repr__(self):
...
+        # outside the base distribution
+        #
+        # We prefer RPi.GPIO here as it supports PWM, and all Pi revisions. If
+        # no third-party libraries are available, however, we fall back to a
+        # pure Python implementation which supports platforms like PyPy
+        dist = pkg_resources.get_distribution('gpiozero')
+        for name in ('rpigpio', 'rpio', 'pigpio', 'native'):
+            try:
+                return pkg_resources.load_entry_point(dist, group, name)()
+            except Exception as e:
```
lurch: If we're on a RPi3 with only the 'rpio' and 'pigpio' entry points available, would it select 'rpio' as the default factory (and then later fail when actually trying to use it, as rpio doesn't work on the Pi3), or is it clever enough to actually select 'pigpio' in that case?

gpiozero/pins/__init__.py
```diff
@@ -6,39 +8,133 @@
...
+    """
+    Generates pins, SPI, and I2C interfaces for devices. This is an abstract
+    base class for pin factories. Descendents must override:
+
+    * :meth:`_get_address`
+    * :meth:`pin_address`
```
lurch: Only a minor point, but I wonder if these "must implement" methods should appear first in the actual list of method definitions below?

gpiozero/pins/__init__.py
```diff
@@ -6,39 +8,133 @@
...
+        specified pin.
+
+        .. warning::
+
+            Descendents must ensure that pin instances representing the same
+            hardware are identical; i.e. two separate invocations of
+            :meth:`pin` for the same pin specification must return the same
+            object.
+        """
+        raise PinGPIOUnsupported("GPIO not supported by this pin factory")
```
lurch: Bit confusing... the docstring for the Pin class says "Abstract base class representing a pin attached to some form of controller, be it GPIO, SPI, ADC, etc." and yet the default exception here explicitly mentions "GPIO"?

docs/api_pins.rst
waveform80: Possibly - I was trying to demonstrate both methods in a single example, which is possibly too ambitious.

gpiozero/devices.py
waveform80: Ah, that's a subtle point:

1. If someone's evil enough to use multiple local factories the reservation system will still catch them
2. Someone may legitimately use multiple factories if they're, e.g. remote PiGPIOFactory instances. But still the reservation system has to remain centralized (it's not perfect, i.e. it won't defend against "localhost/GPIO12" vs "127.0.0.1/GPIO12" but you can't block *every* foot-gun combination ;)

gpiozero/devices.py
waveform80: Yes, that should be - I'll get it fixed up

gpiozero/pins/__init__.py
```diff
@@ -205,6 +312,26 @@ def _set_bounce(self, value):
...
+
+            HIGH - - - - >       ,--. ,--------------. ,--.
+                                 |  | |              | |  |
+                                 |  | |              | |  |
+            LOW  ----------------'  `-'              `-'  `-----------
+                                 :                     :
+                                 :                     :
+                           when_changed          when_changed
+                               fires                 fires
+
```
lurch: Hmmm, I guess this means that the 3rd part of #378 didn't get addressed yet?

gpiozero/pins/local.py
```diff
@@ -0,0 +1,240 @@
...
+    :class:`RPIOPin`, and :class:`NativePin`).
+    """
+    pins = {}
+
+    def __init__(self):
+        super(LocalPiFactory, self).__init__()
+        self.spi_hardware_class = LocalPiHardwareSPI
+        self.spi_software_class = LocalPiSoftwareSPI
+        self.shared_spi_hardware_class = LocalPiHardwareSPIShared
+        self.shared_spi_software_class = LocalPiSoftwareSPIShared
```
lurch: I wonder if this would be neater as a dict?
```python
self.spi_classes = {
    'hardware': LocalPiHardwareSPI,
    'software': LocalPiSoftwareSPI,
    'shared_hardware': LocalPiHardwareSPIShared,
    'shared_software': LocalPiSoftwareSPIShared,
}
```

gpiozero/pins/mock.py
lurch: > Whether pin_class should be ... erm ... not sure.

I was thinking specifically of the other Mock... pins (MockSPIClockPin, MockChargingPin, etc.) but perhaps they're a bit too special-case to be selectable from the command-line, and should always just be selected in code instead.

gpiozero/pins/mock.py
lurch: The bit that concerned me was seeing `Device._pin_factory.reset()` which kinda implies that all Device pin_factories have a reset() method?

I think it's something I might have touched on before, but it would be great if some/all of the MockPin unit tests could also (somehow) be (manually?) run using real pins on a RPi.

gpiozero/pins/native.py
```diff
@@ -202,113 +206,86 @@ class NativePin(LocalPin):
...
     def close(self):
+        self.frequency = None
         self.when_changed = None
         self.function = 'input'
-        self.pull = 'up' if self.PI_INFO.pulled_up('GPIO%d' % self.number) else 'floating'
+        self.pull = 'up' if self.factory.pi_info.pulled_up(self.address[-1]) else 'floating'
 
     def _get_function(self):
-        return self.GPIO_FUNCTION_NAMES[(self._MEM[self._func_offset] >> self._func_shift) & 7]
+        return self.GPIO_FUNCTION_NAMES[(self.factory.mem[self._func_offset] >> self._func_shift) & 7]
```
lurch: IMHO directly accessing `self.factory.mem` looks a little weird. I wonder if the init should do `self._mem = factory.mem`, and then all the methods could just access `self._mem` ?

gpiozero/pins/pi.py
lurch: Fair enough.

gpiozero/pins/pi.py
```diff
@@ -0,0 +1,214 @@
...
+        return self.factory.address + ('GPIO%d' % self.number,)
```
lurch: Nice to see that code elsewhere has been refactored to use this. To save repeated string-interpolation, would it be worth caching this? i.e.
```python
    def _get_address(self):
        if self._address is None:
            self._address = self.factory.address + ('GPIO%d' % self.number,)
        return self._address
```

gpiozero/pins/pigpiod.py
```diff
@@ -68,10 +71,65 @@ class PiGPIOPin(Pin):
...
     """
+    def __init__(
+            self, host=os.getenv('PIGPIO_ADDR', 'localhost'),
+            port=int(os.getenv('PIGPIO_PORT', 8888))):
+        super(PiGPIOFactory, self).__init__()
+        self.pin_class = PiGPIOPin
+        self.spi_hardware_class = PiGPIOHardwareSPI
+        self.spi_software_class = PiGPIOSoftwareSPI
+        self.shared_spi_hardware_class = PiGPIOHardwareSPIShared
+        self.shared_spi_software_class = PiGPIOSoftwareSPIShared
```
lurch: See my previous dict comment ;)

gpiozero/pins/pigpiod.py
```diff
@@ -68,10 +71,65 @@ class PiGPIOPin(Pin):
...
+        self._connection = pigpio.pi(host, port)
+        self._host = host
+        self._port = port
+        self._spis = []
```
lurch: Should this be a set rather than a list?

gpiozero/pins/pigpiod.py
```diff
@@ -68,10 +71,65 @@ class PiGPIOPin(Pin):
...
-    _CONNECTIONS = {} # maps (host, port) to (connection, pi_info)
-    _PINS = {}
+    def close(self):
+        super(PiGPIOFactory, self).close()
+        # We *have* to keep track of SPI interfaces constructed with pigpio;
+        # if we fail to close them they prevent future interfaces from using
+        # the same pins
+        if self.connection:
+            while self._spis:
+                self._spis[0].close()
```
lurch: LOL, these two lines look a bit 'magic' ;-)

gpiozero/pins/__init__.py
waveform80: Good point - should probably have a more generic exception class too

waveform80: Might just change this to NotImplementedError for now

gpiozero/pins/native.py
lurch: ...or maybe I simply haven't got into the right "factory mindset" yet.

gpiozero/pins/pigpiod.py
```diff
@@ -259,20 +277,224 @@ def _set_when_changed(self, value):
...
+            self._bits_per_word << 16
+            )
+
+    def _get_clock_mode(self):
+        return self._clock_mode
+
+    def _set_clock_mode(self, value):
+        self._check_open()
+        if not 0 <= value < 4:
+            raise SPIInvalidClockmode("%d is not a valid SPI clock mode" % value)
```
lurch: Seems like this identical check occurs in multiple places - I wonder if it could be moved to some parent class?

gpiozero/pins/data.py
```diff
@@ -363,12 +363,12 @@
...
-    0x12:     ('A+',   '1.1', '2014Q4', 'BCM2835', 'Sony',      256,  'MicroSD', 1,  0,  False, False, 1,  1,  {'P1': PLUS_P1},                APLUS_BOARD,  ),
-    0x13:     ('B+',   '1.2', '2015Q1', 'BCM2835', 'Egoman',    512,  'MicroSD', 4,  1,  False, False, 1,  1,  {'P1': PLUS_P1},                BPLUS_BOARD,  ),
+    0x12:     ('A+',   '1.1', '2014Q4', 'BCM2835', 'Sony',      256,  'MicroSD', 1,  0,  False, False, 1,  1,  {'J8': PLUS_J8},                APLUS_BOARD,  ),
+    0x13:     ('B+',   '1.2', '2015Q1', 'BCM2835', 'Egoman',    512,  'MicroSD', 4,  1,  False, False, 1,  1,  {'J8': PLUS_J8},                BPLUS_BOARD,  ),
     0x14:     ('CM',   '1.1', '2014Q2', 'BCM2835', 'Embest',    512,  'eMMC',    1,  0,  False, False, 2,  2,  {'SODIMM': CM_SODIMM},          CM_BOARD,     ),
-    0x15:     ('A+',   '1.1', '2014Q4', 'BCM2835', 'Embest',    256,  'MicroSD', 1,  0,  False, False, 1,  1,  {'P1': PLUS_P1},                APLUS_BOARD,  ),
+    0x15:     ('A+',   '1.1', '2014Q4', 'BCM2835', 'Embest',    256,  'MicroSD', 1,  0,  False, False, 1,  1,  {'J8': PLUS_J8},                APLUS_BOARD,  ),
```
waveform80: Yeah, they should be ... I'll see if I can find some time to pull those out into the pin numbering PR; that's probably where it belongs

gpiozero/pins/local.py
waveform80: That could indeed be neater - perhaps keyed by a (hardware, shared) tuple of bools. I'll try it and see how it looks.

gpiozero/pins/pigpiod.py
```diff
@@ -259,20 +277,224 @@ def _set_when_changed(self, value):
...
+
+    def _get_select_high(self):
+        return self._select_high
+
+    def _set_select_high(self, value):
+        self._check_open()
+        self._factory.connection.spi_close(self._handle)
+        self._select_high = bool(value)
+        self._handle = self._factory.connection.spi_open(
+            self._device, self._baud, self._spi_flags())
```
lurch: Took me a while to figure out how on earth this was working!
Maybe it'd be more obvious if it replaced `self._select_high = bool(value)` with something like `self._set_spi_flag('select_high', bool(value)` ?

In fact, thinking some more, I guess the `_set_spi_flag` could also be responsible for closing and opening the factory connection bits. Which would make all theses PiGPIO SPI prop-setters much simpler :)

gpiozero/pins/pigpiod.py
```diff
@@ -259,20 +277,224 @@ def _set_when_changed(self, value):
...
+        except DeviceClosed:
+            return 'SPI(closed)'
+
+    def _spi_flags(self):
+        return (
+            self._mode          << 0  |
+            self._select_high   << 2  |
+            self._lsb_first     << 14 |
+            self._lsb_first     << 15
+            )
```
lurch: Hmmm, why is this different to the _spi_flags for PiGPIOHardwareSPI ?

setup.py
lurch: Oh? I can find a `class MockFactory` in this PR, but I can't find a `class MockPWMFactory` ? (or is there more automagic code going on behind the scenes?)

tests/test_devices.py
```diff
@@ -6,86 +6,87 @@
...
-    with GPIODevice(pin) as device:
-        with GPIODevice(pin2) as device2:
+    with GPIODevice(2) as device:
+        with GPIODevice(3) as device2:
             pass
 
 def test_device_close():
-    pin = MockPin(2)
-    device = GPIODevice(pin)
+    device = GPIODevice(2)
```
lurch: Looking at the other tests, should tests like this also be refactored to `with GPIODevice(2) as device:`?
Is there any appreciable difference, or is it just a stylistic thing?

tests/test_outputs.py
```diff
@@ -872,17 +882,21 @@ def test_motor_value_nonpwm():
...
+    b = Device._pin_factory.pin(2, pin_class=MockPWMPin)
     with Motor(f, b) as device:
         with pytest.raises(ValueError):
             device.value = -2
         with pytest.raises(ValueError):
             device.value = 2
+        with pytest.raises(ValueError):
+            device.forward(2)
+        with pytest.raises(ValueError):
+            device.backward(2)
```
lurch: Haha #492  :wink:

docs/api_pins.rst
lurch: https://www.youtube.com/watch?v=Y3lQSxNdr3c

gpiozero/devices.py
lurch: Ah, that solves my previous question :)

gpiozero/pins/__init__.py
lurch: Hmmm, I wonder if a Pin object returned by a factory should have a "list of possible functions"?
With the obvious understanding that we may not know until much later whether a specific function is actually usable on a specific pin.

gpiozero/pins/__init__.py
lurch: Dunno, it's not something I've ever looked at in detail. I guess one way of doing it would be to only report state-changes when the state has remained constant for a specific time, but then you obviously lose your 'sharp' edge-detection (which I guess implies that it'd be handled by a separate attribute entirely).

gpiozero/devices.py
```diff
@@ -265,13 +260,13 @@ def _release_all(self):
...
         with self._res_lock:
-            Device._reservations = {
+            Device._reservations = defaultdict(list, {
                 address: [
                     ref for ref in conflictors
                     if ref() not in (self, None)
                     ]
                 for address, conflictors in self._reservations.items()
-                }
+                })
```
lurch: I guess in effect, this is doing a filter-esque operation?

gpiozero/pins/pi.py
```diff
@@ -93,7 +95,7 @@ def spi(self, **spi_args):
         doesn't matter).
         """
         spi_args, kwargs = self._extract_spi_args(**spi_args)
-        shared = kwargs.pop('shared', False)
+        shared = 'shared' if kwargs.pop('shared', False) else 'exclusive'
```
lurch: Nice.

gpiozero/pins/pi.py
```diff
@@ -125,10 +124,7 @@ def spi(self, **spi_args):
             key: pin.number if isinstance(pin, Pin) else pin
             for key, pin in spi_args.items()
             }
-        if shared:
-            return self.shared_spi_software_class(self, **spi_args)
-        else:
-            return self.spi_software_class(self, **spi_args)
+        return self.spi_classes[('software', shared)](self, **spi_args)
```
lurch: Yay,  much neater :heart:

lurch:
>> Just a few selected comments for now, rather than a full 'review', but it's a start... ;-)

> Much appreciated!

Done a more thorough read-through, and added some more comments :-)  In case it's not obvious, I've only been reading the code online on github.com, I haven't actually tried running any of it on a real Pi yet. Something for Sunday perhaps... ;)

Always nice to see that my comments are actually useful :-)

docs/api_pins.rst
waveform80: Heh - that takes me back :)

setup.py
waveform80: Doh ... you're absolutely right. There is no MockPWMFactory - can't remember why that's there. Probably a leftover from the initial re-factoring stages. I'll drop it in a mo (there should only be MockFactory)

gpiozero/output_devices.py
```diff
@@ -562,8 +562,12 @@ def __init__(
...
         LEDClass = PWMLED if pwm else LED
         super(RGBLED, self).__init__()
-        self._leds = tuple(LEDClass(pin, active_high) for pin in (red, green, blue))
-        self.value = initial_value
+        try:
+            self._leds = tuple(LEDClass(pin, active_high) for pin in (red, green, blue))
+            self.value = initial_value
+        except:
+            self.close()
+            raise
```
lurch: ...and also for ButtonBoard, SnowPi, FishDish, TrafficHat, Robot, _EnergenieMaster, Motor and Servo? (TBH I'm not sure exactly what the rules are as to when it is or isn't needed)

setup.py
lurch: Hmmm? Wouldn't it be better to **add** MockPWMFactory so that you can do `GPIOZERO_PIN_FACTORY=mockpwm python3 my_script.py` ??

gpiozero/pins/mock.py
waveform80: Must admit I hadn't considered running the mock pin tests on a real Pi (I figured running the real pin tests should be sufficient, assuming we can achieve decent coverage on them). But it's an interesting idea. I think for now I'm going to err on the side of not adding too much - if we move reset up the hierarchy later it won't be an API break as it'll just mean adding a method to the other factories. I'd like to get the "real" pin tests cleaned up a bit before we tackle adding more stuff to them (and there's some mock tests that I doubt we'll ever support on real pins; e.g. testing pins where frequency is unsupported).

gpiozero/pins/__init__.py
waveform80: I should probably word things a bit more clearly, but I think the lists are okay as they are - basically it's a mandatory override list followed by an optional override list (so implementers know the stuff they should tackle first to get something basic working, and what they can then explore to expand their implementation from there). I'll tweak the prose a bit to make it a little clearer.

lurch: I probably didn't explain myself clearly enough either. I wasn't suggesting changing the doc-strings, but instead swapping around the actual method implementations themselves, so that `def _get_address(self):` and `def pin_address(self, spec):` appear earlier in the file than `def close(self):` (if that makes more sense now?)

gpiozero/pins/native.py
waveform80: Well, the factory can be seen as the thing that owns the GPIO registers - I quite like the symmetry of having a single object that represents that single block of memory, with the pin objects representing operations against subsets of that block.

gpiozero/pins/pi.py
waveform80: If it were being used in tight loops, definitely. But I think at the moment it's only used on initialization (and initialization related things like setting pull-ups). Premature optimization in other words.

gpiozero/pins/data.py
lurch: Have a good evening :)

BTW - Argh, it's so hard to find the right comment to reply to on a big PR like this! Stupid github :-(

gpiozero/pins/pi.py
lurch: ...and reprs.

> Premature optimization in other words.

Fair enough.

setup.py
waveform80: If that's a common use-case I'd be more tempted to make the pin-class configurable from the command line as it's more generic and hence will cover more use-cases with a single change (under the assumption we're bound to add more pin classes for things like analog etc. and I'd rather avoid adding yet more mock factory entry points)

gpiozero/output_devices.py
waveform80: Basically it's needed whenever you construct some subordinate Device, then fail to close those subordinates if your own constructor subsequently fails. LEDCollection was definitely an example where that was needed. ButtonBoard needs it too. As it happens SnowPi and TrafficLights don't need it because LEDBoard descends from LEDCollection and that's handling clean-up of any failed LED constructions. FishDish needs it because it constructs its own Buzzer and Button, same story for TrafficHat (note: TrafficLights is *not* the issue here, just the Buzzer and Button). Robot doesn't need it, but the subordinate Motor class does, and so does Servo and _EnergenieMaster. 

I'll get all these fixed up.

waveform80: Hmm, actually looking at it ... it's a little more complex than I thought because some Device references will be lost already in various comprehensions. Hmm. Might be time to tackle the "internal" references that keep certain things like Button alive without an external reference. I'll dig deeper into this tonight.

gpiozero/pins/__init__.py
waveform80: Oh right ... erm, not sure really. I think keeping getters and setters together is generally a good idea and there's cases where a getter is mandatory but the corresponding setter is optional. Would you favour breaking them up in that case?

gpiozero/pins/__init__.py
waveform80: It's an interesting idea; like a "capabilities" property that lists whether the pin can handle PWM output, analog-in, etc. It's a decent idea but I think one that can marked "future work" for now (open a ticket for it?)

gpiozero/pins/data.py
waveform80: I suspect that's more my fault for submitting such a huge PR - unfortunately sometimes it's just necessary. At the moment I'm working off e-mail notifications as a quick way to find things you've replied to (sometimes the web just isn't enough :)

gpiozero/pins/pigpiod.py
```diff
@@ -68,10 +71,65 @@ class PiGPIOPin(Pin):
...
+
+    def _get_revision(self):
+        return self.connection.get_hardware_revision()
 
+    def _get_address(self):
+        return ("%s:%d" % (self.host, self.port),)
+
+    def spi(self, **spi_args):
+        intf = super(PiGPIOFactory, self).spi(**spi_args)
+        self._spis.append(intf)
```
waveform80: Ahh ... there's *meant* to be some magic going on, but I've forgotten it! Basically the reservation system is meant to kill the construction of SPI interfaces on incompatible sets of pins ... which it does for the local SPI implementations (e.g. `LocalPiHardwareSPI._conflicts_with`) but I've completely forgotten to override `_conflicts_with` for the PiGPIO implementations. And *this* is why code review is a great thing ;)

gpiozero/pins/pigpiod.py
waveform80: Not really much point given that (once I've fixed up the prior comment about the missing conflicts ... oops!) duplicate SPI interfaces won't be able to be constructed anyway.

gpiozero/pins/pigpiod.py
waveform80: Yeah - it is a bit "magic" having SPI interfaces remove themselves from the owning factory's list; the reason for this is that SPI interfaces are basically devices in their own right (i.e. it's conceivable users may create them themselves or use them in their own SPI device classes).

Particularly in the case of pigpio we *must* keep track of these SPI interfaces because if you fail to close remote SPI interfaces, the pigpio daemon refuses to let you create new ones on the same pin combinations (quite reasonably) unless you restart the daemon, of course. Hence why I figured it was best to have the SPI interface deal with this itself rather than leaving it in the hands of the owning device (also reduces code repetition a little, but the daemon state issue is my real reason here).

gpiozero/pins/pigpiod.py
waveform80: Hmm, good point - now I look at it, it is pretty obscure. I'll see if I can come up with something more obvious.

gpiozero/pins/spi.py
```diff
@@ -0,0 +1,91 @@
...
+                        self.mosi.value = bool(write_word & mask)
+                    self.clock.on()
+                    if self.miso is not None and not clock_phase:
+                        if self.miso.value:
+                            read_word |= mask
+                    self.clock.off()
+                    if self.miso is not None and clock_phase:
+                        if self.miso.value:
+                            read_word |= mask
+                    mask = shift(mask, 1)
```
waveform80: Yeah, that looks good - must read through the rest of the comments in #470

tests/test_devices.py
waveform80: There's no significant difference in CPython, but there is in pypy where objects have a tendency to hang around a *lot* longer before the GC collects them. So yes, it would be worth re-factoring any tests that don't use `with`. I'll see if I can fit that in (because I already had to do plenty to get the SPI stuff passing on pypy).

waveform80: Ah, on the other hand - that test is explicitly dealing with the `close` method so it should probably be left alone. I'll have a skim through the rest to see if there's any that need conversion though.

gpiozero/devices.py
waveform80: Precisely - just re-writes the dict with all references to self (and dead-refs) removed; easiest way to do it (and safe given it's all done under `_res_lock`).

gpiozero/pins/pi.py
waveform80: Ah, good point!

gpiozero/output_devices.py
lurch: Nice work Dave :-)  Another great example of how things need to get really complicated behind the scenes, to keep outward appearances easy-to-use.

I guess the specific change above can be un-done now?

gpiozero/pins/pi.py
```diff
@@ -0,0 +1,251 @@
...
+                if isinstance(value, MethodType):
+                    self._when_changed = WeakMethod(value)
+                else:
+                    self._when_changed = ref(value)
+                self._enable_event_detect()
+            elif self._when_changed is not None and value is None:
+                self._disable_event_detect()
+                self._when_changed = None
+            elif value is None:
+                self._when_changed = None
```
lurch: I guess the previous
```python
            elif self._when_changed is not None and value is None:
                self._disable_event_detect()
                self._when_changed = None
            elif value is None:
                self._when_changed = None
```
could be collapsed to just:
```python
            elif value is None:
                if self._when_changed is not None:
                    self._disable_event_detect()
                self._when_changed = None
```
(and maybe something similar could be done to merge the `value is not None` cases?)

gpiozero/output_devices.py
waveform80: Ooops - yup, I'll chop that out in a mo - just wrestling with getting weak-method refs working in py2 (stupid errors on my part in the WeakMethod backport).

gpiozero/pins/pi.py
waveform80: Yeah - I need to tidy that up a bit; it used to have a nice "when setting", "when unsetting", "when changing" branch structure and now it's all messy. I'll have a crack at that in a mo.

waveform80:
Okay, Travis seems to playing nice again ... I've no idea why it was skipping the mock pin tests before but it's okay now. I'm pretty much happy with the PR at this point -- so unless anyone has any objections I'll squish and merge it later this week (got Hack Mcr Jr tomorrow and Tuesday so it'll be at least Wednesday before I get back to this).

lurch:
> unless anyone has any objections I'll squish and merge it later this week

I'd still like to try experimenting with this code on a real Pi - hopefully I'll be able to squeeze that in on Monday or Tuesday evening.

Makefile
```diff
@@ -101,7 +101,7 @@ develop: tags
 	$(PIP) install -e .[doc,test]
 
 test:
-	$(COVERAGE) run -m $(PYTEST) tests -v
+	$(COVERAGE) run -m $(PYTEST) tests -v -r sx
```
lurch: If things are working, should this temporary change be removed again?

waveform80: I considered it, but I'm quite liking the fact it now outputs precisely why it skipped particular tests - any objections to keeping it in?

waveform80:
> >  unless anyone has any objections I'll squish and merge it later this week

> I'd still like to try experimenting with this code on a real Pi - hopefully I'll be able to squeeze that in on Monday or Tuesday evening.

No prob - my major goal is to get this into 1.4. I've been running tests on a Pi fairly regularly, but I've only tried remote SPI a few times (and not recently - had my Pi's all unplugged for Hack Manchester Jr - will give it a go now that most are back online).

Makefile
lurch: I haven't tested what that option does, I just assumed it was a temporary debugging measure. But if it makes sense to keep it in permanently, then that's fine :-)

lurch:
I've been busy all this week, and I'm away all this weekend, so hopefully I'll get a chance to experiment with this next week...
Did Hack Mcr go well?

waveform80:
> I've been busy all this week, and I'm away all this weekend, so hopefully I'll get a chance to experiment with this next week...

Here too - in the midst of a house extension at the moment, so things are a little chaotic!

> Did Hack Mcr go well?

It certainly did (well, the junior event; the senior event is this weekend but I can't make that unfortunately). The video coverage of the event is somewhere on the intertubes but I'm proud to say a couple of the kids I was mentoring won one of the challenges (with some Scratch-based robotics) and there were some interesting steganography attempts from the teams doing the GCHQ challenge.

bennuttall:
:fireworks:

WayneKeenan:
I tried this on a Mac talking to a PiZeroW running JessieLite.     Obviously had to apt install pigpio on Jessie lite.

One thing that wasn't in the example but was something I had to specify was the pin factory, so raising this just in case it's unexpected:

```bash
GPIOZERO_PIN_FACTORY=pigpio PIGPIO_ADDR=192.168.1.108 python
from gpiozero import PWMLED
led = PWMLED(17)
led.value=0.5
```

Without GPIOZERO_PIN_FACTORY this occurs:
```python
(gpiozero-env) shinytoo:python-gpiozero wayne$ PIGPIO_ADDR=192.168.1.108 python
Python 3.6.1 (default, Apr  4 2017, 09:40:21) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.38)] on darwin
>>> import gpiozero
/Users/wayne/local.projects/python-gpiozero/gpiozero/devices.py:520: PinFactoryFallback: Failed to load factory rpigpio: No module named 'RPi'
  'Failed to load factory %s: %s' % (name, str(e))))
/Users/wayne/local.projects/python-gpiozero/gpiozero/devices.py:520: PinFactoryFallback: Failed to load factory rpio: No module named 'RPIO'
  'Failed to load factory %s: %s' % (name, str(e))))
```



I originally tried to use the activity LED as this Zero had no header. Apologies if this is actually a separate issue, but when running the following: 

On the pi:
```
echo none | sudo tee /sys/class/leds/led0/trigger
echo gpio | sudo tee /sys/class/leds/led1/trigger
```
On the Mac:
```
GPIOZERO_PIN_FACTORY=pigpio PIGPIO_ADDR=192.168.1.108 python
from signal import pause
from gpiozero import LED

activity = LED(47)
activity.blink()

pause()
```

This error occurs:

```
>>> activity = LED(47)
/Users/wayne/local.projects/python-gpiozero/gpiozero/pins/pi.py:228: PinNonPhysical: no physical pins exist for GPIO47
  'no physical pins exist for %s' % self.address[-1]))
  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/pins/pi.py", line 59, in pin
    pin = self.pins[n]
KeyError: 47

During handling of the above exception, another exception occurred:

  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/pins/pigpiod.py", line 175, in __init__
    self.factory.connection.set_mode(self.number, pigpio.INPUT)
  File "/Users/wayne/local.projects/gpiozero-env/lib/python3.6/site-packages/pigpio.py", line 1291, in set_mode
    return _u2i(_pigpio_command(self.sl, _PI_CMD_MODES, gpio, mode))
  File "/Users/wayne/local.projects/gpiozero-env/lib/python3.6/site-packages/pigpio.py", line 964, in _u2i
    raise error(error_text(v))
pigpio.error: 'no permission to update GPIO'

During handling of the above exception, another exception occurred:

  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/devices.py", line 96, in __call__
    self = super(GPIOMeta, cls).__call__(*args, **kwargs)
  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/output_devices.py", line 132, in __init__
    super(DigitalOutputDevice, self).__init__(pin, active_high, initial_value)
  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/output_devices.py", line 41, in __init__
    super(OutputDevice, self).__init__(pin)
  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/mixins.py", line 69, in __init__
    super(SourceMixin, self).__init__(*args, **kwargs)
  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/devices.py", line 443, in __init__
    pin = Device._pin_factory.pin(pin)
  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/pins/pi.py", line 61, in pin
    pin = self.pin_class(self, n)
  File "/Users/wayne/local.projects/python-gpiozero/gpiozero/pins/pigpiod.py", line 177, in __init__
    raise ValueError(e)

```

waveform80:
> The above blinks an LED, but without GPIOZERO_PIN_FACTORY this occurs:...

Those warnings are actually expected (though for some utterly bizarre reason I haven't yet fathomed they don't occur on Ben's laptop!). Anyway, they're normal: by way of explanation there's four pin implementations underlying GPIO Zero which, in the absence of an explicit specification via the GPIOZERO_PIN_FACTORY env-var, the library attempts to load in the following order: RPi.GPIO, RPIO, pigpio, and "native" (a hacky implementation I threw together as a back-stop, which I ought to revise at some point). You got two warnings because it tried (and failed) to load RPi.GPIO, then RPIO, then succeeded when it reached pigpio.

The reason the warnings are there is that it's quite common for more advanced Python developers to work in a virtualenv; GPIO Zero doesn't *directly* depend on any of the pin implementations (after all it could be installed on a non-Pi, or someone might actually want to use "native" ... for some strange reason :) so installing GPIO Zero in the virtualenv doesn't pull in any of the other implementations. At that point the developer fires up Python, tries it out and wonders why it works ... sort of, but all the PWM fails. The reason is that GPIO Zero (in the absence of any pin libraries) has fallen back to the hacky native implementation which doesn't do PWM. So, the warnings were added to make it more obvious to people working in a virtualenv that they've probably forgotten to install whichever pin implementation they want.

Now ... in future the plan is eventually to switch pigpio to the top of the list so the warnings would disappear in your particular scenario, but we're not *quite* there yet.

@bennuttall aha! Are you sure there's nothing weird in *your* Python config? ;)

> I originally tried to use the activity LED as this Zero had no header. Apologies if this is actually a separate issue...

Ahh, that's interesting. I hadn't tried using activity LEDs remotely yet. The first bit there ("PinNonPhysical") isn't actually an exception; it's just another warning that we fire off when a GPIO is specified that isn't present on any known header. However, after that it appears something rather more serious goes wrong. A quick guess is that the pigpio daemon might refuse to serve GPIOs outside the normal range, which would be an upstream issue but we ought to have some better error handling there.

WayneKeenan:
ah, I thought the warnings were fatal errors and so I stopped, I should have just pressed ahead, it works ok, as expected.  My excuse is the usual: it was early and pre-coffee :)

Although this may soon to be a moot point when `pigpio` becomes king, but, perhaps change the wording to output something like `logger.info("Optional library XXX not found.")` - I did this in one of my libraries to lower end user confusion.

Issue #467, opened 2016-09-25
Pin numbering


waveform80:
Following on from #459 ... here's the contentious change I've currently got in the patch: pin numbering.

Yeah, that old chestnut. Funny story...

So, @bennuttall and I are at PyConUK and we bump into @gbaman and get chatting about GPIO Zero. Now, it turns out that @gbaman might be _almost_ as stubborn as I am, to the extent that his local RJam has voted to reject GPIO Zero purely on the basis of its pin numbering (despite acknowledging that it's very nice in most other respects).

Obviously I think this is a bit crazy: why reject plenty of other advantages over the single issue of pin numbering, which arguably isn't _that_ much more complex than board numbering once one takes into account the necessity of still teaching that not all pins are equal.

Nonetheless we get to discussing it, and I make the case that adding mutable global state (a la "GPIO.setmode") is a bad idea because it leaves your script open to the whims of anything you import that might change it (if anyone thinks this is unlikely: go look at early versions of picamera where I had no choice but to force BCM mode to control the camera LED because I couldn't query the mode back then; yes, it most certainly can happen).

However, I then mention the possibility that we _could_ support alternate numberings anyway via an expanded definition of a pin specification. In other words, permit more than just a straight integer GPIO number to be specified. I think I'm right in characterizing his response as "excited" at this possibility :)

So, why acquiesce when I still think their position is, as stated above, "crazy"?
1. I want to encourage as many as possible to use GPIO Zero, because I'm firmly convinced it's better for teaching physical computing.
2. I don't have a magic switch I can flip to convince @gbaman to change his mind, and I judge his stubbornness to be akin to my own.
3. I _do_ have the ability to implement an alternate scheme that he (and hopfully his local RJam) will find acceptable.
4. If I want to advance point 1 above, the only thing _I_ (or I think _we_) can do is try point 3. So that's what I'm attempting.

I'm not proposing we remove the current integer GPIO addressing, or that we use these alternate schemes in _any_ of our recipes, or encourage its use in any way (simply to minimize confusion on the part of users). As far as the Recipes chapter is concerned: we use integer GPIO addressing and that's it. However, in the PR for #459 I have added a brief bit under Notes detailing the alternate schemes available anywhere a pin specification is called for.

So, what alternate schemes have I (currently) included in the PR?
- Obviously the same old GPIO integer addressing: `LED(2)`
[...]

PR #470, opened 2016-09-27
Argh its pin numering again

waveform80:
This is the separated PR for #467. At the moment it looks horribly complex and long but actually it isn't (it's a tiny change to about 3 files). The reason it looks that way at the moment is that this extends my remote-spi branch, i.e. this PR builds on PR #468. Once that merges you should just see the second commit in this PR which is the actual pin numbering PR.

### DIFF
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -1078,6 +1078,93 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
         else:
             return self.headers[header].pins[number].pull_up
 
+    def find_gpio(self, spec):
+        """
+        Parses a pin *spec*, returning the equivalent Broadcom GPIO port number
+        or raising a :exc:`ValueError` exception if the spec does not represent
+        a GPIO port.
+
+        The *spec* may be given in any of the following forms:
+
+        * 'GPIOn' or 'BCMn' where n is the GPIO number
+        * 'WIPIn' or 'WPIn' where n is the `wiringPi`_ pin number
+        * 'PHYSn' where n is the physical pin number on header P1
+        * 'h:n' where h is the header name and n is the physical pin number
+          (for example P1:5 is physical pin 5 on header P1)
+
+        .. _wiringPi: http://wiringpi.com/pins/
+        """
+        spec = spec.upper()
+        if spec.startswith('GPIO') and spec[4:].isdigit():
+            gpio = int(spec[4:])
+            if not 0 <= gpio < 54:
+                raise PinInvalidPin('%s is not a valid GPIO port' % spec)
+            return gpio
+        elif spec.startswith('WPI') and spec[3:].isdigit():
+            main_head = 'P1' if 'P1' in self.headers else 'J8'
+            try:
+                return self.find_gpio({
+                    0:  '%s:11' % main_head,
+                    1:  '%s:12' % main_head,
+                    2:  '%s:13' % main_head,
+                    3:  '%s:15' % main_head,
+                    4:  '%s:16' % main_head,
+                    5:  '%s:18' % main_head,
+                    6:  '%s:22' % main_head,
+                    7:  '%s:7' % main_head,
+                    8:  '%s:3' % main_head,
+                    9:  '%s:5' % main_head,
+                    10: '%s:24' % main_head,
+                    11: '%s:26' % main_head,
+                    12: '%s:19' % main_head,
+                    13: '%s:21' % main_head,
+                    14: '%s:23' % main_head,
+                    15: '%s:8' % main_head,
+                    16: '%s:10' % main_head,
+                    17: 'P5:3',
+                    18: 'P5:4',
+                    19: 'P5:5',
+                    20: 'P5:6',
+                    21: '%s:29' % main_head,
+                    22: '%s:31' % main_head,
+                    23: '%s:33' % main_head,
+                    24: '%s:35' % main_head,
+                    25: '%s:37' % main_head,
+                    26: '%s:32' % main_head,
+                    27: '%s:36' % main_head,
+                    28: '%s:38' % main_head,
+                    29: '%s:40' % main_head,
+                    30: '%s:27' % main_head,
+                    31: '%s:28' % main_head,
+                    }[int(spec[3:])])
+            except KeyError:
+                raise PinInvalidPin('%s is not a valid wiringPi pin' % spec)
+        elif ':' in spec:
+            header, pin = spec.split(':', 1)
+            if pin.isdigit():
+                try:
+                    header = self.headers[header]
+                except KeyError:
+                    raise PinInvalidPin(
+                        'there is no %s header on the Pi' % header)
+                try:
+                    function = header.pins[int(pin)].function
+                except KeyError:
+                    raise PinInvalidPin(
+                        'no such pin %s on header %s' % (pin, header.name))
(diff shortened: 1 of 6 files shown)


gpiozero/devices.py
```diff
@@ -229,13 +189,100 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+        :meth:`_release_pins` for further information).
+        """
+        with self._res_lock:
+            Device._reservations = {
+                address: [
+                    ref for ref in conflictors
+                    if ref() not in (self, None)
+                    ]
+                for address, conflictors in self._reservations.items()
+                }
```
SrMouraSilva: What is the standard?

a) 
```python
address: [
    ref for ref in conflictors
    if ref() not in (self, None)
    ]
```

b) 
```python
warnings.warn(
    PinFactoryFallback(
        'Failed to load factory %s: %s' % (name, str(e))))
```

c)
```python
if pin in _PINS:
    raise GPIOPinInUse(
        'pin %r is already in use by another gpiozero object' % pin
    )

Comment (waveform80) [inline gpiozero/devices.py]: I think perhaps bracket closing indentation and newlines? Personally I favour (a) but if you read PEP-8's [indentation section](https://www.python.org/dev/peps/pep-0008/#indentation) it's actually pretty ambivalent about this. Frankly, my only reason for preferring (a) is simply that it's vim's default so it's the path of least resistance for me, but I don't really care what people use so long as it's readable and vaguely PEP-8 compliant (I think Ben favours (c) for example).

File: gpiozero/pins/spi.py [hunk 2]
```diff
@@ -0,0 +1,86 @@
...
+                    self.clock.on()
+                    if self.miso is not None and not clock_phase:
+                        if self.miso.value:
+                            read_word |= mask
+                    self.clock.off()
+                    if self.miso is not None and clock_phase:
+                        if self.miso.value:
+                            read_word |= mask
+                    mask = shift(mask, 1)
+                result.append(read_word)
```
waveform80: I think this comment is probably for #468 rather than this particular PR; the only reason these changes *currently* appear under here is that #470 branches off #468. Sorry for the confusion; that's my fault!

Still, I'm not sure that two nested loops is *that* complex especially when they're pretty simple loops; the current structure also makes it clear that a transfer happens entirely under the aegis of the bus' lock.

SrMouraSilva: > I think this comment is probably for #468 rather than this particular PR; the only reason these changes currently appear under here is that #470 branches off #468. Sorry for the confusion; that's my fault!

I saw this other, but I should comment this because it is more recent.

> Still, I'm not sure that two nested loops is that complex especially when they're pretty simple loops; the current structure also makes it clear that a transfer happens entirely under the aegis of the bus' lock.

_I need add IMHO in my comments._

The really question is the combination (if inside if inside for inside for inside with), for example

```python
with self.lock:
    for write_word in data:
        for _ in range(bits_per_word):
            if self.miso is not None and not clock_phase:
                if self.miso.value:
                   read_word |= mask
```

I spent between 10 ~ 30 minutes trying to understand, refactor and suggest.

gpiozero/devices.py
SrMouraSilva: Grateful for the clarification!

gpiozero/pins/spi.py
lurch: Yeah, that's just @waveform80 's penchant for writing exceedingly dense code ;)

> I spent between 10 ~ 30 minutes trying to understand, refactor and suggest.

Good luck trying to understand https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/spi.py#L308  !! :laughing:

lurch: ...although I _do_ wonder if e.g.
```python
    self.clock.on()
    if self.miso is not None and not clock_phase:
        if self.miso.value:
            read_word |= mask
    self.clock.off()
    if self.miso is not None and clock_phase:
        if self.miso.value:
            read_word |= mask
```
would be slightly more readable when rearranged as:
```python
    self.clock.on()
    # read bit when clock is on
    if not clock_phase:
        if self.miso is not None and self.miso.value:
            read_word |= mask
    self.clock.off()
    # read bit when clock is off
    if clock_phase:
        if self.miso is not None and self.miso.value:
            read_word |= mask
```

You could _even_ create a nested function and do:
```python
    def read_bit():
        return mask if self.miso is not None and self.miso.value else 0
[... other code ...]
    self.clock.on()
    # read bit when clock is on
    if not clock_phase:
            read_word |= read_bit()
    self.clock.off()
    # read bit when clock is off
    if clock_phase:
            read_word |= read_bit()
```
...but that's probably taking things a bit too far for such a "low level" method. (we need the software SPI to be as fast as possible)

waveform80: > The really question is the combination (instruction inside if inside if inside for inside for inside with), for example...

Hmm, I could probably add a link to the [wikipedia SPI article](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus#Example_of_bit-banging_the_master_protocol) which has a lovely little snippet of C demonstrating the SPI master protocol. That was the basis for my implementation (which I then expanded to handle multiple words, clock phases, etc.)

Once I've incorporated lurch's suggestion of simplifying the clock_phase stuff I'd hope it's close enough to that basic implementation that the correspondence was clear, but I'll add a comment with that link to help future readers.

> Good luck trying to understand https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/spi.py#L308 !! :laughing:

Hmmm, that *is* pretty dense (if it was pure Python 3 I could've made it even denser by skipping the set conversions as Python 3's dict keys operate like sets), but it's hardly code golf material ;)

gpiozero/pins/data.py
```diff
@@ -1078,6 +1078,101 @@ def pulled_up(self, function):
...
+                    try:
+                        header = self.headers[header]
+                    except KeyError:
+                        raise PinInvalidPin(
+                            'there is no %s header on this Pi' % header)
+                    try:
+                        function = header.pins[int(pin)].function
+                    except KeyError:
+                        raise PinInvalidPin(
+                            'no such pin %s on header %s' % (pin, header.name))
```
lurch: Took me a while to figure out why the PinInvalidPin exception on line 1160 used `header` but the PinInvalidPin exception on line 1166 used `header.name` - I _eventually_ realised it's because line 1158 does `header = self.headers[header]` :confused:  (i.e. the type of the `header` variable depends on whether that line threw an exception or not! :nauseated_face: )

gpiozero/pins/data.py
```diff
@@ -1078,6 +1078,101 @@ def pulled_up(self, function):
...
+                    if function.startswith('GPIO') and function[4:].isdigit():
+                        return self.to_gpio(int(function[4:]))
+                    else:
+                        raise PinInvalidPin('%s is not a GPIO pin' % spec)
+            elif spec.startswith('BOARD') and spec[5:].isdigit():
+                main_head = ({'P1', 'J8', 'SODIMM'} & set(self.headers)).pop()
```
lurch: Huh? Is this line a clever way of doing:
```python
for header in ('P1', 'J8', 'SODIMM'):
    if header in self.headers:
        main_head = header
        break
```
?

Are we really expecting people to use `BOARD` numbering with the ComputeModules???

PR #473, opened 2016-09-28
Various refactorings of the SPI Interface classes

lurch:
I haven't checked if or how much this conflicts with #468 , so like #465 this should probably be treated as more of a suggestion for now.

I had a look at the SPI classes on the train this evening, and decided to do a bit of refactoring...
- As I mentioned [here](https://github.com/RPi-Distro/python-gpiozero/pull/469#issuecomment-250038658) I think `InputDevice` and `OutputDevice` should be treated as 'abstract', so I replaced all occurrences with `DigitalInputDevice` and `DigitalOutputDevice` instead. IMHO this makes it clearer that these pins will only be used in digital-only mode, and not potentially in analogue or PWM mode.
- As I mentioned [here](https://github.com/RPi-Distro/python-gpiozero/issues/464#issuecomment-249170711) I think the fact that `SPISoftwareInterface` derives from `OutputDevice` is pretty 'yucky', so I changed it to derive from `Device` instead, and added a `self._select` attribute
- I wasn't sure why the code wasn't using the `@property` notation that other parts of GpioZero use (is there a reason @waveform80 ?) so I 'fixed' that. And re-ordered the properties in `SPISoftwareInterface` so that they matched the same order as in `SPIHardwareInterface`
- As both `SPIHardwareInterface` and `SPISoftwareInterface` are basically implementing the same SPI interface, I made an explicit base `SPIInterface` class (a bit like the base `Pin` class) and changed `SPIHardwareInterface` and `SPISoftwareInterface` to derive from that. (which makes [this](https://cloud.githubusercontent.com/assets/476186/18896265/35680334-8518-11e6-8bec-ee95e1aa9911.png) a bit neater)

All tests still pass! :-)  But I'm not that familiar with SPI so it's possible I've broken things (especially with my second change above) in unexpected ways... ?

### DIFF
--- a/gpiozero/spi.py
+++ b/gpiozero/spi.py
@@ -17,12 +17,74 @@ except ImportError:
     SpiDev = None
 
 from .devices import Device, SharedMixin, _PINS, _PINS_LOCK
-from .input_devices import InputDevice
-from .output_devices import OutputDevice
+from .input_devices import DigitalInputDevice
+from .output_devices import DigitalOutputDevice
 from .exc import SPIBadArgs, SPISoftwareFallback, GPIOPinInUse, DeviceClosed
 
 
-class SPIHardwareInterface(Device):
+class SPIInterface(Device):
+    def __init__(self, *args, **kwargs):
+        super(SPIInterface, self).__init__()
+
+    def read(self, n):
+        return self.transfer((0,) * n)
+
+    def write(self, data):
+        return len(self.transfer(data))
+
+    def transfer(self, data):
+        raise NotImplementedError
+
+    @property
+    def clock_mode(self):
+        raise NotImplementedError
+
+    @clock_mode.setter
+    def clock_mode(self, value):
+        raise NotImplementedError
+
+    @property
+    def clock_polarity(self):
+        raise NotImplementedError
+
+    @clock_polarity.setter
+    def clock_polarity(self, value):
+        raise NotImplementedError
+
+    @property
+    def clock_phase(self):
+        raise NotImplementedError
+
+    @clock_phase.setter
+    def clock_phase(self, value):
+        raise NotImplementedError
+
+    @property
+    def lsb_first(self):
+        raise NotImplementedError
+
+    @lsb_first.setter
+    def lsb_first(self, value):
+        raise NotImplementedError
+
+    @property
+    def select_high(self):
+        raise NotImplementedError
+
+    @select_high.setter
+    def select_high(self, value):
+        raise NotImplementedError
+
+    @property
+    def bits_per_word(self):
+        raise NotImplementedError
+
+    @bits_per_word.setter
+    def bits_per_word(self, value):
+        raise NotImplementedError
+
+
+class SPIHardwareInterface(SPIInterface):
     def __init__(self, port, device):
         self._device = None
         super(SPIHardwareInterface, self).__init__()
@@ -64,12 +126,6 @@ class SPIHardwareInterface(Device):
         except DeviceClosed:
             return "hardware SPI closed"
 
-    def read(self, n):
-        return self.transfer((0,) * n)
-
-    def write(self, data):
-        return len(self.transfer(data))
-
     def transfer(self, data):
         """
         Writes data (a list of integer words where each word is assumed to have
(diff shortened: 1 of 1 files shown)


waveform80:
> As I mentioned here I think InputDevice and OutputDevice should be treated as 'abstract', so I replaced all occurrences with DigitalInputDevice and DigitalOutputDevice instead. IMHO this makes it clearer that these pins will only be used in digital-only mode, and not potentially in analogue or PWM mode.

Yeah, that's a good idea.

> As I mentioned here I think the fact that SPISoftwareInterface derives from OutputDevice is pretty 'yucky', so I changed it to derive from Device instead, and added a self._select attribute

Fair enough - I don't particularly mind deriving from OutputDevice (as the select pin is effectively an "output" which activates the connected SPI device), but this is more intuitive.

> I wasn't sure why the code wasn't using the @property notation that other parts of GpioZero use (is there a reason @waveform80 ?) so I 'fixed' that. And re-ordered the properties in SPISoftwareInterface so that they matched the same order as in SPIHardwareInterface

Ahhh ... this is where #468 comes in. So, in that PR just like Pin, SPI becomes an abstract base class. At the moment when a device needs a pin, it asks the current pin factory to construct one and gets back a something which conforms to the Pin interface. In #468 the same thing happens for SPI devices: upon construction they effectively ask the current pin factory "give me a SPI implementation for these parameters" where the parameters are things like port/device/clock_pin/etc. The SPI implementation returned conforms to the SPI interface (abstract class) defined in `pins/__init__`.

For local pin implementations (RPi.GPIO, RPIO, native) what's returned is either a SpiDev based hardware implementation or SPISoftwareInterface (though it's renamed LocalPiSoftwareSPI in #468). For remote pin implementations (pigpio) it's either PiGPIOHardwareSPI or PiGPIOSoftwareSPI which use pigpio's spi and bb_spi interfaces respectively.

So, the reason the property decls use the non-decorator format is, like Pin implementations, they need to be overridden (before #468 they didn't need to be, but I was anticipating that would come). Overriding property declarations is ... tricky ... using the decorator format (more details [here](http://stackoverflow.com/questions/3336767/overriding-inherited-properties-getters-and-setters-in-python) and [here](http://stackoverflow.com/questions/7019643/overriding-properties-in-python)). Using the straight method implementations (no decorators) is generally easier for people wanting to override stuff - means they don't need to learn the quirks of Python's descriptor and class attribute handling.

> As both SPIHardwareInterface and SPISoftwareInterface are basically implementing the same SPI interface, I made an explicit base SPIInterface class (a bit like the base Pin class) and changed SPIHardwareInterface and SPISoftwareInterface to derive from that. (which makes this a bit neater)

Ah, that basically mirrors the abstract SPI class in #468 (declared next to the abstract Pin class in `pins/__init__`).

I think basically we've got similar ideas - SPI does indeed need some cleaning up, but I think #468 handles a fair bit of this, and remote support besides (though it could do with some of your eagle eye too - e.g. select deriving from OutputDevice is still there).

lurch:
> Overriding property declarations is ... tricky ... using the decorator format

Thanks for the links, will read through them later.

> I think basically we've got similar ideas...

Fantastic :-)

>  #468 ... could do with some of your eagle eye too

Yeah, I _am_ planning to have a look at it, just need to find a large chunk of time to devote to it :)  (maybe tomorrow evening)

waveform80:
Oh cripes, I remember this. Right, I *will* review this (because I vaguely recall thinking it was actually a great idea), but I haven't got time tonight (and it needs rebasing anyway)!

waveform80:
So a mere ... *cough* ... four years later I manage to rebase and review this (finally!). There's good stuff here which I've pushed in 7dbb2df76876e8a66c76e57aa62a1fe7a40724c5 and then tweaked a bit further. The move to a common single base for all the SPI interfaces: great stuff - keeping that. The move to separate out the select pin the software implementations: good. The move to the Digital* devices: nice idea but sadly not practicable -- the DigitalInputDevice in particular hijacks the pin's when_changed event for its events, but we want to hook the pin itself (and avoid all the event handling overhead) for SPI decoding so in the end it makes more sense to keep using the lower-level Input/OutputDevice classes.

Anyway, sorry it's taken sooooo long to get back to GPIO-zero! I wanted to get this done first as I'm having a blitz on the low level pin classes in preparation for Ubuntu (and eventually Raspios) moving to using the gpiochip interfaces, and it's better to get the SPI stuff clean before going and writing anything more on top of them. So, many thanks for the work on this, however long it took for me to pull my finger out and get it (mostly) in here :)

PR #486, opened 2016-10-17
Add socket method and improved close method for Energenie

SteveAmor:
Added a method to return the socket number of the object.

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1175,11 +1175,7 @@ class Energenie(SourceMixin, Device):
 
     @property
     def value(self):
-        try:
-            self._check_open()
-            return self._value
-        except DeviceClosed:
-            return "<gpiozero.Energenie object closed>"
+        return self._value
 
     @value.setter
     def value(self, value):


gpiozero/boards.py
```diff
@@ -1156,6 +1156,8 @@ def close(self):
         if self._master:
             m = self._master
             self._master = None
+            self._socket = None
+            self._value = None
```
bennuttall: Why this line?

bennuttall:
Looks ok to me

lurch:
Looks good to me.

gpiozero/boards.py
SteveAmor: Without these, after you close a socket, methods socket and value still return a valid number.. Didn't seem right to me.

lurch: Hmmm, perhaps the `value` property (and also `socket`?) should call `_check_open` ? https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/devices.py#L217
(just like `__repr__` does)

lurch: ... so looking at https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/input_devices.py#L200 perhaps `self._socket` *shouldn't* be set to `None`, since I guess there's no harm in knowing what socket it *used* to be connected to? Whereas trying to query or set the value should definitely raise `DeviceClosed`.

waveform80:
First commit looks fine - would be happy to merge that, but I'm afraid I have to reject the second: closed objects should blow up with an exception when operations are attempted against them (not return a string - which I'm guessing is an attempt to be friendly at the REPL but which would be rather confusing within a script).

gpiozero/boards.py
bennuttall: @waveform80 see this

SteveAmor:
@waveform80 I've replaced this PR with #629 as this PR has conflicts that I couldn't fix and I'd deleted my fork so the PR is from an unknown repository causing further difficulties.

Issue #485, opened 2016-10-17
7 seg displays


martinohanlon:
Im going to be using a 7 seg display in an upcoming project, so I thought it was time it was added to gpiozero, can I get some comments on my thinking so far:

- class name of 7SegmentDisplay
- extending LEDCollection
- overwriting value to take natural values of numbers and letters to display on the 7 seg display i.e. 7seg.value = "3" would display a 3, likewise 7seg.value = "a" would display A.

Next step would be to create a multi digit 7 segment display. The way these work (generally) is that in addition to the 8 pins for the display they have additional pins to state which digit should be turned on. To make each digit show a different number you would use plexing to flash individual digits to give the appearance of  a multi digit display.

lurch:
See also #357 
I haven't got round to looking through the sample code posted there myself yet.

martinohanlon:
Thanks, I did a search for 'seg' on the issues too and didnt find any.

bennuttall:
`7SegmentDisplay` is not a valid class name in Python, and `7seg` is not a valid object name :P

I'm not sure about `seg.value = "a"` because `.value` should be boolean or float to be compatible with the rest of the library.

Other than that, sounds good. Do various 7seg displays tend to work the same way?

lurch:
> I'm not sure about seg.value = "a" because .value should be boolean or float to be compatible with the rest of the library.

Yeah, that was my gut reaction too - glad to see I wasn't alone ;-)
But it then begs the question, how _do_ we map the LEDs illuminated on a 7-segment display (or 8-segment if it also contains a decimal-point) to a single `.value` between 0 and 1 (or -1 and +1)? There doesn't seem to be any obvious solution...
The only thing I can think of is to treat the 8-segments as elements of a bit-field, and than divide by 256.0 to get a floating-point value (so all segments off would be `0.0`; all segments on would be `1.0`; and segments 'a' and 'c' turned on would be `((2**0 + 2**2) / 256.0) = ((1 + 4) / 256.0) = 0.01953125`). So we could then do `seg2.source = seg1.values`, and it'd work as expected, provided seg1 and seg2 were initialised with their segments in the same order.

But to maintain ease-of-use, we'd also want some kind of `.character` property, where you could do `seg.character = 3` (with the `int` automatically being converted to `str`) or `seg.character = "A"`.
Or perhaps `seg.number = 3` and `seg.hexnumber = 0xa`? \*shrug\*

[...]

PR #488, opened 2016-10-18
Seven segment display

martinohanlon:
Initial PR for seven segment display

Related to issue #485


gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+    def display(self, char):
+        char = str(char).upper()
+        if len(char) > 1:
+            raise ValueError('Only a single character can be displayed')
+        if char in SevenSegmentDisplay.layouts:
+            layout = SevenSegmentDisplay.layouts[char]
+            for led in range(7):
+                self[led].value = layout[led]
+        else:
+            raise ValueError('There is no layout for character %s')
```
lurch: %s doesn't get substituted (whoops)

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+        'Y': (False, True, True, True, False, True, True),
+        'Z': (True, True, False, True, True, False, True),
+        '-': (False, False, False, False, False, False, True),
+        ' ': (False, False, False, False, False, False, False)
+        }
+    
+    def __init__(self, *pins, **kwargs):
+        # 7 segment displays must have 7 or 8 pins
+        if len(pins) < 7 or len(pins) > 8:
+            raise ValueError('7 segment display must have 7 or 8 pins')
```
lurch: Perhaps change `7 segment display` to `SevenSegmentDisplay` ?

lurch:
Ping @SrMouraSilva in case he has any comments.

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+        # Don't allow 7 segments to contain collections
+        for pin in pins:
+            assert not isinstance(pin, LEDCollection)
+        pwm = kwargs.pop('pwm', False)
+        active_high = kwargs.pop('active_high', True)
+        if kwargs:
```
lurch: In common with LEDBoard, should we also have a boolean `initial_value` property?

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+    def display_hex(self, hexnumber):
+        self.display(hex(hexnumber)[2:])
+
+    @property
+    def decimal_point(self):
+        # does the 7seg display have a decimal point (i.e pin 8)
+        if len(self) > 7:
+            return self[7].value 
+        else:
+            raise OutputDeviceError('There is no 8th pin for the decimal point')
```
lurch: I can't decide whether trying to get the decimal point on a 7-pin SevenSegment should just always return `False`, rather than raising an exception?

(OTOH, trying to set the dp on a 7-pin SevenSegment probably _should_ still raise an exception?)

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+        'M': (True, False, True, False, True, False, False),
+        'N': (True, True, True, False, True, True, False),
+        'O': (True, True, True, True, True, True, False),
+        'P': (True, True, False, False, True, True, True),
+        'Q': (True, True, False, True, False, True, True),
+        'R': (True, True, False, False, True, True, False),
+        'S': (True, False, True, True, False, True, True),
+        'T': (False, False, False, True, True, True, True),
+        'U': (False, True, True, True, True, True, False),
+        'V': (False, True, True, True, True, True, False),
```
lurch: 'U' and 'V' are also indistinguishable.

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+        'X': (False, True, True, False, True, True, True),
+        'Y': (False, True, True, True, False, True, True),
+        'Z': (True, True, False, True, True, False, True),
+        '-': (False, False, False, False, False, False, True),
+        ' ': (False, False, False, False, False, False, False)
+        }
+    
+    def __init__(self, *pins, **kwargs):
+        # 7 segment displays must have 7 or 8 pins
+        if len(pins) < 7 or len(pins) > 8:
```
SrMouraSilva: ``if 6 < len(a) < 8:``

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
     def off(self):
         self.value = False
 
+class SevenSegmentDisplay(LEDBoard):
```
SrMouraSilva: if I want to know if it is on or off, how do?

I suggest a status attribute for it

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+        if kwargs:
+            raise TypeError('unexpected keyword argument: %s' % kwargs.popitem()[0])
+        super(SevenSegmentDisplay, self).__init__(*pins, pwm=pwm, active_high=active_high)
+    
+    def display(self, char):
+        char = str(char).upper()
+        if len(char) > 1:
+            raise ValueError('Only a single character can be displayed')
+        if char in SevenSegmentDisplay.layouts:
+            layout = SevenSegmentDisplay.layouts[char]
```
SrMouraSilva: I prefer that the denial be up to reduce the complexity idiomatic

```python
if char not in SevenSegmentDisplay.layouts:
    raise ValueError('There is no layout for character %s')

layout = SevenSegmentDisplay.layouts[char]
for led in range(7):
    self[led].value = layout[led]
```

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
     def off(self):
         self.value = False
 
+class SevenSegmentDisplay(LEDBoard):
+    layouts = {
```
SrMouraSilva: If the layout contemplate that both possibilities, I am in favor of a way to get (@property only) the string representation of what is being displayed: char + dp

gpiozero/boards.py
SrMouraSilva: A custom __repr__ is a good idea

gpiozero/boards.py
```diff
@@ -1185,3 +1185,86 @@ def on(self):
...
+        'F': (True, False, False, False, True, True, True),
+        'G': (True, False, True, True, True, True, False),
+        'H': (False, True, True, False, True, True, True),
+        'I': (False, False, False, False, True, True, False),
+        'J': (False, True, True, True, True, False, False),
+        'K': (True, False, True, False, True, True, True),
+        'L': (False, False, False, True, True, True, False),
+        'M': (True, False, True, False, True, False, False),
+        'N': (True, True, True, False, True, True, False),
+        'O': (True, True, True, True, True, True, False),
```
lurch: See, that's what I mean about different people wanting to map different characters to different 7-seg combinations... ;-)

gpiozero/boards.py
lurch: > if I want to know if it is on or off, how do?

`is_active` will tell you if any LEDs are lit https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/devices.py#L357
(unfortunately that's not in the [docs](http://gpiozero.readthedocs.io/en/latest/api_boards.html#compositedevice) yet)

> A custom `__repr__` is a good idea

What do you suggest? Is the `__repr__` provided by [CompositeDevice](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/devices.py#L315) not good enough?

gpiozero/boards.py
lurch: ...and what should it do if the 7-seg is displaying a combination that doesn't map to a character?

gpiozero/boards.py
lurch: I suspect you might have meant `if not 6 < len(pins) < 9:` ? IMHO that's less readable though.

SrMouraSilva: It's true. I'm sorry

gpiozero/boards.py
SrMouraSilva: > What do you suggest? Is the __repr__ provided by CompositeDevice not good enough?
I was thinking also display the character represented

gpiozero/boards.py
lurch: Being (ultimately) based on `CompositeOutputDevice`, the user is free to set the `.value` property (tuple of bools) to whatever combination they see fit - we don't restrict what combinations they can use. And as mentioned before, I'm not sure that having a property-getter that throws Exceptions is a good idea :-/

gpiozero/boards.py
lurch: > I was thinking also display the character represented

If that _was_ to be added, wouldn't that be a better fit for `__str__` rather than `__repr__` ? \*shrug\*

gpiozero/boards.py
martinohanlon: duh

gpiozero/boards.py
martinohanlon: Everyone's a critic...

I'll add a method to `set_char_layout(char, layout)`

gpiozero/boards.py
martinohanlon: Yep

gpiozero/boards.py
martinohanlon: I think its more readable as is.

gpiozero/boards.py
martinohanlon: yep - made u a 'little u' `'U': (False, False, True, True, True, False, False),`

gpiozero/boards.py
martinohanlon: Yep, I like that

martinohanlon:
Updated PR based on comments above.

gpiozero/boards.py
```diff
@@ -1185,3 +1185,97 @@ def on(self):
...
+            raise ValueError('SevenSegmentDisplay must have 7 or 8 pins')
+        # Don't allow 7 segments to contain collections
+        for pin in pins:
+            assert not isinstance(pin, LEDCollection)
+        pwm = kwargs.pop('pwm', False)
+        active_high = kwargs.pop('active_high', True)
+        initial_value = kwargs.pop('initial_value', False)
+        if kwargs:
+            raise TypeError('unexpected keyword argument: %s' % kwargs.popitem()[0])
+
```
lurch: Might be worth adding a code-comment in here that the LED segments in the `_layouts` array are listed in the order (a, b, c, d, e, f, g) ?

gpiozero/boards.py
```diff
@@ -1185,3 +1185,97 @@ def on(self):
...
+            raise OutputDeviceError('there is no 8th pin for the decimal point')
+    
+    @decimal_point.setter
+    def decimal_point(self, value):
+        if len(self) > 7:
+            self[7].value = value
+        else:
+            raise OutputDeviceError('there is no 8th pin for the decimal point')
+            
+    def set_char_layout(self, char, layout):
```
lurch: I guess a docstring explaining the `char` and `layout` parameters would be really useful? :)

gpiozero/boards.py
```diff
@@ -1185,3 +1185,97 @@ def on(self):
...
+        char = str(char)
```
lurch: Hmm, this should probably be `char = str(char).upper()` to match the `display` function?

lurch:
Yep, looking nice :-)
Could you undo the file-mode changes though? (I think this sometimes happens by accident when you use git on Windows)

Would it be worth adding an `allow_override=False` parameter to `set_char_layout`, such that existing characters in _layouts can only be replaced when allow_override is True?

What does @bennuttall think of my suggestion of the decimal_point property-getter always returning False for 7-segments that don't have a decimal point LED?

gpiozero/boards.py
```diff
@@ -1260,11 +1263,19 @@ def decimal_point(self):
...
     @decimal_point.setter
     def decimal_point(self, value):
         if len(self) > 7:
             self[7].value = value
         else:
-            raise OutputDeviceError('There is no 8th pin for the decimal point')
\ No newline at end of file
+            raise OutputDeviceError('there is no 8th pin for the decimal point')
+            
+    def set_char_layout(self, char, layout):
```
SrMouraSilva: 1. A setter method is pythonic?
2. For me the `set_char_layout` gives the impression that I **can only change** an existing layout.
2. If the layout is attribute (`self._layout`) not a static method, if I change a char layout of a 7seg display, then is desirable that another 7seg haven't this layout?

martinohanlon: I purposely made `._layouts' private to the instance rather than static as I felt it might be confusing if a change was reflected across multi instances.

gpiozero/boards.py
martinohanlon: I've added docstrings to the class now as we seem to be reaching a concensus.

martinohanlon:
Updated PR based on comments & added docs

gpiozero/boards.py
```diff
@@ -1186,7 +1186,45 @@ def off(self):
...
+    and an optional 8th pin for a decimal point. 7 segment displays 
+    typically have either a common anode or common cathode pin, when
+    using a common anode display 'active_high' should be set to False.
+    Instances of this class can be used to display characters or control 
+    individual leds on the display. For example::
+
+        from gpiozero import SevenSegmentDisplay
+
+        seven = SevenSegmentDisplay(1,2,3,4,5,6,7,8,active_high=False)
+        seven.display("7")
```
lurch: 'seven display seven' sounds a bit tautological ;-)
Perhaps `seven.display("5")`

And GPIO1 is one of the 'reserved' pins, so maybe pins 2-9 would be better? (only a minor point)

gpiozero/boards.py
```diff
@@ -1267,13 +1320,26 @@ def decimal_point(self):
     
     @decimal_point.setter
     def decimal_point(self, value):
+        """
+        Sets the status of the decimal point led
+        """
```
lurch: I don't *think* property-setters need their own docstring?

lurch:
Great docs, thanks :-)

gpiozero/boards.py
```diff
@@ -1185,3 +1185,163 @@ def on(self):
...
+        If ``False`` (the default), all LEDs will be off initially. If
+        ``None``, each device will be left in whatever state the pin is found
+        in when configured for output (warning: this can be on). If ``True``,
+        the device will be switched on initially. This parameter can only be
+        specified as a keyword parameter.    
+    
+    """
+    def __init__(self, *pins, **kwargs):
+        # 7 segment displays must have 7 or 8 pins
+        if len(pins) < 7 or len(pins) > 8:
```
SrMouraSilva: The commentary is redundant (see your documentation and the `if len(pins) < 7 or len(pins) > 8:`)

gpiozero/boards.py
```diff
@@ -1185,3 +1185,163 @@ def on(self):
...
+            raise ValueError('SevenSegmentDisplay must have 7 or 8 pins')
+        # Don't allow 7 segments to contain collections
+        for pin in pins:
```
SrMouraSilva: For me, the commentary is redundant

gpiozero/boards.py
```diff
@@ -1185,3 +1185,163 @@ def on(self):
...
+        """
+        self.display(hex(hexnumber)[2:])
+
+    @property
+    def decimal_point(self):
+        """
+        Represents the status of the decimal point led
+        """
+        # does the 7seg display have a decimal point (i.e pin 8)
+        if len(self) > 7:
```
SrMouraSilva: This commentary is necessary?

SrMouraSilva:
It is being planned you perform some test?

martinohanlon:
I've added tests.

tests/test_boards.py
```diff
@@ -741,3 +741,162 @@ def test_energenie():
...
+
+def test_seven_segment_display():
+    pin1 = MockPin(2)
+    pin2 = MockPin(3)
+    pin3 = MockPin(4)
+    pin4 = MockPin(5)
+    pin5 = MockPin(6)
+    pin6 = MockPin(7)
+    pin7 = MockPin(8)
+    pin8 = MockPin(9)
```
lurch: I wonder if it'd be worth renaming these to `pinA`, `pinB`, ... `pinG`, `pinDP` ?

tests/test_boards.py
```diff
@@ -741,3 +741,162 @@ def test_energenie():
...
+    with SevenSegmentDisplay(pin1, pin2, pin3, pin4, pin5, pin6, pin7, pin8, initial_value=True) as seven_seg:
+        assert (seven_seg[0].value and seven_seg[1].value and seven_seg[2].value and seven_seg[3].value and seven_seg[4].value and seven_seg[5].value and seven_seg[6].value)
```
lurch: This should probably test `seven_seg[7].value` too?

tests/test_boards.py
```diff
@@ -741,3 +741,162 @@ def test_energenie():
...
+        assert isinstance(seven_seg[7], LED)
+        assert seven_seg.active_high
+        assert seven_seg[0].active_high
+        assert seven_seg[1].active_high
+        assert seven_seg[2].active_high
+        assert seven_seg[3].active_high
+        assert seven_seg[4].active_high
+        assert seven_seg[5].active_high
+        assert seven_seg[6].active_high
+        assert seven_seg[7].active_high
```
martinohanlon: My thinking exactly, it provides a level of protection against future change and it's consistent with other tests such as LEDBarGraph.

tests/test_boards.py
martinohanlon: Good spot.

lurch:
Great set of tests, thanks Martin! :+1:  Probably worth adding a `test_seven_segment_display_no_decimal` test too?
And could you revert your changes to `debian/rules` please.
As you've seen, testing the error-cases is just as important as testing the working-cases :)

> Any advice on locally testing tests, so I dont have to commit changes to get the tests to run.

You need to setup a virtualenv, activate the virtualenv, install python-gpiozero into the venv with `python setup.py install`, then run `make develop`, and then finally you can run `make test` to run all the unit-tests.
I like to have separate venvs for Python2 and Python3 (with `make develop` run in each) so that I can check that all my tests still pass on both 'old' and 'new' Python :)
@waveform80 's said that at some point he plans to add a page like http://picamera.readthedocs.io/en/latest/development.html to the GpioZero docs.


Hmm, I've just had a crazy idea - what about modifying `display_char` to something like:
```python
   def display(self, char, brightness=1.0):
        """
        Display a character on the 7 segment display.

        :param string char:
            A single character to be displayed.

        :param string brightness:
            The brightness to display the character at, if the display was initialised with ``pwm=True``.
        """
        if isinstance(seven_seg[0], LED):
            if brightness != 1.0:
                raise ValueError('brightness must be 1.0 with non-PWM LEDs')
        char = str(char).upper()
        if len(char) > 1:
            raise ValueError('only a single character can be displayed')
        if char not in self._layouts:
            raise ValueError('there is no layout for character - %s' % char)
        layout = self._layouts[char]
        for led in range(7):
            self[led].value = brightness if layout[led] else 0
```
;-)

tests/test_boards.py
```diff
@@ -741,3 +741,162 @@ def test_energenie():
...
+
+        seven_seg.display("8")
+        assert (seven_seg[0].value and seven_seg[1].value and seven_seg[2].value and seven_seg[3].value and seven_seg[4].value and seven_seg[5].value and seven_seg[6].value)
+
+        seven_seg.display_hex(15)
+        assert (seven_seg[0].value and not seven_seg[1].value and not seven_seg[2].value and not seven_seg[3].value and seven_seg[4].value and seven_seg[5].value and seven_seg[6].value)
```
lurch: I thought I'd already made this comment, but I can't see it, so perhaps it went missing...
(apologies if it comes up as a duplicate)

Perhaps these assert-value lines would be better as e.g.
```python
    assert seven_seg.value[0:6] == (True, False, False, False, True, True, True)
```
and / or:
```python
    expected = (True, False, False, False, True, True, True)
    for i in range(7):
        assert seven_seg[i].value == expected[i]
```

martinohanlon: 'And could you revert your changes to debian/rules please.

No idea how that change got into the commit - any advice on how reverse it? So far I've managed to just totally mess up my local repo!

lurch:
BTW based on what @waveform80 is doing over in #459 , you might need to modify your tests to not use pins 2 and 3.

tests/test_boards.py
martinohanlon: Ta - the simplest method is often the easiest...

lurch:
The 'bad' tests still use MockPin(2) and MockPin(3) ?

tests/test_boards.py
```diff
@@ -836,7 +836,7 @@ def test_seven_segment_initial_value():
     pin7 = MockPin(8)
     pin8 = MockPin(9)
     with SevenSegmentDisplay(pin1, pin2, pin3, pin4, pin5, pin6, pin7, pin8, initial_value=True) as seven_seg:
-        assert (seven_seg[0].value and seven_seg[1].value and seven_seg[2].value and seven_seg[3].value and seven_seg[4].value and seven_seg[5].value and seven_seg[6].value)
+        assert (seven_seg[0].value and seven_seg[1].value and seven_seg[2].value and seven_seg[3].value and seven_seg[4].value and seven_seg[5].value and seven_seg[6].valueand seven_seg[7].value)
```
SrMouraSilva: Is really `valueand` (in end of line)?

martinohanlon:
Arggghhh, unsaved changes in the editor. I really need to sort out a local dev environment. Rather than having to commit each time to test.

tests/test_boards.py
martinohanlon: Resolved

martinohanlon:
Ive added MultiSevenSegmentDisplay to this PR. Additional tests are needed, which I will do soon, but I wanted to get this in.

gpiozero/boards.py
```diff
@@ -1345,4 +1358,204 @@ def set_char_layout(self, char, layout):
...
+        if self.has_decimal_point:
+            output = ""
+            for i in range(len(message)):
+                if message[i] == '.' and (message[i-1] == '.' or i == 0):
+                    output += ' '
+                output += message[i]        
+            message = output
+
+        #validate the messages length
+        if (self.has_decimal_point and len(message) - message.count('.') > len(self.digits)) or (not self.has_decimal_point and len(message) > len(self.digits)):
```
SrMouraSilva: What are the practical effects of putting a dot if the display does not have support?
If 'ignore' or 'error', I suggest changing it to:

```python
text_size = len(message) - message.count('.')
if (text_size > len(self.digits)):
    ...
```

else

```python
text_size = len(message) - message.count('.') if self.has_decimal_point else len(message)
if (text_size > len(self.digits)):
    ...
```

martinohanlon:
Resolved test issues with pypy... Threading - bah.

gpiozero/boards.py
martinohanlon: If the display doesnt support a decimal point, it full stop would be considered to be a character in the message. This would generate an error unless the user had created a character layout for "." using `set_char_layout`.

SrMouraSilva:
Hello. Any news?

waveform80:
Love the idea, and most of the PR (especially that there's a layouts dict which is exactly what I'd want to see). Needs a little love before merge though, and I need to ponder whether this is an example of a class where it'd be acceptable to have a string-based "value" (esp. for the multi-segment one). Scheduling for 1.5.

gpiozero/boards.py
```diff
@@ -1185,3 +1187,375 @@ def on(self):
...
+    digit. 
+    
+    Different values are displayed on different digits by multiplexing the display.
+    
+    Instances of this class can be used to display numbers, messages or control 
+    individual leds and digits on the display. For example::
+
+        from gpiozero import MultiSevenSegmentDisplay
+
+        multi_seven = MultiSevenSegmentDisplay((1,2,3,4,5,6,7,8)(9,10,11,12))
```
lurch: Seems to be a missing comma between the two tuples

gpiozero/boards.py
```diff
@@ -1185,3 +1187,375 @@ def on(self):
...
+        Pins should be in the LED segment order A,B,C,D,E,F,G,decimal_point 
+        (the decimal_point is optional).
+
+    :param bool pwm:
+        If ``True``, construct :class:`PWMLED` instances for each pin. If
+        ``False`` (the default), construct regular :class:`LED` instances. This
+        parameter can only be specified as a keyword parameter.
+
+    :param bool active_high:
+        If ``True`` (the default), the :meth:`on` method will set all the
```
lurch: Perhaps the documentation here should talk about common-anode and common-cathode, rather than talking about HIGH and LOW?

gpiozero/boards.py
bennuttall: I'd use `if len(pins) not in (7, 8)`

waveform80:
Right, erm, four years later (sorry!) I've finally gotten around to doing something about this! I rebased the original PR on the current master and had a play around with it, and came back to my original comment:

> Love the idea, and most of the PR (especially that there's a layouts dict which is exactly what I'd want to see). Needs a little love before merge though, and I need to ponder whether this is an example of a class where it'd be acceptable to have a string-based "value" (esp. for the multi-segment one). Scheduling for 1.5.

I was still of the opinion that this is one of those cases where not having a numeric value is justified. After all, composite boards don't have numeric values (okay, they've got a *tuple* of numeric values, but it's still not strictly a number), and a 1-char value in this case can be seen as a reasonable representation of the device's state (just as the single number in the case of the other LEDCollection derivative, LEDBarGraph, is a reasonable representation of its state).

Anyway, as a result I've done quite a bit of re-working on this and it'll need a separate PR for review so I'll close this and anyone interested can go have a look at #930 which is the continuation.

PR #494, opened 2016-10-28
Fix AttributeError exceptions when Exceptions are raised during __init__

lurch:
Prevents the `Exception AttributeError: "'XXXXXXXX' object has no attribute
'YYYYYYYY'" in  ignored` messages which sometimes pop up when running the
unit tests.


lurch:
Superseded by #507

PR #501, opened 2016-11-25
Add StatusBoard

bennuttall:
**Not ready to merge - awaiting confirmation of final pin numbers.**

STATUS Board is a new board from The Pi Hut, designed by @RZRZR:

https://twitter.com/RachelRayns/status/801873111675183104
https://twitter.com/RachelRayns/status/801874146342895616

It's four sets of traffic light LEDs. I decided to make the `StatusBoard` class refer to each traffic lights set, rather than one class for the whole board, so that you can name each object something context-specific, like with `Energenie`.

It just extends `TrafficLights` and the init allows an `index` (1, 2, 3 or 4) for the strip number, and optionally, `pwm` and `initial_value`.

@waveform80 @lurch I *think* I've done everything needed: docstring, added to `__init__.py`, added to docs page. Tests passing. Look ok?

Note: The URL is not live yet, but it's printed on the PCB.
Note: Some of the LEDs in the pictures/video in the tweets are blue. They're going to be red/amber/green.

Some examples in [this gist](https://gist.github.com/bennuttall/f50201405f521ce807a75b39d7759b04)


bennuttall:
I guess I should add some tests...

bennuttall:
Oh, and the diagrams and stuff. Must look at those...

lurch:
I'd be tempted to name the class `StatusStrip` (or `StatusBoardStrip`) as each object refers to just one strip, rather than the whole board. I guess in effect this is very similar to what I did in #441

bennuttall:
> I'd be tempted to name the class `StatusStrip` (or `StatusBoardStrip`) as each object refers to just one strip, rather than the whole board

I know that makes sense but it's not a very good name. I mean `Energenie` isn't called `EnergenieSocket`. You're essentially creating an *instance* of an `Energenie`, on socket 1, the same way you create an instance of an `LED` on pin 17.

> I guess in effect this is very similar to what I did in #441

I don't see the similarity?

@waveform80 any thoughts?

lurch:
> I know that makes sense but it's not a very good name. I mean Energenie isn't called EnergenieSocket. 

Yeah, you're probably right, especially with the way that ADC boards work currently (which we've discussed before).
I guess in an ideal world, `Energenie` and `StatusBoard` would be factory classes, which return `EnergenieSocket` and `StatusBoardStrip` objects respectively.

> I don't see the similarity?

Really?? `PiStop` descends from `TrafficLights` and has a `location` init-parameter. `StatusBoard` descends from `TrafficLights` and has an `index` init-parameter.

BTW your `StatusBoard` isn't currently using the `initial_value` parameter.

Perhaps it's just me, but IMHO your examples would be more readable if you re-ordered the lines like this:
```python
        wifi.green.source = google.values
        wifi.green.source_delay = 60
        wifi.red.source = negated(wifi.green.values)
```
(IMHO your current example looks kinda weird: setting red.source to green.values, when green.source itself hasn't even been set yet.)

bennuttall:
> BTW your `StatusBoard` isn't currently using the initial_value parameter.

Thanks. Fixed.

> Perhaps it's just me, but IMHO your examples would be more readable if...

You're probably right. Fixed.

lurch:
Just occurred to me that an alternative (better?) approach might be to have `StatusBoard` as a `CompositeOutputDevice` of `TrafficLights`s? And then you could do `status.line1.red.on()` or `status.line4.green.off()`. And I guess you could even add a https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/boards.py#L238 property.

bennuttall:
That was my first approach. It seemed to make sense at first but then controlling individual LEDs was longer than it needed to be, and I couldn't see a use case for wanting to iterate over the lot of them, really.

I ended up finding myself doing this:

```python
sb = StatusBoard()
wifi = StatusBoard[0]
```

and thought the idea of naming the individual strips was better than naming the board, like with Energenie. because really the strips are unrelated to each other.

lurch:
I assume you meant:
```python
sb = StatusBoard()
wifi = sb[0]
```
?

IMHO if you're using a StatusBoard, rather than just a regular TrafficLight board, chances are you're going to _want_ to use more than on strip at a time?

Also, if you want to PWM the LEDs, it's probably simpler to just pass a single `pwm=True` into a CompositeOutputDevice-style StatusBoard, than it is to pass `pwm=True` to each of 4 separate TrafficLights-style StatusBoards?

bennuttall:
Alternative:

```python
class StatusBoard(CompositeOutputDevice):                                
    def __init__(self, *names, **kwargs):
        pwm = kwargs.pop('pwm', False)
        if 1 > len(names) > 4:
            raise ValueError
        strips = OrderedDict()
        pins = (
            (4, 17, 27),
            (22, 10, 9),
            (11, 5, 6),
            (13, 19, 26),
            )
        for index, name in enumerate(names):
            strips[name] = TrafficLights(*pins[index], pwm=pwm)
        super(StatusBoard, self).__init__(_order=strips.keys(), **strips)
```

(a work on progress @lurch and I put together)

bennuttall:
Update on this: there will now be two different status boards: Status and StatusZero. Status will have five strips, StatusZero will have three. The big difference from earlier concepts is that they're now just two-colour strips, rather than three. They're now green/red instead of green/amber/red.

I've thrown together a potential implementation, based on the above example. I think it works quite well.

I use a base class called `StatusBase` (possibly not the right name, but it's internal), which contains the implementation for an arbitrary set of strips, given a collection of pin number pairs. Then `StatusBoard` and `StatusZero` are provided with the appropriate number of pins:

```python
class StatusBase(LEDBoard):
    def __init__(self, pins, *names, **kwargs):
        if len(names) == 0:
            names = ['one', 'two', 'three', 'four', 'five'][:len(pins)]
        elif len(names) > len(pins):
            raise ValueError
        strips = OrderedDict()
        for index, name in enumerate(names):
            green, red = pins[index]
            strips[name] = LEDBoard(green=green, red=red, **kwargs)
        super(StatusBase, self).__init__(_order=strips.keys(), **strips)

class StatusBoard(StatusBase):
    def __init__(self, *names, **kwargs):
        pins = (
            (4, 27),
            (22, 9),
            (11, 6),
            (13, 26),
            (5, 7),
        )
        super(StatusBoard, self).__init__(pins, *names, **kwargs)

class StatusZero(StatusBase):
    def __init__(self, *names, **kwargs):
        pins = (
            (4, 27),
            (22, 9),
            (11, 6),
        )
        super(StatusZero, self).__init__(pins, *names, **kwargs)
```

Usage:

You can name the strips on init, and access them using properties with the same name:

```python
>>> sb = StatusBoard('lan', 'google', 'nas', 'rpi', 'zero')
>>> sb
<gpiozero.StatusBoard object containing 5 devices: lan,google,nas,rpi,zero and 0 unnamed>
>>> list(sb)
[<gpiozero.LEDBoard object containing 2 devices: green,red and 0 unnamed>,
<gpiozero.LEDBoard object containing 2 devices: green,red and 0 unnamed>,
<gpiozero.LEDBoard object containing 2 devices: green,red and 0 unnamed>,
<gpiozero.LEDBoard object containing 2 devices: green,red and 0 unnamed>,
<gpiozero.LEDBoard object containing 2 devices: green,red and 0 unnamed>]
```

If you don't name them all, you only get the ones you named:

```python
>>> sb = StatusBoard('lan', 'google')
>>> sb
<gpiozero.StatusBoard object containing 2 devices: lan,google and 0 unnamed>
```

If created with no parameters, each strip is accessible via a numerically named property:

```python
>>> sb = StatusBoard()
>>> sb
<gpiozero.StatusBoard object containing 5 devices: one,two,three,four,five and 0 unnamed>
>>> sb.one
<gpiozero.LEDBoard object containing 2 devices: green,red and 0 unnamed>
```

```python
>>> sb = StatusZero()
>>> sb
<gpiozero.StatusBoard object containing 3 devices: one,two,three,four,five and 0 unnamed>
>>> sb.one
<gpiozero.LEDBoard object containing 2 devices: green,red and 0 unnamed>
```

PWM and initial_value are possible, the usual way:

```python
>>> sb = StatusZero(initial_value=True, pwm=True)
>>> sb.pulse()
```

Any thoughts?

Note: pin numbers TBC

lurch:
From a cursory glance, it looks fairly reasonable :)
I guess `initial_value=True` will simply turn on all the LEDs for each strip that's been 'named'?

It's been a while since I've written any Python code, so I can't remember the pros/cons of specifically specifying each kwarg param like done [here](https://gpiozero.readthedocs.io/en/stable/api_boards.html#gpiozero.PiLiter) as opposed to just using a generic `**kwargs` as you're doing here?

bennuttall:
> I guess initial_value=True will simply turn on all the LEDs for each strip that's been 'named'?

I believe so

> ...I can't remember the pros/cons of specifically specifying each kwarg param...

I think I did this just out of uncertainty that there were any more besides `pwm` and `initial_value`. Would probably be better for documentation if these were explicit.

lurch:
Out of curiosity, what does the `.values` from the code you've got here look like? A 5-tuple of 2-tuples?

bennuttall:
Yes:

```python
>>> sb.value
StatusBoardValue(one=LEDBoardValue(green=0.0, red=0.0), two=LEDBoardValue(green=0.0, red=0.0), three=LEDBoardValue(green=0.0, red=0.0), four=LEDBoardValue(green=0.0, red=0.0), five=LEDBoardValue(green=0.0, red=0.0))
>>> tuple(sb.value)
(LEDBoardValue(green=0.0, red=0.0),
 LEDBoardValue(green=0.0, red=0.0),
 LEDBoardValue(green=0.0, red=0.0),
 LEDBoardValue(green=0.0, red=0.0),
 LEDBoardValue(green=0.0, red=0.0))
>>> [tuple(v) for v in sb.value]
[(0.0, 0.0), (0.0, 0.0), (0.0, 0.0), (0.0, 0.0), (0.0, 0.0)]
```

lurch:
Cool. Don't forget that you can also do `[led.value for led in sb.leds]` if you _really_ wanted a flat tuple ;-)

bennuttall:
Closing in favour of #558

PR #502, opened 2016-11-25
Add XmasTree Board

bennuttall:
**Not ready to merge - awaiting confirmation of final pin numbers.**

The XMAS board is a new board from The Pi Hut, designed by @RZRZR:

https://twitter.com/RachelRayns/status/801790484205174784
https://twitter.com/RachelRayns/status/801446992828305409
https://twitter.com/RachelRayns/status/801445922651639808

It's a Christmas tree board with 3 red LEDs on each of 8 faces, and a white LED as a star on top.

It just extends `LEDBoard` and the init just has optional `pwm` and `initial_value`:

```python
>>> from gpiozero import XmasTree
>>> tree = XmasTree()
```

On top of the usual on/off/etc. methods, the API provides the following properties:

- `star` - a single `LED` (or `PWMLED`) object representing the star
- `baubles` - a dict of `LED` (or `PWMLED`) objects (all LEDs except the star). It's a dict rather than a tuple so that the indexes are numbered 1:24 rather than 0:23. Final name TBC.

```python
>>> tree.star
<gpiozero.LED object on pin GPIO2, active_high=True, is_active=False>
>>> len(tree.baubles)
24
```

Iterating over `tree` will yield the LEDs from 1:24 and then the star:

```python
>>> tree[:]
(<gpiozero.LED object on pin GPIO4, active_high=True, is_active=False>,
...
<gpiozero.LED object on pin GPIO27, active_high=True, is_active=False>)
>>> tree.baubles
(<gpiozero.LED object on pin GPIO4, active_high=True, is_active=False>,
...
<gpiozero.LED object on pin GPIO2, active_high=True, is_active=False>)
>>> len(tree)
25
>>> len(tree.baubles)
24
```

Indexing `tree` will be inconsistent with indexing `tree.baubles`:

```python
>>> tree[1]
<gpiozero.LED object on pin GPIO5, active_high=True, is_active=False>,
>>> tree.baubles[1]
<gpiozero.LED object on pin GPIO4, active_high=True, is_active=False>,
```

Is it worth investigating a way around this?

I don't think it's worth adding the concept of groups of LEDs for each face.

@waveform80 @lurch I've included a docstring, added to `__init__.py`, added to docs page. Old tests passing, yet to add new ones. Will look at adding to the diagrams too. Look ok so far?

Note: The URL is not live yet, but it's printed on the PCB.
Note: I have yet to find out the order of the pins. It will use all available pins besides GPIO3 (and GPIO0 & GPIO1), so GPIOs 4-27 and GPIO2 (the star). Rachel has labelled them with a number from 1 to 24 at random, so there'll be a mapping of index: pin.


lurch:
> Indexing tree will be inconsistent with indexing tree.baubles:

Why not just move the star to be the first item in the list, rather than the last?

Constructing a dict-comprehension every single time you use the `baubles` property seems horribly inefficient. I don't have time to play with the code right now, but I wonder if it could be made to work like the way SnowPi does, instead of adding your own custom properties?

I guess a natural usage-example would be to light up bauble 1 on 1st Dec, baubles 1->3 on 3rd Dec... etc. and all baubles and the star on Christmas Day (25th Dec) :-)

bennuttall:
> Why not just move the star to be the first item in the list, rather than the last?

Because then the star would come on first when iterating over the tree object.

> I wonder if it could be made to work like the way SnowPi does, instead of adding your own custom properties?

It could be that `self.baubles` is a nested LEDBoard, and `self.star` is a single LED. That might make more sense, except that you couldn't loop over the LEDS in the same way. I think I considered this but it didn't seem worth it.

> Constructing a dict-comprehension every single time you use the baubles property seems horribly inefficient

I guess I could move the dict-comprehension to the init under `self._baubles` and let the `self.baubles` property return that.

> I guess a natural usage-example would be to light up bauble 1 on 1st Dec, baubles 1->3 on 3rd Dec... etc. and all baubles and the star on Christmas Day (25th Dec) :-)

Exactly. That's why the star is last.

@waveform80 any thoughts?

lurch:
> Because then the star would come on first when iterating over the tree object.

Ahh, fair enough.

> It could be that self.baubles is a nested LEDBoard, and self.star is a single LED. That might make more sense, except that you couldn't loop over the LEDS in the same way.

Yes you can, that's exactly what http://gpiozero.readthedocs.io/en/v1.3.1/api_boards.html#gpiozero.LEDBoard.leds is for ;-)

bennuttall:
> Yes you can, that's exactly what http://gpiozero.readthedocs.io/en/v1.3.1/api_boards.html#gpiozero.LEDBoard.leds is for ;-)

So `XmasTree` would extend `CompositeOutputDevice` and comprise a `LED` and `LEDBoard`?

lurch:
Nope, it would still extend `LEDBoard`. Have another look at how https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/boards.py#L827 builds the `eyes` and `nose` parameters.

bennuttall:
@lurch You mean like this?

```python
class XmasTree(LEDBoard):
    def __init__(self, pwm=False, initial_value=False):
        super(XmasTree, self).__init__(
            baubles = LEDBoard(
                4, 15, 13, 21, 22, 6, 12, 25, 16, 17, 27, 26,
                9, 23, 11, 5, 20, 19, 14, 18, 7, 8, 10, 24,
                pwm=pwm, initial_value=initial_value
            ),
            star=2,
            pwm=pwm, initial_value=initial_value,
            _order=('baubles', 'star'),
        )
```

However this means the baubles are now zero-indexed, which is a problem?

If there's a way of 1-indexing the baubles, then I'm happy with this. If not, we might have to stick with the current implementation.

lurch:
Off the top of my head, the best I can think of using the current `LEDBoard` machinery is:
```python
class XmasTree(LEDBoard):
    def __init__(self, pwm=False, initial_value=False):
        bauble_pins = (4, 15, 13, 21, 22, 6, 12, 25, 16, 17, 27, 26, 9, 23, 11, 5, 20, 19, 14, 18, 7, 8, 10, 24)
        bauble_dict = OrderedDict()
        for i, pin in enumerate(bauble_pins):
            bauble_dict['bauble%d' % (i+1)] = pin
        super(XmasTree, self).__init__(
            baubles = LEDBoard(
                pwm=pwm, initial_value=initial_value,
                _order=bauble_dict.keys(),
                **bauble_dict
            ),
            star=2,
            pwm=pwm, initial_value=initial_value,
            _order=('baubles', 'star')
        )
```
(python identifiers have to start with a letter, they can't start with a number)

Which would then allow you to do `tree.baubles.bauble1.on()` (which _I think_ would do the same as `tree.baubles[0].on()` and `tree.leds[0].on()`), and then `tree.star.on()` should do the same as `tree.leds[24].on()`.

(This is all written from memory, so some of the details may be slightly wrong)

martinohanlon:
Should we use a different name to XmasTree ? I know of at least 2 other christmas tree boards.

PiHutXmasTree?

lurch:
Yeah, I was just wondering if `XmasTree` was a bit too generic...

bennuttall:
@lurch I like that implementation, but I'm not sure about `tree.baubles.bauble1`. Maybe choose something shorter like `tree.baubles.b1`? I'm not really sure.

An alternative could be that it's just LEDBoard of 25x LEDs, with no nesting, but use your naming strategy:

```
>>> len(tree)
25
>>> tree.star.on()
>>> tree.bauble1.on()
```

That means you can't do `tree.baubles.on()` but makes individual number-based access easier/shorter.

Just want to pick something and go with it really. Anyone with any suggestions, please reply!

lurch:
If you want to get rid of the baubles / star sub-devices, then I guess another approach might be:
```python
class XmasTree(LEDBoard):
    def __init__(self, pwm=False, initial_value=False):
        leds_dict = OrderedDict()
        leds_dict['star'] = 2
        bauble_pins = (4, 15, 13, 21, 22, 6, 12, 25, 16, 17, 27, 26, 9, 23, 11, 5, 20, 19, 14, 18, 7, 8, 10, 24)
        for i, pin in enumerate(bauble_pins):
            leds_dict['bauble%d' % (i+1)] = pin
        super(XmasTree, self).__init__(
            pwm=pwm, initial_value=initial_value,
            _order=leds_dict.keys(),
            **leds_dict
            )
        )
```

Which would then mean that `tree.bauble1` is the same thing as `tree.leds[1]` (with `tree.star` then being the same thing as `tree.leds[0]`).

bennuttall:
That sounds like what I'm after!

Btw campug tonight! See you there?

waveform80:
My first thought is "why shouldn't the star be first?"

bennuttall:
I think because if you iterated over them all, you'd want the star to come on last (like the final piece of an advent calendar).

bennuttall:
I think this is complete. Just need to add tests. And resolve the merge conflict...

bennuttall:
Nope. Need to use `OrderedDict`.

bennuttall:
I'm going to merge this as soon as I've doubled checked the pin numbers are correct. @lurch are you happy with the implementation?

martinohanlon:
My hairs gone grey since you opened this pull request.

PR #497, opened 2016-11-07
Add pinout.py command-line tool, close #444

stewartadcock:
Add a "pinout" command-line tool which will write information about GPIO pins for the various RPi revisions.

This corresponds to issue #444.

PR #504, opened 2016-12-05
Add pinout cli tool, close #444, close #497

### LINKED ISSUE #444: pinout command line tool
We currently have a `pi_info` function which allows you to find out information on the Pi you're on, or another specified Pi. We want to add a command line tool `pinout` which would print the pinout for your Pi by default, or a specified Pi.

See #354 for info on how to get pinout from pi_info.

An example of argparse: https://github.com/pyjokes/pyjokes/blob/master/scripts/pyjoke

bennuttall:
New PR introducing small changes to @stewartadcock's pinout cli tool PR #497.

@waveform80 I'm about to make a few notes in a code review but can you please check it over? Looks good to me, it works, and tests pass!


docs/cli_tools.rst
```diff
@@ -0,0 +1,26 @@
+==================
+Command-line Tools
+==================
+
+Pinout
+======
+
+The gpiozero package contains a database of information about the various
+revisions of Raspberry Pi. This is queried by the ``pinout`` command-line
+tool to write details of the GPIO pins available.
```
lurch: Perhaps this should use "display" instead of "write"?

lurch:
@bennuttall One of the other things we discussed "fixing" was changing the `exit` calls to use sys.exit, and changing all errors to consistently return an error-code of `1`.
(unless @stewartadcock can explain why he's used `exit(-1)` and `exit(2)` ??)

docs/cli_tools.rst
stewartadcock: My thought is that "display" could be misleading. How about "output"?

Depends on the audience. Thinking of my own children, for end of KS2 "output" seems a good choice, but earlier I'd stick to "write"/"display" or, maybe, "give".

lurch: Yeah, "output" matches the wording used in `pinout.py` itself. Fixed.

lurch:
> I suggest the exit codes are changed as suggested.

Done :)

waveform80:
Okay, been over most of this today and it looks good to me - sorry for the long delay getting back to GPIO Zero, but it looks like everyone's been busy in my absence! Will try to get on with plenty of bits this week :)

lurch:
Nice to "see" you again @waveform80 :smiley:

Issue #511, opened 2017-01-01
controlling ACT LED gives a warning message


TonyLHansen:
creating an LED with the Activity LED or Power LED GPIO generates an erroneous warning message:

```
/usr/lib/python3/dist-packages/gpiozero/pins/rpigpio.py:95: PinNonPhysical: no physical pins exist for GPIO47
   'no physical pins exist for GPIO%d' % number))
/usr/lib/python3/dist-packages/gpiozero/pins/rpigpio.py:95: PinNonPhysical: no physical pins exist for GPIO35
   'no physical pins exist for GPIO%d' % number))
```

On A and B, the status ACT LED is controlled by GPIO 16, active low. On a A+, B+ and beyond, the status ACT LED is controlled by GPIO47, and its actually controlling a FET, so on/off is inverted, with high being ON. In addition, on A+, B+ and beyond, you can also sense the status of the PWR LED, that is the brownout detector driving it, by using GPIO35 to read it out, but if you change GPIO35 to an output you can actually control the PWR LED now too, same way as the ACT LED.

Other than the warning message, using LED(47) and LED(35) controls the ACT and Power LEDs just fine on the newer RPIs, and LED(16) controls the ACT LED just fine on the older RPIs.

IMO, there should be no warning message when these are instantiated. Yes, there is a workaround for the warning message, as in:

```
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    led = LED(blinkpin)
```

but this code should not be necessary.

I have two suggestions: 

* Add some intelligence to gpiozero.pi_info to request the LED GPIO numbers for the ACT and Power LEDs (return None if not available). This allows portable code to be written for both types of RPIs.
* Then use that information within the LED class to allow those pins to be used without generating the warning message.

lurch:
Yeah, see also https://gpiozero.readthedocs.io/en/v1.3.1/recipes.html#controlling-the-pi-s-own-leds and https://github.com/RPi-Distro/python-gpiozero/issues/148#issuecomment-181649790

lurch:
[...]

PR #512, opened 2017-01-03
add GPIOZeroWarning.printWarnings. 

TonyLHansen:
Add `GPIOZeroWarnings.printWarnings` class variable.
Use subclass versions of it before calling `warnings.warn()`.
Add missing exceptions/warnings to docs/api_exc.rst, and document new class variable and show an example of its use.

Fixes issue #511.

### DIFF
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -219,9 +219,10 @@ class NativePin(LocalPin):
             try:
                 cls.PI_INFO.physical_pin('GPIO%d' % number)
             except PinNoPins:
-                warnings.warn(
-                    PinNonPhysical(
-                        'no physical pins exist for GPIO%d' % number))
+                if PinNonPhysical.printWarnings:
+                    warnings.warn(
+                        PinNonPhysical(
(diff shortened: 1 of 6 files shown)


gpiozero/pins/native.py
```diff
@@ -219,9 +219,10 @@ def __new__(cls, number):
             try:
                 cls.PI_INFO.physical_pin('GPIO%d' % number)
             except PinNoPins:
-                warnings.warn(
-                    PinNonPhysical(
-                        'no physical pins exist for GPIO%d' % number))
+                if PinNonPhysical.printWarnings:
+                    warnings.warn(
+                        PinNonPhysical(
+                            'no physical pins exist for GPIO%d' % number))
```
lurch: It's just a gut feeling, but having `PinNonPhysical` mentioned here twice feels like an anti-pattern.
I wonder if it might be better to wrap `warnings.warn` inside another function, and have that function check the `.printWarnings` class attribute of the passed-in warning?

@waveform80 will probably have a better idea about the 'most Pythonic' solution ;)

Issue #518, opened 2017-01-08
python-gpiozero should depend on python-pkg-resources


m-reiter:
Importing Button from gpiozero fails if python-pkg-resources is not installed:

```
    from gpiozero import Button
  File "/usr/lib/python2.7/dist-packages/gpiozero/__init__.py", line 58, in <module>
    from .devices import (
  File "/usr/lib/python2.7/dist-packages/gpiozero/devices.py", line 18, in <module>
    import pkg_resources
ImportError: No module named pkg_resources
```

So I guess a dependency should probably be added.

broo0ose:
I'm getting this as well on python and python3

bennuttall:
You need to install `python-dev` and/or `python3-dev`:

```
sudo apt install python-dev python3-dev
```

lurch:
...or you can just install python-pkg-resources and/or python3-pkg-resources :
```bash
sudo apt-get install python-pkg-resources python3-pkg-resources
```

(see https://github.com/RPi-Distro/python-gpiozero/issues/538#issuecomment-283366919 )

PR #519, opened 2017-01-08
Update debian package description

bennuttall:
All ok with this? Anything else wrong/missing from this file? #518 ?


waveform80:
Oh, and I'm afraid the first line should be one sentence too as it's used in single-line synopsis listings (see [Description](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Description) field in the Debian docs). I'll have a go at re-jigging it in a mo.

PR #532, opened 2017-01-26
Add LoadAverage internal device class

bennuttall:
It's heavily based on `CPUTemperature` so could probably do with refactoring, especially if we want to add more file-based internal devices, but it works!

Any comments?


gpiozero/other_devices.py
```diff
@@ -147,6 +147,92 @@ def is_active(self):
...
+        The load average above which the device will be considered "active".
+        This defaults to 0.8.
+
+    :param int minutes:
+        The number of minutes over which to average the load. Must be 1, 5 or
+        15.
+    """
+    def __init__(self, load_average_file='/proc/loadavg', min_load_average=0.0,
+        max_load_average=1.0, threshold=0.8, minutes=1):
+        self.load_average_file = load_average_file
```
lurch: Perhaps this could check that the file actually exists?

bennuttall:
Example for the docs:

```python
from gpiozero import TrafficLights, LoadAverage
from signal import pause

lights = TrafficLights(9, 10, 11)
las = [LoadAverage(minutes=m) for m in (1, 5, 15)]

for led, la in zip(lights, las):
    led.source = la.values

pause()
```

bennuttall:
Perhaps there should be a `LoadAverages` class which comprises all three values?

gpiozero/other_devices.py
bennuttall: Yes - that would be good. Would also need fixing in CPUTemperature.

gpiozero/other_devices.py
```diff
@@ -147,6 +147,92 @@ def is_active(self):
...
+        super(InternalDevice, self).__init__()
+        self.min_load_average = min_load_average
+        self.max_load_average = max_load_average
+        self.threshold = threshold
```
bennuttall: Yes - that would be good. Would also need fixing in CPUTemperature.

gpiozero/other_devices.py
```diff
@@ -147,6 +147,92 @@ def is_active(self):
...
+        if minutes not in (1, 5, 15):
```
bennuttall: Done

gpiozero/other_devices.py
```diff
@@ -147,6 +147,92 @@ def is_active(self):
...
+            raise ValueError('minutes must be 1, 5 or 15')
+        self._load_average_file_column = {
+            1: 0,
+            5: 1,
+            15: 2,
+        }[minutes]
```
bennuttall: We could - we just tend to go the extra mile with init params to give a better error message.

bennuttall: Sorry, I misunderstood you. I still think I prefer this way, as the try/except makes the error message longer and scarier. Compare: http://imgur.com/a/TFCPP

gpiozero/other_devices.py
```diff
@@ -147,6 +147,92 @@ def is_active(self):
...
+            return float(file_contents[self._load_average_file_column])
+
+    @property
+    def value(self):
+        """
+        Returns the current load average as a value between 0.0 (representing
+        the *min_load_average* value) and 1.0 (representing the
+        *max_load_average* value). These default to 0.0 and 1.0 respectively.
+        """
+        load_average_range = self.max_load_average - self.min_load_average
```
bennuttall: Yup - also needs addressing in CPUTemperature.

gpiozero/other_devices.py
```diff
@@ -147,6 +147,92 @@ def is_active(self):
...
+    def __repr__(self):
+        return '<gpiozero.LoadAverage load average=%.2f>' % self.load_average
+
+    @property
+    def load_average(self):
+        """
+        Returns the current load average
+        """
+        with io.open(self.load_average_file, 'r') as f:
+            file_contents = f.readline().strip().split()
```
bennuttall: Done

gpiozero/other_devices.py
```diff
@@ -147,6 +147,92 @@ def is_active(self):
...
+
+    def __repr__(self):
+        return '<gpiozero.LoadAverage load average=%.2f>' % self.load_average
+
+    @property
+    def load_average(self):
+        """
+        Returns the current load average
+        """
+        with io.open(self.load_average_file, 'r') as f:
```
bennuttall: :)

gpiozero/other_devices.py
waveform80: Ah, Python 3's chained exceptions. Personally I'd still prefer the try/except version (EAFP vs LBYL and all that; not that it matters in this case but the style is just ingrained in me!). It does make the error longer but I wouldn't say *much* scarier. Someone who doesn't understand/read the exceptions will be scared off in either case by all the preamble, and someone who does know to read the last line first gets the message straight away in either case. Incidentally, if you're *really* bothered about the extra chained exception you can just do:

```python
try:
    self._load_average_file_column = { ... }[minutes]
except KeyError:
    raise ValueError('minutes must be 1, 5 or 15') from None
```

The `from None` squishes the exception chain (and is compatible with Python 2.x).

gpiozero/other_devices.py
waveform80: For that matter you can just keep the file open (i.e. open it in the constructor, close it in close), then seek back and re-read in the property. Opening a file is a fairly "hefty" operation (requires a fair bit of work kernel-side) compared to seeks and reads, so I could easily believe hitting the property continually would raise the loadavg ... but then it will even with a seek+read implementation because executing Python code involves plenty of user-space work.

In other words, it won't make much difference to anyone constantly hammering the property, and there's an argument that keeping a file open is a waste of resources anyway. Hence I've no particular opinion on this one: either's fine with me.

gpiozero/other_devices.py
waveform80: Sounds reasonable to me.

waveform80:
Interesting; looks good if the few reviews left already can be dealt with. My one concern at the moment is that the class leaves out a rather interesting bit of info that /proc/loadavg gives you:

If your loadavg is 2 1 1 (i.e. 2 over 1 minute, 1 over 5 minutes, and 1 over 15 minutes) that tells you the load is *rising* (because the short mean is higher than the longer means). Likewise, if loadavg is 1 2 2 that tells you load is *falling* (and in fact is falling within the last 5 minutes). This rising/falling distinction is lost by the class concentrating on a single timing interval. That said, I've no good suggestions how it should be incorporated other than turning it into a class that yields a composite of the three values as a tuple - but that makes it non-trivial to interface to a LED (which is probably a primary use-case). Perhaps something to consider in another class?

bennuttall:
Nice idea. I like the concept of monitoring load rising or falling.

No idea what they'd be called but two classes for this would be good.

PR #536, opened 2017-02-09
Moved pull_up from super to init, close #505

### LINKED ISSUE #505: Problem with MotionSensor()
I am using a PIR sensor and on my RPI is running a webserver. Following code works fine with RPi.GPIO lib. But I want to go away from RPi.GPIO and so I tryed gpiozero. But second code is not working. Why? 
is working:
```python
PIR_PIN = 20
@app.route('/info',  method='POST')
def info_post(db):
    alarm =  int(request.forms.get('checkbox'))
    db.execute('UPDATE alarm SET status=? WHERE id=1',
                       (alarm,))
    GPIO.setmode(GPIO.BCM)
    GPIO.setup(PIR_PIN, GPIO.IN)
    
    if alarm:
        print 'Start PIR-Sensor'
        GPIO.add_event_detect(PIR_PIN, GPIO.RISING, callback=get_motion)
    else:
        print 'Exit'
        GPIO.cleanup()
```
not working:
```python
@app.route('/gpio',  method='POST')
def set_gpio_status():
    pir_status =  int(request.forms.get('pir_status'))
    pir = MotionSensor(20)
    
    if pir_status:
        print('start')
        pir.when_motion = get_motion
    else:
        print('exit')
        pir.close()
```

delciotorres:
Some PIR sensors are not pull_down by default but pull_up.
At `input_devices.py` the `pull_up=False` is hardcoded at `super`.
Moved up to `__init__` so for that case a user could do:

```python
from gpiozero import MotionSensor

pir = MotionSensor(4, pull_up=True)
while True:
    if pir.motion_detected:
        print("Motion detected!")
```

( @lurch edited this comment to add extra code-formatting )

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -394,12 +394,18 @@ class MotionSensor(SmoothedInputDevice):
         :attr:`~SmoothedInputDevice.is_active` until the internal queue has
         filled with values.  Only set this to ``True`` if you require values
         immediately after object construction.
+
+    :param bool pull_up:
+        If ``False`` (the default), the GPIO pin will be pulled low by default.
+        In this case, connect the other side of the button to 3V3. If
+        ``True``, the GPIO pin will be pulled high by default. In this case,
+        connect the other side of the button to ground.
     """
     def __init__(
             self, pin=None, queue_len=1, sample_rate=10, threshold=0.5,
-            partial=False):
+            partial=False, pull_up=False):
         super(MotionSensor, self).__init__(
-            pin, pull_up=False, threshold=threshold,
+            pin, pull_up, threshold=threshold,
             queue_len=queue_len, sample_wait=1 / sample_rate, partial=partial
         )
         try:
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -397,15 +397,15 @@ class MotionSensor(SmoothedInputDevice):
 
     :param bool pull_up:
         If ``False`` (the default), the GPIO pin will be pulled low by default.
-        In this case, connect the other side of the button to 3V3. If
+        In this case, connect the other side of the sensor to 3V3. If
         ``True``, the GPIO pin will be pulled high by default. In this case,
-        connect the other side of the button to ground.
+        connect the other side of the sensor to ground.
     """
     def __init__(
             self, pin=None, queue_len=1, sample_rate=10, threshold=0.5,
             partial=False, pull_up=False):
         super(MotionSensor, self).__init__(
-            pin, pull_up, threshold=threshold,
+            pin, pull_up=pull_up, threshold=threshold,
             queue_len=queue_len, sample_wait=1 / sample_rate, partial=partial
         )
         try:


gpiozero/input_devices.py
```diff
@@ -394,12 +394,18 @@ class MotionSensor(SmoothedInputDevice):
         :attr:`~SmoothedInputDevice.is_active` until the internal queue has
         filled with values.  Only set this to ``True`` if you require values
         immediately after object construction.
+
+    :param bool pull_up:
+        If ``False`` (the default), the GPIO pin will be pulled low by default.
+        In this case, connect the other side of the button to 3V3. If
```
lurch: This docstring obviously shouldn't be talking about 'buttons'.

gpiozero/input_devices.py
```diff
@@ -394,12 +394,18 @@ class MotionSensor(SmoothedInputDevice):
...
+        ``True``, the GPIO pin will be pulled high by default. In this case,
+        connect the other side of the button to ground.
     """
     def __init__(
             self, pin=None, queue_len=1, sample_rate=10, threshold=0.5,
-            partial=False):
+            partial=False, pull_up=False):
         super(MotionSensor, self).__init__(
-            pin, pull_up=False, threshold=threshold,
+            pin, pull_up, threshold=threshold,
```
lurch: This should be `pull_up=pull_up`
[ It looks confusing, but the left-hand `pull_up` is the parameter-name that gets passed into the `super` call (with so many optional parameters, it's always better to pass them by name than by position), and the right-hand `pull_up` is the value of the `pull_up` parameter that got passed into _this_ `__init__` function - which you're defaulting to `False` ]

gpiozero/input_devices.py
delciotorres: right, I will change it to sensors, I copied from the buttons description.

gpiozero/input_devices.py
delciotorres: Yep ok! Will do that. I was following the same standard found at Button class. Thanks!

gpiozero/input_devices.py
```diff
@@ -397,15 +397,15 @@ class MotionSensor(SmoothedInputDevice):
 
     :param bool pull_up:
         If ``False`` (the default), the GPIO pin will be pulled low by default.
-        In this case, connect the other side of the button to 3V3. If
+        In this case, connect the other side of the sensor to 3V3. If
         ``True``, the GPIO pin will be pulled high by default. In this case,
-        connect the other side of the button to ground.
+        connect the other side of the sensor to ground.
```
lurch: Looking at https://github.com/RPi-Distro/python-gpiozero/pull/536/commits/bdb3f9dbc6f32c8d579648a99e25fd052cc9755d#diff-58c739be690512be707a7a9a3e174730L360 IMHO these wiring instructions still don't make sense :-/

waveform80:
I'm fine with adding `pull_up` as a param for MotionSensor - makes good sense. But the doc string needs some tweaking: it's suggesting connecting the "other side" of the sensor but the typical PIR sensors (e.g. from the CamJam kits) have three pins, Vcc, Gnd, and Out so "other side" is a bit confusing. Also it mentions connecting to 3V3 but the CamJam kits are 5V components (hence why the lead paragraph goes on about connecting to the 5V pin).

This is simple enough though that I'm happy to merge and tweak the docs.

Issue #279, opened 2016-04-13
Add-on board use with remote pins?


bennuttall:
It's great that I can great an instance of an LED using a remote pin like so:

``` python
from gpiozero.pins.pigpiod import PiGPIOPin
from gpiozero import LED
from signal import pause

led = LED(PiGPIOPin(12, host='192.168.0.2'))
```

However, I don't believe it's possible to create an instance of, say, a `SnowPi` attached to another Pi, because that doesn't take pin numbers on init, so I can't provide alternative pins.

Would it be possible to achieve this with an init kwarg?

``` python
sp = SnowPi(host='192.168.0.2')
```

or perhaps configure the host at the top of the script, similar to RPi.GPIO's `setmode()`?

I imagine it could be useful to be able to control devices on multiple Pis at once, in the same script, so perhaps the `setmode` style configuration would not work for all cases.

lurch:
I guess this is where `DefaultPin` comes into play? http://gpiozero.readthedocs.org/en/latest/api_pins.html

Perhaps you could make a subclass of `PiGPIOPin` which is hard-coded to use a particular `host` (i.e. your remote Pi) and then set `gpiozero.devices.DefaultPin` to your new subclass? (and you could have multiple subclasses, each using a different IP address; changing the DefaultPin each time you want to address a different Pi).

All in theory, I've not actually tested any of this myself! ;-)

waveform80:
Yup, that's what `DefaultPin` is for, and @lurch is absolutely right that if you wanted a remote pin as the default that would currently mean sub-classing `PiGPIOPin` to change the default host in the constructor (if you want to be strict about it).

As a bit of background, `DefaultPin` is something I'm not terribly happy with: it works, but it's a bit crude and rather difficult to modify. I'd much prefer to replace it with a factory function/class that was called whenever a pin needed to be made with whatever arguments had been given to construct that pin (in fact this is more a documentation change than a code change - very little would change in `devices.py`). In this case instead of sub-classing PiGPIOPin you'd do something more like:

```python
from gpiozero.pins.pigpiod import PiGPIOPin
import gpiozero.devices

def make_remote_pins(number):
    return PiGPIOPin(number, host='my_pi')

gpiozero.devices.DefaultPin = make_remote_pins
```

[...]

PR #547, opened 2017-03-21
Check pin parameter type in GPIODevice class, close #546

### LINKED ISSUE #546: Check "pin" parameter type in GPIODevice class
If you try to create a GPIO device object with a string as the pin number, you get a strange error:

```python
>>> led = LED("2")
[abridged traceback...]
--> 135             self = super(GPIOMeta, cls).__call__(*args, **kwargs)
--> 131         super(DigitalOutputDevice, self).__init__(pin, active_high, initial_value)
---> 47             self.pin.output_with_state(self._value_to_state(initial_value))
AttributeError: 'str' object has no attribute 'output_with_state'
```

First of all, it would be better if the error was caught straight away, not coincidentally that something later tried to operate on a string and failed.

Secondly, this causes an issue when closing the device (e.g. when closing the shell):

```python
_PINS.pop().close()
AttributeError: 'str' object has no attribute 'close'
[snip]
AttributeError: 'LED' object has no attribute '_controller'
```

This is because [`GPIODevice`](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/devices.py#L362) only [checks](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/devices.py#L381) if `pin` is an int, and converts it to a `Pin` object. Either way, it continues from this point assuming `pin` is a `Pin` object, and adds it to the `_PINS` set. Then when it comes to close on exit, it fails to close this pin, because it's a string not a pin.

I will PR a suggested fix but there's every chance I'll trivialise it and it'll need to be done differently.

bennuttall:
Potential fix for #546

### DIFF
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -18,7 +18,7 @@ from threading import RLock
 import pkg_resources
 
 from .threads import _threads_shutdown
-from .pins import _pins_shutdown
+from .pins import Pin, _pins_shutdown
 from .mixins import (
     ValuesMixin,
     SharedMixin,
@@ -380,6 +380,8 @@ class GPIODevice(Device):
             raise GPIOPinMissing('No pin given')
         if isinstance(pin, int):
             pin = pin_factory(pin)
+        if not isinstance(pin, Pin):
+            raise ValueError('pin must be an int or a gpiozero Pin object')
         with _PINS_LOCK:
             if pin in _PINS:
                 raise GPIOPinInUse(


gpiozero/devices.py
```diff
@@ -380,6 +380,8 @@ def __init__(self, pin=None):
             raise GPIOPinMissing('No pin given')
         if isinstance(pin, int):
             pin = pin_factory(pin)
+        if not isinstance(pin, Pin):
```
lurch: @waveform80 's more familiar with the Pin internals than me ;-)
With the code on the current master branch, I think this looks fine but it might need to be enhanced when the pin-factory stuff gets merged (which I need to get back to reviewing at some point, when I'm not so busy!)

Only a minor difference, but I'd be tempted to change `if not isinstance(pin, Pin):` to:
```python
elif isinstance(pin, Pin):
  pass
else:
```

lurch:
To keep codecov happy, you could add a unit-test for this exception being thrown ;-)

waveform80:
Sorry, this is effectively made redundant by the recent fix for #279. Now it's no longer valid to pass a Pin implementation to GPIODevice (or any descendent), just a pin-specification, and what a pin-specification is depends entirely on the default or specified pin-factory. I'll add some further comments on #546 about pin specifications...

bennuttall:
:+1:

PR #558, opened 2017-04-23
Add StatusZero and StatusBoard

bennuttall:
This PR supersedes #501 - third time lucky, here's an implementation for the new Status Board, and Status Zero.

The original prototypes were LED-only. Now, the full-sized Status Board comes with a button for each strip, as well as red and green LEDs. There are five button-led strips on the Status Board and just three led-only strips on the Status Zero.

I implemented `StatusZero` similarly to the examples in the comments in #501: as a `LEDBoard` derivative nesting three `LEDBoard`s. Its value looks like:

```python
StatusZeroValue(
    one=LEDBoardValue(red=False, green=False),
    two=LEDBoardValue(red=False, green=False),
    three=LEDBoardValue(red=False, green=False)
)
```

*(corrected red/green order)*

I've implemented `StatusBoard` as a `CompositeOutputDevice` derivative, nesting five `CompositeOutputDevice`s, each comprising a `LEDBoard` and a `Button`. Its value looks like:

```python
StatusBoardValue(
    one=CompositeOutputDeviceValue(
        button=False,
        lights=LEDBoardValue(red=False, green=False)
    ),
    two=CompositeOutputDeviceValue(
        button=False,
        lights=LEDBoardValue(red=False, green=False)
    ),
    three=CompositeOutputDeviceValue(
        button=False,
        lights=LEDBoardValue(red=False, green=False)
    ),
    four=CompositeOutputDeviceValue(
        button=False,
        lights=LEDBoardValue(red=False, green=False)
    ),
    five=CompositeOutputDeviceValue(
        button=False,
        lights=LEDBoardValue(red=False, green=False)
    )
)
```

More commits to the test suite, docs and so on to follow.


lurch:
Given that it contains Buttons, should it be using `CompositeDevice` instead of `CompositeOutputDevice`? (can't remember off-hand what the difference is right now)
I think I'd prefer `leds=` to `lights=` - maybe it's just me but I always think of "lights" as something much brighter than "leds".
Just out of curiosity, is there any reason that `StatusZeroValue` lists `green` before `red`, but `StatusBoardValue` lists `red` before `green` ?

gpiozero/boards.py
```diff
@@ -795,10 +795,100 @@ def __init__(self, location=None, pwm=False, initial_value=False):
...
+        status = StatusZero('wifi', 'raining')
+        status.wifi.green.on()
+        status.raining.red.on()
+
+    :param str \*labels:
+        Specify the names of the labels you wish to designate the strips to.
+        You can list up to three labels. If no labels are given, three strips
+        will be initialised with names 'one', 'two', and 'three'. If some, but
+        not all strips are given labels, any remaining strips will not be
+        initialised.
```
lurch: Interesting. I guess an alternative approach would be to label the remaining strips with number-names, so that `status = StatusZero('wifi', 'raining')` would effectively be the same as `status = StatusZero('wifi', 'raining', 'three')`.

bennuttall:
Well, the oddly-named `TrafficLightsBuzzer` ([src](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/boards.py#L850)) (should be `TrafficLightsBuzzerButton`) extends `CompositeOutputDevice`, so I just copied that.

I assume it has something to do with `CompositeOutputDevice` ([src](https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/boards.py#L38)) using the `SourceMixin`.

gpiozero/boards.py
```diff
@@ -795,10 +795,100 @@ def __init__(self, location=None, pwm=False, initial_value=False):
...
+    .. _STATUS Zero: https://thepihut.com/statuszero
+    """
+    def __init__(self, *labels, **kwargs):
+        pins = (
+            (4, 17),
+            (27, 22),
+            (10, 9),
+        )
+        if len(labels) == 0:
+            labels = ['one', 'two', 'three'][:len(pins)]
```
lurch: This `[:len(pins)]` on the end isn't necessary, is it?

gpiozero/boards.py
```diff
@@ -795,10 +795,100 @@ def __init__(self, location=None, pwm=False, initial_value=False):
...
+        elif len(labels) > len(pins):
+            raise ValueError
+        strips = OrderedDict()
+        for index, label in enumerate(labels):
+            green, red = pins[index]
+            strips[label] = LEDBoard(red=red, green=green, **kwargs)
```
lurch: Should probably specify the `_order` to match line 882

bennuttall: This is missing `_order`

gpiozero/boards.py
```diff
@@ -795,10 +795,100 @@ def __init__(self, location=None, pwm=False, initial_value=False):
...
+        status = StatusBoard('wifi', 'raining')
+        status.wifi.lights.green.on()
+        status.wifi.button.when_pressed = status.wifi.lights.toggle
+
+    :param str \*labels:
+        Specify the names of the labels you wish to designate the strips to.
+        You can list up to three labels. If no labels are given, three strips
+        will be initialised with names 'one', 'two', and 'three'. If some, but
+        not all strips are given labels, any remaining strips will not be
+        initialised.
```
lurch: I guess you need to change the references to 'three' here to 'five' ;-)

bennuttall:
> Just out of curiosity, is there any reason that `StatusZeroValue` lists `green` before `red`, but `StatusBoardValue` lists `red` before `green` ?

That's a mistake - it's missing an `_order` within the inner-`LEDBoard`: https://github.com/RPi-Distro/python-gpiozero/pull/558/files#r113163516

gpiozero/boards.py
```diff
@@ -795,10 +795,100 @@ def __init__(self, location=None, pwm=False, initial_value=False):
...
+            (11, 5, 26),
+            (6, 13, 18),
+        )
+        if len(labels) == 0:
+            labels = ['one', 'two', 'three', 'four', 'five'][:len(pins)]
+        elif len(labels) > len(pins):
+            raise ValueError
+        strips = OrderedDict()
+        for index, label in enumerate(labels):
+            green, red, button = pins[index]
```
lurch: Just out of curiosity, why order them as `red, green` in the object, but order them `green, red` in the `pins` tuples?

gpiozero/boards.py
```diff
@@ -795,10 +795,100 @@ def __init__(self, location=None, pwm=False, initial_value=False):
...
+                    red=red, green=green, _order=('red', 'green'), **kwargs
+                ), _order=('button', 'lights'), **kwargs)
+        super(StatusBoard, self).__init__(_order=strips.keys(), **strips)
+
+
 class SnowPi(LEDBoard):
     """
     Extends :class:`LEDBoard` for the `Ryanteck SnowPi`_ board.
 
+
```
lurch: Hmm? ;)

lurch:
> (should be TrafficLightsBuzzerButton really)

Yeah, I also pointed that out ages ago, but by that point it was too late to change (backwards-compatibility, etc.).

> Perhaps it has to be CompositeOutputDevice in order to have its value set with source.

Ahh, you might be right - rings a few bells :)

gpiozero/boards.py
bennuttall: Yep, realised after I pushed. Will delete the accidental newline...

gpiozero/boards.py
bennuttall: Because `TrafficLights` starts with red. That's all...

gpiozero/boards.py
bennuttall: Yep

gpiozero/boards.py
bennuttall: No - it came from the previous implementation which had a base class, so this was necessary.

gpiozero/boards.py
bennuttall: I chose not to do that in favour of allowing people to use the pins for other devices, if they need to.

lurch: Huh? But (IIRC) we don't do that with any other addon boards.

gpiozero/boards.py
lurch: I haven't seen what the statusboard looks like, but IMHO the ordering of the leds in the code should match what the board does, rather than matching what TrafficLights does (if different).

lurch:
> But it's essentially an internal class. I don't think changing it is an issue.

Well I certainly won't object ;-)

gpiozero/boards.py
bennuttall: Well I was thinking of them like a subset of traffic lights - they're status lights, were originally traffic lights but the yellow light got removed, so in the same vein. I considered proposing a generic class for them but a) couldn't think of a good name and b) *shrugs*

bennuttall:
@lurch any idea how to prevent strings like `''` and `'1'` being used on init?

bennuttall:
I've added tests and it's just failing for the cases of `StatusBoard` or `StatusZero` being initialised with invalid labels such as `''` and `'1'`. There must be other cases (I jut thought of `'hello world'` which includes a space, and anything like `in` which is a Python keyword) that need to be accounted for too.

When creating a `StatusBoard` object with invalid labels, their names get renamed somewhere anyway:

```python
>>> sb = StatusBoard('hello', 'hello world', '1', '2', 'in')
>>> sb.value
StatusBoardValue(
    hello=CompositeOutputDeviceValue(...),
    _1=CompositeOutputDeviceValue(...),
    _2=CompositeOutputDeviceValue(...), 
    _3=CompositeOutputDeviceValue(...), 
    _4=CompositeOutputDeviceValue(...)
)
```

so perhaps they could be identified as invalid at that point?

lurch:
> @lurch any idea how to prevent strings like '' and '1' being used on init?

Easiest way is probably just with a regex? A quick search finds http://stackoverflow.com/questions/5474008/regular-expression-to-confirm-whether-a-string-is-a-valid-identifier-in-python#10134719

> and anything like `in` which is a Python keyword

Hmmm, another search finds http://stackoverflow.com/questions/22864221/is-the-list-of-python-reserved-words-and-builtins-available-in-a-library


... Although if you're adding code to check whether the labels supplied are "valid", shouldn't that be done at the `CompositeDevice` level, rather than adding special code just to `StausZero` and `StatusBoard`?
Ping @waveform80

lurch:
In the unit-tests, why do you sometimes check against `True` / `False`, and other times check against `0` / `1` ?

bennuttall:
> In the unit-tests, why do you sometimes check against `True` / `False`, and other times check against `0` / `1` ?

I use `True` / `False` for `LED` mode and `0` / `1 for `PWMLED` mode.

> Easiest way is probably just with a regex?

Thanks, I'll take a look

> ... Although if you're adding code to check whether the labels supplied are "valid", shouldn't that be done at the `CompositeDevice` level, rather than adding special code just to `StausZero` and `StatusBoard`?

Sure

lurch:
> I use `True` / `False` for `LED` mode and `0` / `1` for `PWMLED` mode.

Ahaa, makes sense :+1:  Probably doesn't hurt to also check at least one `0.5` value, to check that `PWMLED` is working as expected?

bennuttall:
:+1:

waveform80:
> When creating a StatusBoard object with invalid labels, their names get renamed somewhere anyway: ...

> ... Although if you're adding code to check whether the labels supplied are "valid", shouldn't that be done at the CompositeDevice level, rather than adding special code just to StausZero and StatusBoard?
> Ping @waveform80

Yup, that's already dealt with by `namedtuple`; because its labels have to be valid Python identifiers it includes a ["rename" option](https://docs.python.org/3/library/collections.html#collections.namedtuple) which causes invalid names to become _1, _2, _3, etc. (which explains what Ben was seeing earlier).

Personally I'm fairly happy with things as they are: the namedtuple deals with invalid identifiers (and its own rules, which include no fields with a leading underscore), but we still permit them as the user specified for attribute retrieval. In case anyone's wondering how the hell you get "obj.hello world":

```python
>>> class A(object):
...     pass
...
>>> a = A()
>>> setattr(a, 'hello world', 1)
>>> getattr(a, 'hello world')
1
```

Which I acknowledge is thoroughly evil, but if someone *really* wants it, it is technically valid Python (after all, the attributes of an object are just entries in a dict).

waveform80:
Rebased and fixed tests. At this point I think the PR is probably good to go; the only thing I might be tempted to add would be to tweak invalid names to share values with the results of namedtuple's rename operation. The argument for is consistency. The argument against is that it's more "silent" alteration of the user's intention. Or we could simply raise an exception if the user specifies an invalid name (i.e. stop namedtuple from renaming stuff and just let it raise an exception if the user specifies something silly).

Thoughts or preferences?

bennuttall:
I'd rather it raised an exception if it's not too much trouble. Would that be down in LEDBoard?

waveform80:
Actually it's down in CompositeDevice (where the namedtuple is constructed). I'll add a commit for that in a mo (probably also need to move the namedtuple construction; I think it's after the super call at the moment and it should probably be before).

waveform80:
All done - namedtuple call needed to stay where it was but I threw in a few other changes too (e.g. reject duplicate labels).

waveform80:
Eurgh, spoke too soon. There are object lifetime issues with leaving namedtuple there, but only on 2.x and pypy. And apparently only *some* 2.x because I can't reproduce the issue Travis has found on my local box. This may take a little debugging ...

waveform80:
Well, that was annoying but I think it's all sorted now. Probably worth testing it with some actual boards to make sure the pin assignments are all accurate, but otherwise merge if you're happy!

bennuttall:
Thanks - I'll give it a whirl tonight

bennuttall:
Tested - I'm happy with this :+1:

gpiozero/boards.py
```diff
@@ -795,6 +795,102 @@ def __init__(self, location=None, pwm=False, initial_value=False):
...
+        from gpiozero import StatusBoard
+
+        status = StatusBoard('wifi', 'raining')
+        status.wifi.lights.green.on()
+        status.wifi.button.when_pressed = status.wifi.lights.toggle
+
+    :param str \*labels:
+        Specify the names of the labels you wish to designate the strips to.
+        You can list up to three labels. If no labels are given, three strips
```
lurch: Looks like you forgot to change the `three` mentions here to `five`?

tests/test_boards.py
```diff
@@ -778,3 +780,113 @@ def test_energenie():
...
+    with StatusZero('a') as sz:
+        assert sz.namedtuple._fields == ('a',)
+    with StatusZero('a', 'b') as sz:
+        assert sz.namedtuple._fields == ('a', 'b')
+    with StatusZero('a', 'b', 'c') as sz:
+        assert sz.namedtuple._fields == ('a', 'b', 'c')
+    with pytest.raises(ValueError):
+        StatusZero('a', 'b', 'c', 'd')
+    with pytest.raises(ValueError):
+        StatusZero('0')
```
lurch: Should these 'invalid device identifier' tests have been added to the CompositeDevice tests, rather than to just StatusZero and StatusBoard?

gpiozero/boards.py
bennuttall: Ah - missed a couple. Have pushed a fix.

tests/test_boards.py
bennuttall: Probably but not urgent to fix

PR #577, opened 2017-07-17
Prepare for v1.4.0 release

bennuttall:
Final check of docs, examples, pinout tool and man page.

Docs here: http://gpiozero.readthedocs.io/en/pinout-man/


bennuttall:
Any thoughts on the new advanced/remote recipes? Any more suggestions to add?

waveform80:
>  @waveform80 Does Debian's changelog get auto-generated? If not, it'll need this bit adding.

No - I generate it as part of the release. If you look at the "release" target in the `Makefile` you'll see some `dch` stuff in there which is what auto-generates the template of the new changelog section (which ensures the timestamp is reasonably accurate and that the maintainer is correct and so on). Hence, when I run `make release` it fires up an editor at which point I do a little vim magic to copy over the latest chunk of the changelog.rst and strip out the ReST formatting. That in turn then enables the `make upload` target to accurately build the packages with the right version.

> Any thoughts on the new advanced/remote recipes? Any more suggestions to add?

They look great! Can't think of much else to add right now, but I'll see if I can come up with anything today when I'm testing.

bennuttall:
Ok cool thanks!

bennuttall:
@waveform80 some hard-coded links to docs pages in README link to /latest/ branch. Should they be /stable/?

waveform80:
Oh yeah, they probably should - good point.

bennuttall:
It looks like there's a missing dependency in the "Building the docs" section of development docs (how meta): I think it's the `sphinx-common` package.

bennuttall:
Ok - ready for release, besides any small tweaks @waveform80 wants to make.

bennuttall:
:ship: ship it

PR #579, opened 2017-07-17
Added required dependencies in setup.py (RPi.GPIO, spidev)

matthewgall:
gpiozero relies on these libraries in order to correctly identify
the pin setup on the Raspberry Pi, and the absence of these libraries
results in weird issues relating to the incorrect identification of
PWM support on unsupported pins.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,8 @@ __keywords__ = [
 ]
 
 __requires__ = [
+    'RPi.GPIO',
+    'spidev'
 ]
 
 __extra_requires__ = {


matthewgall:
Looks like documentation is improving in latest release. Closing as no longer necessary.

bennuttall:
Thanks for contributing, but those libraries are not dependencies. For most users, rpi.gpio or some other supported pin library is required, that's why `sudo apt install python3-gpiozero` brings rpi.gpio and spidev with it. But for some users, they are not required.

For example, if you're using the remote gpio feature from your laptop, you don't want (and probably can't get) rpi.gpio. Read more:
- http://bennuttall.com/whats-new-gpio-zero-v1-2/
- http://bennuttall.com/whats-new-gpio-zero-v1-3/
- http://gpiozero.readthedocs.io/en/latest/api_pins.html
- http://gpiozero.readthedocs.io/en/latest/remote_gpio.html

What happens when you use gpiozero without a pin library is it uses NativePin, an experimental pin implementation which uses the filesystem gpio interface. It works ok for LEDs and buttons but isn't recommended - and doesn't support PWM. Because `Robot` uses PWM by default (for variable speed), it will fail for that. You can disable this (`Robot(left=(pin, pin), right=(pin, pin), pwm=False)`) but you're better off just installing rpi.gpio.

Hope that helps! Sorry for the mix-up. Our documentation of this is better in v1.4 and the warnings when pin libraries are not found is better. It'll be out this week, feel free to test!

Issue #385, opened 2016-06-21
DistanceSensor returning fixed values 


waveform80:
During picademy today we had an issue with a couple of ultrasonic distance sensors - they seemed to return 19-20cm no matter what (or 0 when the target was particularly close). Failed to debug the issue during the hack session but I tried again at home: same issue on the development pi.

After hooking up the oscilloscope it looks like the Event.wait (or possibly the Event.set?) that are called in _read take a very long time and don't seem to vary with the echo pin (leading to the fixed readings).

Still not sure about the root cause- will dig further tomorrow!

bennuttall:
I think I've experienced something similar.

waveform80:
Well, this has been a fascinating morning! The oscilloscope has proved a magnificent tool in debugging and the results are rather interesting (they're probably also going to lead to some modifications of the pins callback API - but that's okay because it's not declared stable yet).

First I'll describe the problem in a bit more detail:

The issue appears that be echo timings from the ultrasonic sensor weren't being measured accurately by Python. The question was whether this was down to the callbacks being too slow or whether the threading primitives (used to synchronize the callback thread with the reading thread) were either too slow or introducing some form of quantization to the timings. There was also a question as to whether there's a difference between Python 2 and 3 in this area, and whether the pin implementation in use makes a difference (particularly given pigpiod runs over a socket connection).

The test setup used was as follows:

* Hook the ultrasonic sensor up to the Pi with an appropriate voltage divider (in my case I used 3x330ohm resistors as I couldn't seem to find the 470 in my box, but the 2/3 divider is perfect anyway)
* Hooked the trigger pin of the sensor into L0 on the Bitscope Micro
* Hooked the echo pin into L1
* Hook a couple more GPIO pins for testing into L2 and L3
* Adjusted the DistanceSensor code so that the L2 pin copies the echo pin's state as soon as the callback is called
* Adjusted the DistanceSensor code so that the L3 pin is set high by the first event wait, and low by the second event wait

So ... once DistanceSensor is  constructed and the background reading thread starts using the sensor constantly, in an ideal world what ought to happen is that we see L0 (trigger) blip high briefly, followed by L1 (echo) raising high for a length of time proportional to the echo, with L2 closely following L1's state, and L3 closely following L2's state.

Under Python 3, with the RPi.GPIO implementation this is more or less what we see (L0 is at the bottom, L1 is next up and so on):

[...]

PR #595, opened 2017-10-20
Fix distance sensor for Pi3 #385

SteveAmor:
Works like a treat on Pi3.  Tested ok on Jessie lite Pi3 but not brilliant on ZeroW.  Fix for #385 will need the DMA engine level change timestamp.


SteveAmor:
Fix for #385

gbaman:
Our robots team at @niraspberryjam have done tests with this PR and can confirm it fixes the issues with the ultrasonic sensor, given the sensor actually works now 👍

SteveAmor:
Included in #617

PR #604, opened 2017-11-17
Add --xyz option to pinout cli tool

bennuttall:
I suggested at one point that we add an option to `pinout` to open [pinout.xyz](https://pinout.xyz) in a browser but we decided against it due to cross-platform issues.

However, it occured to me that `import antigravity` is in the standard library and I assume it works cross-platform.

I looked at the [source](https://github.com/python/cpython/blob/master/Lib/antigravity.py) for `import antigravity` and it uses the built-in `webbrowser` module ([source](https://github.com/python/cpython/blob/master/Lib/webbrowser.py)).

I haven't tested it on another machine yet.

Usage:

```
$ pinout -x
$ pinout --xyz
```

WIP: needs docs updating if approved in principle.

### DIFF
--- a/docs/cli_pinout.rst
+++ b/docs/cli_pinout.rst
@@ -43,7 +43,7 @@ Options
 
 .. option:: -x, --xyz
 
-    Open pinout.xyz in the default web browser
+    Open `pinout.xyz`_ in the default web browser
 
 Examples
 --------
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Ahoy! I started some work on an API for Pinout.xyz. I'm keen to know if I can do anything to help your goals, and help expose more detailed device information for GPIO Zero tools to consume?

PR #616, opened 2018-01-21
Fix SmoothedInputDevice to use the mean of the values, matching the docs

### LINKED ISSUE #615: SmoothedInputDevice (wrongly?) uses the Median (instead of the Average) on the queue to return the value
While in the documentation it says that the *mean* or *average* of the values in the internal queue will be compared to `threshold` to determine whether device `is_active`, `.value` in fact returns the **median** of the queue. I fixed it for myself in `mixins.py`, by importing the `mean` and then passing it as the default `average` keyword parameter to the `__init__()`-function of `GPIOQueue()`.

I wonder whether taking the median serves a purpose i'm unaware of (apart from contradicting the documentation, hehe), but given digital inputs, the median can only return 0 or 1 (or very rarely 0.5), strongly reducing the functionality of the `threshold`.

I suggest changing that behavior to using the **arithmetic mean** instead of the median?

lurch:
Fixes #615 

I'm afraid I couldn't think of any way of unit-testing this.

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -9,6 +9,10 @@ from __future__ import (
 
 from time import sleep, time
 from threading import Event
+try:
+    from statistics import mean
+except ImportError:
+    from .compat import mean
 
 from .exc import InputDeviceError, DeviceClosed
 from .devices import GPIODevice
@@ -151,7 +155,7 @@ class SmoothedInputDevice(EventsMixin, InputDevice):
             pin, pull_up, pin_factory=pin_factory
         )
         try:
-            self._queue = GPIOQueue(self, queue_len, sample_wait, partial)
+            self._queue = GPIOQueue(self, queue_len, sample_wait, partial, mean)
             self.threshold = float(threshold)
         except:
             self.close()


waveform80:
Oh hell - I actually said mean in the docs. Hmmm. I'm sorely tempted to change the docs to say median (or just more vaguely "average") rather than switch to mean (median's awfully nice for discarding silly outliers which is exactly what we want for jittery sensors - fully half the queue's values have to be "wrong" before the median is affected, unlike the mean).

bennuttall:
That makes sense, do you want to just use median?

holzfigure:
i'm using the SmoothedInputDevice to read back values from a PWMOutputDevice, where i really need the "mean", otherwise i only get either 0 or 1. But maybe that's a silly way of me to try to achieve reading the PWM easily.. ?

waveform80:
Yes, to be honest I'd much rather stick to median, and expose the choice to the user (and correct the docs to say something vague like "average") rather than switch to mean. For things like DistanceSensor, median is *much* better as it means the occasional "max distance" reading (which can occur when an echo edge is missed) doesn't "leap" the reading.

Exposing it to the user shouldn't be too hard - just means exposing some more params of GPIOQueue in SmoothedInputDevice and its descendants. I'll see if I can squeeze that in tonight.

waveform80:
Actually not even the descendents - just SmoothedInputDevice would do for @holzfigure's use case. Right, I'll bash this together now.

holzfigure:
Thanks so much! This is perfect for my use case and allows for so much more fun! ,)

PR #629, opened 2018-03-25
Add socket method for Energenie

SteveAmor:
Replaces #486

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1603,6 +1603,10 @@ class Energenie(SourceMixin, Device):
             return "<gpiozero.Energenie object closed>"
 
     @property
+    def socket(self):
+        return self._socket
+
+    @property
     def value(self):
         return self._value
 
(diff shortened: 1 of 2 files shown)


bennuttall:
Thanks @SteveAmor :+1: 

Note: docs currently show no useful properties/methods for Energenie, I think things like on/off/socket need to be specified.

SteveAmor:
@bennuttall building the docs is very broken.  First off, it complains about not having Sphinx installed.  I installed Sphinx with pip and then when you run ```make doc``` pretty well everything errors and the build fails.  So I can't test that I've updated the docstrings correctly :-(

bennuttall:
I assume it's `apt install sphinx` rather than `pip`. If that's missing in the docs, we'll add it.

I can't test easily as I know I have a working setup on my machine, but I guess I could always try on a Pi.

SteveAmor:
sudo apt install sphinx
E: Unable to locate package sphinx
:-(  Hence trying pip :-(

Issue #583, opened 2017-08-04
Neopixel support


bennuttall:
A big task, but we should see if we can add neopixel support.

MauiJerry:
check the pi neopixel library (https://github.com/jgarff/rpi_ws281x) for example code. 
since that is all built separately from rest of this system, it might take a fair bit of effort.

jantman:
I'd posit that these days, a library without Neopixel support will probably be ignored (with frustration) by a *lot* of people...

bennuttall:
Don't write comments like that. Open source maintainers don't owe you anything.

gpiozero had simple goals to make it easier for educators to teach basic physcial computing. We expanded the library a lot but things like I2C and neopixels add a whole layer of complexity to it. There are other libraries you can use to control neopixels and you can use them alongside gpiozero.

dsssssssss9:
> I'd posit that these days, a library without Neopixel support will probably be ignored (with frustration) by a _lot_ of people...

Not a very kind thing to say!

You can  successfully run GPIZERO & separate Neopixel / WS2812 library simultaneously 

Perhaps you should try this ?

Issue #551, opened 2017-03-28
Add SPI output support for APA102 LEDs


bennuttall:
The APA102 is an RGB LED similar to the neopixel. They're easy to control with SPI, they're cheap, and they're getting more common.

They're used in Pimoroni's [Blinkt](https://github.com/pimoroni/blinkt), and Alex Eames's [Inspiring](https://www.kickstarter.com/projects/raspitv/raspio-inspiring-connect-rgb-led-shapes-sculpt-you).

Pimoroni's library uses RPi.GPIO to bit-bang SPI: https://github.com/pimoroni/blinkt/blob/master/library/blinkt.py

Our SPI only currently supports analogue input devices like ADCs, so the SPI base class would need some work, but would support both hardware-based (via spidev or pigpio) SPI, and software-based bit-banging.

joan2937:
Coincidentally http://abyz.co.uk/rpi/pigpio/examples.html#Python_test-APA102_py

bennuttall:
Great - thanks

rgm3:
I'd love support for Blinkt via pigpio SPI so that it'd be easy to synchronize patterns across many Blinkt boards over the network.

dglaude:
I found this pigpio APA102 example:
http://abyz.co.uk/rpi/pigpio/examples.html
http://abyz.co.uk/rpi/pigpio/code/test-APA102_py.zip

It might be of interest.

One tricky part will have to decide between bit bang or native SPI
depending on the accessory you want to work with.

Blinkt! is not on the right GPIO for native SPI, so bit-bang is the only
option without hardware modification.
Rainbow HAT (with 7 APA102) is on SPI and can be controlled that way
(however Pimoroni library does not use spidev library, so maybe bit bang
again?
Inspire APA102 are using SPI and the (not released) library rely on spidev.

If someone work on gpiozero implementation of APA102 support, I can help
with testing and doing demo.
I have many APA102 accessories for the Pi.

2017-05-01 22:08 GMT+02:00 rgm <notifications@github.com>:

tamboekie:
It would be good to include support for WS2812B / neopixels, too? Currently, I use the [rpi_ws281x](https://github.com/jgarff/rpi_ws281x) library, but integrating this into `gpiozero` would be fantastic.

bennuttall:
[...]

PR #631, opened 2018-03-25
Add RGBLEDBoard Class

### LINKED ISSUE #81: Provide wrappers for LED groups on add-on boards
```python
fish = FishDish()

fish.lights_on()
fish.lights_off()
fish.toggle_lights()
fish.blink_lights()
```

should become:

```python
fish = FishDish()

fish.lights.on()
fish.lights.off()
fish.lights.toggle()
fish.lights.blink()
```

by use of the `TrafficLights` class. Same applies to `TrafficHat`.

bennuttall:
Adding an `RGBLEDBoard` class which allows users to create arbitrarily sized boards comprising multiple RGB LEDs. The API is assumed based on a mix of `LEDBoard` and `RGBLED`, i.e. it has on/off/toggle/blink/pulse, optional PWM, on/off/toggle should take a numerical argument (to control a single LED or a number of LEDs rather than the whole board), blink should allow on_color, off_color, etc.

```python
from gpiozero import RGBLEDBoard

leds = RGBLEDBoard((2, 3, 4), (5, 6, 7), (7, 8, 9))

leds.value = ((1, 0, 0), (0, 1, 0), (0, 0, 1))
```

This paves the way for the API for classes for smarter RGB LEDs like APA102 (#551) and neopixels (#583) when we support those underlying protocols.


gpiozero/boards.py
```diff
@@ -59,15 +59,15 @@ def on(self):
         Turn all the output devices on.
         """
         for device in self:
-            if isinstance(device, (OutputDevice, CompositeOutputDevice)):
+            if isinstance(device, (OutputDevice, CompositeOutputDevice, RGBLED)):
```
bennuttall: I don't like this. @waveform80 @lurch do you think it's possible for `RGBLED` to extend `OutputDevice` or `CompositeOutputDevice` rather than just `Device`?

bennuttall: Actually I got around this by implementing on/off/toggle directly in RGBLEDBoard rather than using the parent class. Just working on getting blink to work too.

gpiozero/boards.py
```diff
@@ -1053,6 +1053,71 @@ def __init__(self, pwm=False, initial_value=False, pin_factory=None):
         )
 
 
+class RGBLEDBoard(CompositeOutputDevice):
+    """
+    Extends :class:`CompositeOutputDevice`. and represents a generic RGBLED
+    board or collection of LEDs.
+
+    The following example turns on all the LEDs on a board containing 5 LEDs
+    attached to GPIO pins 2 through 6::
```
lurch: Did you mean "2 through 9" ?  ;)

gpiozero/boards.py
```diff
@@ -1053,6 +1053,71 @@ def __init__(self, pwm=False, initial_value=False, pin_factory=None):
...
+
+        from gpiozero import RGBLEDBoard
+
+        leds = RGBLEDBoard((2, 3, 4), (5, 6, 7), (7, 8, 9))
```
lurch: Hmm, probably shouldn't be using GPIO7 twice?

bennuttall: whoops!

gpiozero/boards.py
bennuttall: copy/paste error!

gpiozero/boards.py
```diff
@@ -1053,6 +1053,71 @@ def __init__(self, pwm=False, initial_value=False, pin_factory=None):
...
+        active_high = kwargs.pop('active_high', True)
+        initial_value = kwargs.pop('initial_value', False)
+        pin_factory = kwargs.pop('pin_factory', None)
+        order = kwargs.pop('_order', None)
+        super(RGBLEDBoard, self).__init__(
+            *(
+                pins_or_board
+                if isinstance(pins_or_board, RGBLEDBoard) else
+                RGBLED(
+                    *pins_or_board, pwm=pwm, active_high=active_high, initial_value=(initial_value, ) * 3,
```
lurch: Why have you got `initial_value=(initial_value, ) * 3,` ? IMHO RGBLEDBoard should have the same init-params as https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/output_devices.py#L581

gpiozero/boards.py
lurch: IIRC @waveform80 implemented `RGBLED` as a child of `Device` rather than as a child of `OutputDevice` or `CompositeOutputDevice` because the `value` of an RGBLED is a 3-tuple and not a boolean or 0 -> 1 float.
And yup, RGBLED also implements it's own on / off / toggle / blink / pulse methods.

gpiozero/boards.py
bennuttall: I grabbed some implementation from `LEDCollection` which uses `True` / `False` for `initial_value`. Why do `LEDCollection` and `LEDBoard` have boolean `initial_value` when `.value` itself is an n-tuple?

lurch: I'm not sure - perhaps @waveform80 remembers more? 
You could argue that the `initial_value` for an `LEDCollection` should be a tuple-of-booleans of the same size as the `.value`, or you could argue that the `initial_value` is only a default value, so you'd want to initialise every LED in an LEDCollection with the same value? https://gpiozero.readthedocs.io/en/stable/api_boards.html#ledboard

I guess another option would be to for the `__init__` to support both an `initial_value` param (a boolean) **and** an `initial_values` param (a tuple of values), with an exception being raised if both are supplied, or an exception being raised if the number of tuples in `initial_values` is incorrect?

\*shrug\*

Oh, and is `LEDCollection` the class that allows nesting, i.e. each child-element can be an `LED` _or_ another `LEDCollection` object? I guess that would also make an `initial_values` param slightly more complex? (because you'd need to check that nested `LEDCollection`s have the same number of tuples / sub-nested `LEDCollection`s etc.)

bennuttall: I think we need consistency foremost. So I'll go for `initial_value` being a single value for now. Do we have any classes where `initial_value` is a tuple the length (and shape) of the board itself? I do like the idea of an optional single value or tuple. The tricky thing (I think) is working out the length and shape at the point of looking it up in `kwargs`.

bennuttall: Actually, on second thoughts, it makes more sense that it follows `RGBLED` in expecting a ringle 3-tuple (in this case all RGBLEDs would be set to the same colour). This is also consistent with `LEDBoard` in that `LEDBoard` takes the same `initial_value` as `LED` or `PWMLED`.

Anyone wanting to set a more specific range of colours can do so manually after init.

lurch: > Actually, on second thoughts, it makes more sense that it follows RGBLED in expecting a ringle 3-tuple (in this case all RGBLEDs would be set to the same colour).

Haha, exactly as I suggested [above](https://github.com/RPi-Distro/python-gpiozero/pull/631#discussion_r177042139) ;-)

bennuttall: Got there in the end!

bennuttall:
Getting there! All tests passing except the blink ones, and I need to add some more blink and pulse tests, as well as look at nested RGBLEDBoards.

I'm not entirely happy with the implementation of the blink tests:

```python
board.blink()
assert board.value == ((1, 1, 1), (1, 1, 1))
board.off()
board[0].blink()
assert board.value == ((1, 1, 1), (0, 0, 0))
```

Ping @martinohanlon

tests/test_boards.py
```diff
@@ -695,6 +701,135 @@ def test_snow_pi_initial_value_pwm():
...
+    with pytest.raises(ValueError):
+        RGBLEDBoard()
+    with pytest.raises(ValueError):
+        RGBLEDBoard(2)
+    with pytest.raises(ValueError):
+        RGBLEDBoard(pin=2)
+    with pytest.raises(ValueError):
+        RGBLEDBoard(2, 3, 4)
+
+def test_rgbledboard_bad_init_pwm():
```
martinohanlon: I am confused by these tests.

`RGBLEDBoard((2, 3, 4))` is expected to raise PinPWMUnsupported because the default for `pwn` is `True` (https://github.com/RPi-Distro/python-gpiozero/pull/631/files#diff-8e73cd872390a78e8e626359ec7c5837R1074)

But why are tests such as `RGBLEDBoard((2, 3, 4), (5, 6, 7))` expected to pass?

bennuttall: Because those tests are in the list of tests to use `MockPWMPin`: https://github.com/RPi-Distro/python-gpiozero/blob/70fc6af18b2a2dc07dced4a85c2ba62830f7517b/tests/test_boards.py#L20

martinohanlon: got ya.

I like the comment above it `# dirty, but it does the job`

bennuttall: :)

bennuttall:
@lurch I'm struggling to comprehend the herculean `blink` method as implemented in `RGBLED`. Would you be able to take a look and see if you can implement it for `RGBLEDBoard` or perhaps we could pair on it some time?

PR #632, opened 2018-03-26
Update instructions to create a Python virtualenv

SteveAmor:
Also removed the ```>``` from the sudo apt install lines to make cut and pasting easier

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -33,9 +33,11 @@ within a virtual Python environment:
 .. code-block:: console
 
     $ sudo apt install lsb-release build-essential git git-core \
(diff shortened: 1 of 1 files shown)


SteveAmor:
@waveform80 without the source statement in the .bashrc file I get ```mkvirtualenv: command not found```.  The terminal session doesn't "appear" to be in a virtualenv and I can create other virtualenvs and deactivte and workon the second virtualenv I created (which also created a .virtualenv directory).

SteveAmor:
Adding the source statement in the .bashrc file is causing issues with ```make doc``` so it's not the answer.  I see that the piwheels numpy test doesn't use the virtualenvwrapper.  So I'm not sure what the answer is.

Issue #630, opened 2018-03-25
Development - mkvirtualenv: command not found


SteveAmor:
In the docs, section 11.1. (Development installation) details the process to install a virtual Python environment.  However, after installing the Python packages/modules, the command ```mkvirtualenv -p /usr/bin/python3 python-gpiozero``` returns ```-bash: mkvirtualenv: command not found```. 

Tested with both apt and apt-get on new install of Stretch Lite (after a sudo apt-get update).

SteveAmor:
Just been looking at piwheels and the test and the virtualenv it uses so I'm working on a fix for sectionn 11.1.  Will raise a PR later (I hope).

SteveAmor:
Fixed in #632

SteveAmor:
According to the official documentation (https://virtualenvwrapper.readthedocs.io/en/latest/install.html) after creating a virtualenv directory using ```mkdir ~/.virtualenvs``` the following should be added to .bashrc:
```
export WORKON_HOME=/home/pi/.virtualenvs
source /usr/share/virtualenvwrapper/virtualenvwrapper.sh
````
Then run ```source ~/.bashrc```

waveform80:
Fixed in 3685ffb72a1ffb70bc18a1fd62169b442adf0999 - remember to close when docs-tidy-up PR is merged

PR #636, opened 2018-03-27
Fix creation of virtualenv using mkvirtualenv

SteveAmor:
See issue reported in #630 
I have tested that you can move around virtualenvs with ```workon <virtualenv>``` and that you can ```make develop``` and ```make doc``` without any errors.
It was almost fixed in #632 but I hadn't created and linked to the .virtualenv directory which broke ```make doc```.  @waveform80 could you please check this PR as @bennuttall is concerned about adding the lines to .bashrc (see https://github.com/RPi-Distro/python-gpiozero/pull/632#issuecomment-376183461).

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -33,9 +33,13 @@ within a virtual Python environment:
 .. code-block:: console
 
     $ sudo apt install lsb-release build-essential git git-core \
-    >   exuberant-ctags virtualenvwrapper python-virtualenv python3-virtualenv \
-    >   python-dev python3-dev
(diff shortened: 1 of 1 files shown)


SteveAmor:
Also removed ```>``` from apt install command to make cutting and pasting easier.

bennuttall:
Ah, that makes sense. Would `hash -r` fix it like with the pip upgrade thing?

bennuttall:
Dave says no. I'm closing this PR as it's not the right solution but let's address in the original issue by correcting the docs :+1: 

See #630

PR #637, opened 2018-03-27
Correct html and pdf doc location

SteveAmor:


bennuttall:
I assume this is intentional and `docs/` is assumed. @waveform80?

SteveAmor:
@bennuttall @waveform80 after ```make docs``` the files are not under ```docs/``` they are under ```build/``` so it could be the ```Makefile``` needs updating?

PR #646, opened 2018-04-23
Fix some of the examples to be Python2-compatible.

### LINKED ISSUE #645: rgbled.py example included int recipes.rst (Basic Recipes) does not work with Python 2
Hi! First, thanks for developing and maintaining this package.

[Line 27 of rgbled.py](https://github.com/RPi-Distro/python-gpiozero/blob/master/docs/examples/rgbled.py#L27) example code that is included in the Basic Recipes documentation does not work in python 2, while it does in python 3.  Under python 2 this division calculation always results in 0, so the blue LED never gets lit, while in python 3 it returns the intended floating point value.

Recommend either dividing by 100.0 to force a floating point context, or putting a comment that python 3 is required.

lurch:
Fixes #645 

(Note that I simply did a `grep` for examples that were using `/` - I didn't physically check that all the examples work correctly on Python2)

### DIFF
--- a/docs/examples/led_bargraph_1.py
+++ b/docs/examples/led_bargraph_1.py
@@ -1,6 +1,6 @@
 from gpiozero import LEDBarGraph
 from time import sleep
-from __future__ import division
+from __future__ import division  # required for python 2
 
 graph = LEDBarGraph(5, 6, 13, 19, 26, 20)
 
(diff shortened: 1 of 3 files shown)


bennuttall:
Thanks! I've added comments to state that's required for Python 2.

PR #654, opened 2018-06-13
Allow SmoothedInputDevice, LineSensor and MotionSensor to have pull-up configured, close #652

### LINKED ISSUE #652: LineSensor - inverted output
I am using this line sensor...  

https://www.aliexpress.com/item/Line-tracking-Sensor-For-robotic-and-car-DIY-Arduino-projects-Digital-Out/32654587628.html?spm=a2g0s.9042311.0.0.72924c4dZUXJvS

In summary - LineSensor.line_detected returns `True` when not on a line and `False` when on a line.

Its output is inverted to the LineSensor implemented in gpiozero. It is 'active' when the line is not detected and gpiozero offers no way of changing this. 

I need to be able to set pull_up=True on the SmoothInputDevice.

My current hack looks like this:

```python
left_sensor = LineSensor(17)
left_sensor._active_state = not left_sensor._active_state
left_sensor._inactive_state = not left_sensor._inactive_state
```

is having a pull_up on the LineSensor constructor the right implementation though? Is it really a pull up? or more of an `active_state`?

bennuttall:
I adding missing `pin` and `pull_up` init params from the docstrings, and tidied up the init params a bit.

Ping @martinohanlon give it a test?

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -80,6 +80,14 @@ class DigitalInputDevice(EventsMixin, InputDevice):
     straight forward on / off states with (reasonably) clean transitions
     between the two.
 
+    :param int pin:
+        The GPIO pin (in Broadcom numbering) that the device is connected to.
+        If this is ``None`` a :exc:`GPIODeviceError` will be raised.
+
+    :param bool pull_up:
+        If ``True``, the pin will be pulled high with an internal resistor. If
+        ``False`` (the default), the pin will be pulled low.
+
     :param float bounce_time:
         Specifies the length of time (in seconds) that the component will
         ignore changes in state after an initial change. This defaults to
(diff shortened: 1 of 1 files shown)


gpiozero/input_devices.py
```diff
@@ -366,13 +386,11 @@ class LineSensor(SmoothedInputDevice):
     .. _CamJam #3 EduKit: http://camjam.me/?page_id=1035
     """
     def __init__(
-            self, pin=None, queue_len=5, sample_rate=100, threshold=0.5,
-            partial=False, pin_factory=None):
+            self, pin=None, pull_up=False, queue_len=5, sample_rate=100,
```
lurch: By changing the order of the init-params, does this break backwards-compatibility for anyone who was using positional args and not keyword args?

lurch:
PR says `LightSensor` but I think you meant `LineSensor` ? ;-)

martinohanlon:
@bennuttall will do

gpiozero/input_devices.py
lurch: I'm not bothered either way (i.e. I'd be happy with this change) - I only mentioned it because I remembered @waveform80 making a big fuss about backwards-compatibility previously :wink:

martinohanlon:
Tested... works :)

PR #655, opened 2018-06-18
Add support for colorzero with RGBLED

bennuttall:
I wondered if we could support using `colorzero.Color` for RGBLEDBoard's `color` property, and similarly `Red`, `Green` and `Blue` while allowing those properties to be set the current way (with integers / integer tuples).

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -8,6 +8,7 @@ from __future__ import (
 from threading import Lock
 from itertools import repeat, cycle, chain
 from collections import OrderedDict
+from colorzero import Color, Red, Green, Blue
 
 from .exc import OutputDeviceBadValue, GPIOPinMissing
 from .devices import GPIODevice, Device, CompositeDevice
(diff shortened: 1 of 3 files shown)


gpiozero/output_devices.py
```diff
@@ -654,7 +642,56 @@ def is_active(self):
...
+
+    @property
+    def red(self):
+        return Red(self._leds[0].value)
+
+    @red.setter
+    def red(self, value):
+        self._stop_blink()
+        if isinstance(value, Red):
+            self.value = float(value)
```
waveform80: Surely this will blow up? In the value setter above it starts with `for component in value` ... but value is a float which isn't iterable. I don't think you need to test whether the value descends from the `Red` class at all (given `Red` is actually a sub-class of `float` anyway), unless you're intending to block assignment of `Green` and `Blue` values to the property, but this test won't actually do that anyway as they'll fall through to the `else` clause successfully?

bennuttall:
Added tests - I've identified that while you can't set a non-PWM RGBLED's `value` using floats, you *can* set it's red/green/blue properties and they get turned into booleans (so `0.5` becomes `True`). I'll address that in my branch too.

gpiozero/output_devices.py
bennuttall: Yup, tests have shown that. I'll get them passing and give you another look.

bennuttall:
Failing on 2.7 but it's just timing issues (unrelated to this PR) - perhaps we need to adjust the tolerances?

PR #658, opened 2018-06-24
Allow input devices to be pulled up or down externally, close #593

### LINKED ISSUE #593: Button with external pullup/pulldown resistor
**Feature Request**:

I have need of a `gpiozero.Button`  or `gpiozero.DigitalInputDevice` object that doesn't pull up or pull down the pin via the internal resistors in the BCM chip, so that instead I can use external resistors to do so.

In my case, I have a voltage divider on the RPi pin to protect it from overvoltage. I am also using very high resistance to slow down the voltage divider's drain on my battery. I need to detect when the voltage on the pin goes to ground.

I've successfully utilized the following `RPi.GPIO` code for my application. Note lack of `pull_up_down` attribute in `GPIO.setup()`.

```python
import RPi.GPIO as GPIO
def poweroff():
        from subprocess import call
        print("shutting down")
        call(["sudo","poweroff"])
slee=10
pin=19
GPIO.setmode(GPIO.BCM)
GPIO.setup(pin, GPIO.IN)
while True:
        val = bool(GPIO.input(pin))
        if not val:
                print "pin {p} reads {v}".format(p=pin,v=val)
                poweroff()
        sleep(slee)
```

If instead I call `GPIO.setup(pin, GPIO.IN,pull_up_down=GPIO.PUD_UP)`, then `val` always evaluates `True` regardless of the voltage on the pin, and if I call `GPIO.setup(pin, GPIO.IN,pull_up_down=GPIO.PUD_DOWN)`, `val` always evaluates `False`. This is because (according to my current theory) since the internal resistor is at 100 kiloohms, it offers much less resistance than the R2 (to ground) resistor on my voltage divider. So the pin never sees the voltage go to zero because the internal resistor remains the lowest resistance to ground or high.

My reading of the documentation for `gpiozero.Button` is that `pull_up` is asserted (`True`) by default and the pin is pulled up; if `pull_up` is set to `False`, then the pin is pulled down.  So I expect it to work like it does in the above `RPi.GPIO` example when pulled up or down, which won't work for my application. The documentation for `gpiozero.DigitalInputDevice` shows `pull_up` is a valid argument in the constructor, but doesn't discuss it further; I assume it has the same behaviour as Button. So it seems in the API, there's no way to avoid using an internal up/down resistor.

Is this reading correct? If so, is it reasonable to ask for a way for the Button or DigitialInputDevice class to neither pull up or pull down, to accommodate for use of external pull up/down resistors? (Maybe by setting `pull_up=None`?)

I can certainly continue to use RPi.GPIO, but I like gpiozero and want to use it more.

**Background**: 

I am trying to read a low battery (LB) notification signal from a LiPoly voltage booster without having to also add an ADC or active voltage reduction circuitry to the RPi. The LB signal is at battery voltage until battery dips to 3.2 Volts, at which point the LB signal goes to nearly zero (a few millivolts). When the battery is full, the LB is at 4.18 Volts, thus the need for the divider.

My Voltage Divider resistor R1, from the LB signal source to RPi pin, is 1.5 megaohms; resistor R2, from RPi pin to ground, is 5.6 megaohms. The LB signal may be anywhere between 3.2 to 4.18 volts, but will drop to nearly 0 when indicating Low Battery state. 

Alternatively, from experimentation it seems I could avoid the Voltage Divider and just put a very high resistor from the LB signal to the RPi pin to drop the voltage. This avoids bringing the pin to ground and the associated power drain (no matter how small it is). A 5.6 Megaohm resister seems to cause about 1 volt of drop, which seems sufficient. But 1) I'm not sure if doing that is a safe/reliable way to ensure reduced voltage, and 2) I realize this isn't the forum for general electrical engineering questions. :)

pmuller:
Fix #593

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -334,7 +334,7 @@ class Button(HoldMixin, DigitalInputDevice):
     :param active_state:
         If ``True``, when the hardware pin state is ``HIGH``,
         the software pin is ``HIGH``.
-        If ``False``, the input polary is reversed:
+        If ``False``, the input polarity is reversed:
         when the hardware pin state is ``HIGH``,
         the software pin state is ``LOW``.
         Use this parameter to set the active state of the underlying pin when
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -50,7 +50,7 @@ class InputDevice(GPIODevice):
     :param active_state:
         If ``True``, when the hardware pin state is ``HIGH``,
         the software pin is ``HIGH``.
-        If ``False``, the input polary is reversed:
+        If ``False``, the input polarity is reversed:
         when the hardware pin state is ``HIGH``,
         the software pin state is ``LOW``.
         Use this parameter to set the active state of the underlying pin when
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -15,7 +15,8 @@ try:
 except ImportError:
     from .compat import median
 
-from .exc import InputDeviceError, DeviceClosed, DistanceSensorNoEcho
+from .exc import InputDeviceError, DeviceClosed, DistanceSensorNoEcho, \
+    PinInvalidState
 from .devices import GPIODevice
 from .mixins import GPIOQueue, EventsMixin, HoldMixin
 
@@ -34,26 +35,52 @@ class InputDevice(GPIODevice):
         The GPIO pin (in Broadcom numbering) that the device is connected to.
         If this is ``None`` a :exc:`GPIODeviceError` will be raised.
 
-    :param bool pull_up:
+    :param pull_up:
         If ``True``, the pin will be pulled high with an internal resistor. If
         ``False`` (the default), the pin will be pulled low.
+        If ``None``, the pin will be floating. As gpiozero cannot automatically
+        guess the active state when not pulling the pin, the ``active_state``
+        parameter must be passed.
+    :type pull_up: ``bool`` or ``None``
 
     :param Factory pin_factory:
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
+
+    :param active_state:
+        If ``True``, when the hardware pin state is ``HIGH``,
+        the software pin is ``HIGH``.
+        If ``False``, the input polary is reversed:
+        when the hardware pin state is ``HIGH``,
+        the software pin state is ``LOW``.
+        Use this parameter to set the active state of the underlying pin when
+        configuring it as not pulled (``pull_up=None``).
+        When ``pull_up`` is ``True`` or ``False``, the active state is
+        automatically set to the proper value.
+
     """
-    def __init__(self, pin=None, pull_up=False, pin_factory=None):
+    def __init__(self, pin=None, pull_up=False, pin_factory=None,
+                 active_state=None):
         super(InputDevice, self).__init__(pin, pin_factory=pin_factory)
         try:
             self.pin.function = 'input'
-            pull = 'up' if pull_up else 'down'
-            if self.pin.pull != pull:
-                self.pin.pull = pull
+            if pull_up is None:
+                self.pin.pull = 'floating'
+            else:
+                self.pin.pull = 'up' if pull_up else 'down'
         except:
             self.close()
             raise
-        self._active_state = False if pull_up else True
-        self._inactive_state = True if pull_up else False
+
+        if pull_up is None:
+            if active_state is None:
+                raise PinInvalidState(
+                    'Pin %d is defined as floating, but "active_state" is not '
+                    'defined' % self.pin.number)
+            self._active_state = bool(active_state)
+        else:
+            self._active_state = False if pull_up else True
+        self._inactive_state = not self._active_state
 
     @property
     def pull_up(self):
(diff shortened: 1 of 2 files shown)


tests/test_inputs.py
```diff
@@ -55,6 +55,26 @@ def test_input_pulled_up():
...
+    pin = Device.pin_factory.pin(4)
+    device = InputDevice(4, pull_up=None, active_state=False)
+    pin.drive_high()
+    assert repr(device) == '<gpiozero.InputDevice object on pin GPIO4, pull_up=None, is_active=False>'
+    assert not device.is_active
+    pin.drive_low()
+    assert repr(device) == '<gpiozero.InputDevice object on pin GPIO4, pull_up=None, is_active=True>'
+    assert device.is_active
+
+def test_input_is_active_low_externally_pulled_down():
```
lurch: Was this supposed to be `test_input_is_active_high_externally_pulled_down` ?

lurch:
Looks pretty reasonable (I think - a bit too hot here to think hard ATM!)
Would you mind also adding a test (for the exception) with `pull_up` and `active_state` both set to `None` ?

@bennuttall Presumably the docs will also need to be regenerated before this PR is merged? Or is that a separate step? (I can't remember)

tests/test_inputs.py
pmuller: Fixed in c64aab36afa33c260980a2776f538bfdcdf636f2

pmuller:
Thank you for your review @lurch.

I added the missing test in https://github.com/RPi-Distro/python-gpiozero/pull/658/commits/b845238e2612c1396221c2a15a24be2a76aec53c.

lurch:
Does this also "overlap" with #654 in that this logic should also be applied to other classes too?

pmuller:
@lurch I can rebase on top of #654 or wait for it to be merged then rebase on master. I don't know what's the preference of @bennuttall. Any hint? :)

gpiozero/input_devices.py
```diff
@@ -34,34 +35,64 @@ class InputDevice(GPIODevice):
...
+    :type pull_up: ``bool`` or ``None``
 
     :param Factory pin_factory:
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
+
+    :param active_state:
+        If ``True``, when the hardware pin state is ``HIGH``,
+        the software pin is ``HIGH``.
+        If ``False``, the input polary is reversed:
```
lurch: polary -> polarity ?

gpiozero/input_devices.py
```diff
@@ -34,34 +35,64 @@ class InputDevice(GPIODevice):
...
-    def __init__(self, pin=None, pull_up=False, pin_factory=None):
+    def __init__(self, pin=None, pull_up=False, pin_factory=None,
+                 active_state=None):
         super(InputDevice, self).__init__(pin, pin_factory=pin_factory)
         try:
             self.pin.function = 'input'
-            pull = 'up' if pull_up else 'down'
-            if self.pin.pull != pull:
-                self.pin.pull = pull
+            if pull_up is None:
```
lurch: The code previously did:
```python
if self.pin.pull != pull:
    self.pin.pull = pull
```

I wonder if that logic ought to be maintained, to prevent modifying the pin's pull if not necessary? :man_shrugging:

gpiozero/input_devices.py
```diff
@@ -295,12 +330,26 @@ class Button(HoldMixin, DigitalInputDevice):
     :param Factory pin_factory:
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
+
+    :param active_state:
+        If ``True``, when the hardware pin state is ``HIGH``,
+        the software pin is ``HIGH``.
+        If ``False``, the input polary is reversed:
```
lurch: "polary" again

bennuttall: :+1:

bennuttall:
Sorry for the delay, and all the rebasing. It should probably be merged for the upcoming release. @lurch do you think this is good or needs more changes?

lurch:
Well I left a couple of questions above...

gpiozero/input_devices.py
```diff
@@ -437,18 +486,20 @@ class MotionSensor(SmoothedInputDevice):
     :param bool pull_up:
         If ``False`` (the default), the GPIO pin will be pulled low by default.
         If ``True``, the GPIO pin will be pulled high by the sensor.
+        If ``None``, the pin must be externally be pulled up or down, and
```
lurch: There's probably one 'be' too many on this line ;)

gpiozero/input_devices.py
waveform80: Yup, good idea not to fiddle with pin state unless we absolutely have to. I'll add some tweaks for this.

gpiozero/input_devices.py
```diff
@@ -34,34 +35,64 @@ class InputDevice(GPIODevice):
...
-        self._inactive_state = True if pull_up else False
+
+        if pull_up is None:
+            if active_state is None:
+                raise PinInvalidState(
+                    'Pin %d is defined as floating, but "active_state" is not '
+                    'defined' % self.pin.number)
+            self._active_state = bool(active_state)
+        else:
+            self._active_state = False if pull_up else True
```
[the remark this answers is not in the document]
waveform80: Yup, good idea - I'll fix this up too (just dealing with merge conflicts).

PR #665, opened 2018-09-14
Pin event timing (fixes #664)

### LINKED ISSUE #664: gpiozero DistanceSensor pin_factory errors
Hi,
I get good Readings of the DistanceSensor class used on a pi, but haven't been enable to get Readings from an another Pi using DistanceSensor with pin_factory remotely.
Does anyone have experience with this?
Thanks

waveform80:
Re-works the pins API to use timing information provided by underlying drivers (in particular pigpio) so that timing events from remote pins will be considerably more accurate (also seems to improve the accuracy of some local implementations too, probably because time sampling is done "closer" to the actual event).

### DIFF
--- a/gpiozero/pins/local.py
+++ b/gpiozero/pins/local.py
@@ -104,7 +104,7 @@ class LocalPiPin(PiPin):
             an opaque value that should only be compared with the associated
             :meth:`Factory.ticks_diff` method.
         """
-        super(LocalPiPin, self)._call_when_changed(self._factory.ticks())
+        super(LocalPiPin, self)._call_when_changed(self._factory.ticks(), self.state)
 
 
 class LocalPiHardwareSPI(SPI, Device):
--- a/gpiozero/pins/local.py
+++ b/gpiozero/pins/local.py
@@ -13,6 +13,10 @@ import struct
 import warnings
 from collections import defaultdict
 from threading import Lock
+try:
+    from time import monotonic
+except ImportError:
+    from time import time as monotonic
 
 try:
     from spidev import SpiDev
@@ -72,13 +76,35 @@ class LocalPiFactory(PiFactory):
                         return int(revision, base=16)
         raise PinUnknownPi('unable to locate Pi revision in /proc/cpuinfo or /proc/device-tree')
 
+    def ticks(self):
+        return monotonic()
+
+    def ticks_diff(self, later, earlier):
+        # NOTE: technically the guarantee to always return a positive result
+        # cannot be maintained in versions where monotonic() is not available
+        # and we fall back to time(). However, in that situation we've no
+        # access to a true monotonic source, and no idea how far the clock has
+        # skipped back so this is the best we can do anyway.
+        return later - earlier
+
 
 class LocalPiPin(PiPin):
     """
     Abstract base class representing a multi-function GPIO pin attached to the
     local Raspberry Pi.
     """
-    pass
+    def _call_when_changed(self):
+        """
+        Overridden to provide ticks from the local Pi factory.
+
+        .. warning::
+
+            The local pin factory uses a seconds-based monotonic value for
+            its ticks but you *must not* rely upon this behaviour. Ticks are
+            an opaque value that should only be compared with the associated
+            :meth:`Factory.ticks_diff` method.
+        """
+        super(LocalPiPin, self)._call_when_changed(self._factory.ticks())
 
 
 class LocalPiHardwareSPI(SPI, Device):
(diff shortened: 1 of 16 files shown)


waveform80:
Had a chance to test some more of this today and I'm happy with it, though it'd be great if someone else could have a play to? The codecov fails can be ignored; they're largely down to lots of stuff changing in native which can't be tested on Travis (but the tests pass just fine on a real Pi - sometime in the new year I must set up a permanent gpiozero test bed).

gpiozero/pins/local.py
```diff
@@ -66,13 +70,39 @@ def _get_revision(self):
...
+    def ticks():
+        return monotonic()
+
+    @staticmethod
+    def ticks_diff(later, earlier):
+        # NOTE: technically the guarantee to always return a positive result
+        # cannot be maintained in versions where monotonic() is not available
+        # and we fall back to time(). However, in that situation we've no
+        # access to a true monotonic source, and no idea how far the clock has
+        # skipped back so this is the best we can do anyway.
```
lurch: I wonder if it'd be worth returning `0` or `None` if `earlier > later` ?

Is the clock-skipping mentioned above referring to e.g. DST changes?

gpiozero/pins/native.py
```diff
@@ -72,19 +76,17 @@ def close(self):
         self.mem.close()
         os.close(self.fd)
 
-    def peripheral_base(self):
+    def peripheral_base(self, rev):
```
lurch: Perhaps this parameter should be renamed `cpu` to avoid confusion with the model revision codes?

gpiozero/pins/native.py
```diff
@@ -169,12 +292,23 @@ class NativeFactory(LocalPiFactory):
     """
     def __init__(self):
         super(NativeFactory, self).__init__()
-        self.mem = GPIOMemory()
+        queue = Queue()
+        self.mem = GPIOMemory(self.pi_info.soc)
+        self.fs = GPIOFS(self, queue)
```
lurch: Why does it now create both `self.mem` and `self.fs` when previously it only created `self.mem` ?

gpiozero/pins/local.py
waveform80: Good idea; I'll go with 0 to keep the API simple for now.

gpiozero/pins/native.py
waveform80: Yup, that's a good idea.

gpiozero/pins/native.py
waveform80: Ah - convoluted story. Previously there was GPIOMemory and GPIOFS, but I was only using GPIOMemory because I thought I could get away with just using the GPIO registers. For a rough'n'ready implementation it sort-of worked, although the edge detection was just downright awful. However, it then turned out that in recent kernels, merely attempting to access the edge detection GPIO registers locked things up, so it was time to tidy up GPIOFS properly and start using epoll on the GPIO sysfs files for edge detection instead (which is what RPi.GPIO and RPIO do).

So, GPIOFS did exist previously but it was incomplete and unused. Now it's been tidied up and is in use to handle edge detection for pins (with quite an interesting threading model - I ought to give a talk on the different pin drivers threading models at some point!).

waveform80:
Bah, no sooner do I rebase and I'm behind already! Right, I'm pretty much happy with this at this point; once Travis gives the thumbs up I'm going to have one more play with a variety of hardware then it's merge time. Last call for objections?

bennuttall:
:shipit:

waveform80:
Tested with GPIO inputs, outputs, and SPI under native, RPi.GPIO, and pigpio backends. All looking good. I've no doubt this'll break a few pending PRs though (given the amount it touches). Point me at any you want me to handle merging!

gpiozero/pins/native.py
```diff
@@ -98,52 +100,173 @@ class GPIOFS(object):
...
+                                         os.O_RDONLY | os.O_NONBLOCK)
                     except IOError as e:
                         if e.errno == errno.ENOENT:
-                            with io.open(self.path('export'), 'wb') as f:
+                            with io.open(self.path('export'), 'w+b') as f:
                                 f.write(str(pin).encode('ascii'))
-                            # Pin wasn't exported, so correct the ref-count
-                            self._pin_refs[pin] = 0
                         elif e.errno == errno.EACCES:
                             sleep(i / 100)
```
lurch: Should there be some kind of timeout code here (and also below), to prevent it sleep()-ing forever if/when something goes wrong?

gpiozero/pins/native.py
```diff
@@ -98,52 +100,173 @@ class GPIOFS(object):
...
-                    f.write(str(pin).encode('ascii'))
+            try:
+                os.close(self._exports.pop(pin))
+            except KeyError:
+                # unexport should be idempotent
+                pass
+            else:
+                try:
+                    with io.open(self.path('unexport'), 'w+b') as f:
+                        f.write(str(pin).encode('ascii'))
```
lurch: Is this assuming that by the time the `export` file is writeable, then the `unexport` file must also be writeable at the same time?

Issue #668, opened 2018-09-19
ABC pending depreciation


ricksbt:
/gpiozero/compat.py:60 has a pending depreciation warning. 

ABCs imported from`collections` will cease to work in python 3.8. Import from `collections.abc` instead.

lurch:
I think you mean PR #673 merged? :wink:

bennuttall:
:laughing: yes

Issue #671, opened 2018-09-19
Escape sequence \* in triple quote block deprecated


ricksbt:
Warning on a number of triple quote blocks

DeprecationWarning: invalid escape sequence \*
  """

Note that you only get this if Python is recompiling the .pyc files, it is usually hidden when running the tests.

bennuttall:
PR #671 merged

lurch:
I think you mean PR #673 merged? :wink:

bennuttall:
:laughing: yes

PR #673, opened 2018-09-19
Fixes issues #668 + #671

ricksbt:
Fixes pending depreciation issues. Build of docs tested and checked html and pdf output. Images resulting from doc builds not pushed.


bennuttall:
This is failing for Python 2.7, 3.2, 3.7 and pypy. Do we need to wrap the import in a try/except?

waveform80:
Yup, sorry but `collections.abc` was introduced in Python 3.3. It definitely needs a try..except, probably something like:

```python
try:
    from collections.abc import Mapping
except ImportError:
    from collections import Mapping
```

At the top, and change the class def to `class frozendict(Mapping)`. The rest of the PR looks good though!

ricksbt:
Sorry for delay - have been catching up with other things. Will fix.

PR #677, opened 2018-09-20
Add optional enable init param to Motor, close #366

### LINKED ISSUE #366: Motor with enable control

bennuttall:
An updated version of #367 - need to add tests and docs


gpiozero/output_devices.py
```diff
@@ -837,29 +838,36 @@ class Motor(SourceMixin, CompositeDevice):
...
+    :param int enable:
+        (Optional) The GPIO pin that enables the motor. Required for some motor
+        controller boards. Defaults to ``None``.
+
     :param bool pwm:
         If ``True`` (the default), construct :class:`PWMOutputDevice`
         instances for the motor controller pins, allowing both direction and
         variable speed control. If ``False``, construct
-        :class:`DigitalOutputDevice` instances, allowing only direction
+        :class:`OutputDevice` instances, allowing only direction
```
lurch: Looks like this should have remained `DigitalOutputDevice` rather than being changed to `OutputDevice` ?
(since the code does `PinClass = PWMOutputDevice if pwm else DigitalOutputDevice`)

PR #678, opened 2018-09-21
Allow Robot classes to take pwm argument, close #481

### LINKED ISSUE #481: Robot classes don't allow pwm=False
I'm trying to use MockPin to test a robot script but I can't disable PWM:

``` python
>>> robot = Robot((1, 2), (3, 4), pwm=False)
TypeError: __init__() got an unexpected keyword argument 'pwm'
>>> robot = CamJamKitRobot(pwm=False)
TypeError: __init__() got an unexpected keyword argument 'pwm'
```

I can create motors with `pwm=False`, and I don't believe there's any reason I shouldn't be able to do the same for Robots.

bennuttall:
Tests added, docs updated

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1180,15 +1180,27 @@ class Robot(SourceMixin, CompositeDevice):
         backward inputs of the right motor's controller. Use three pins if your
         motor controller requires an enable pin.
 
+    :param bool pwm:
+        If ``True`` (the default), construct :class:`PWMOutputDevice`
+        instances for the motor controller pins, allowing both direction and
+        variable speed control. If ``False``, construct
+        :class:`DigitalOutputDevice` instances, allowing only direction
+        control.
+
     :param Factory pin_factory:
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
     """
 
-    def __init__(self, left=None, right=None, pin_factory=None):
+    def __init__(self, left=None, right=None, pwm=True, pin_factory=None, *args):
+        # *args is a hack to ensure a useful message is shown when pins are
+        # supplied as sequential positional arguments e.g. 2, 3, 4, 5
+        if type(left) is not tuple or type(right) is not tuple:
+            raise GPIOPinMissing('left and right motor pins must be given as '
+                                 'tuples')
         super(Robot, self).__init__(
-            left_motor=Motor(*left, pin_factory=pin_factory),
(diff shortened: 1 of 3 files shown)


gpiozero/boards.py
```diff
@@ -1391,15 +1417,27 @@ class PhaseEnableRobot(SourceMixin, CompositeDevice):
...
+        :class:`DigitalOutputDevice` instances, allowing only direction
+        control.
+
     :param Factory pin_factory:
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
     """
 
-    def __init__(self, left=None, right=None, pin_factory=None):
+    def __init__(self, left=None, right=None, pwm=True, pin_factory=None, *args):
```
bennuttall, line 1432: Where the hell did `*args` and this comment come from?? I don't remember writing that

@lurch can you think of any reason this couldn't be taken out?

lurch: > I don't remember writing that

:rofl: 
It _looks_ like you added that "args hack" so that `r = Robot(2, 3, 4, 5, pwm=True)` would give you a "friendly" error message of "raise GPIOPinMissing('left and right motor pins must be given as tuples')" rather than the (unfriendly?) error message of "TypeError: Robot() takes 5 positional arguments but 6 were given" ? :shrug: That's just a guess though.

But I dunno precisely how much hand-holding GpioZero should do to protect the user from making coding mistakes, so I've no objection to that being removed :+1:

bennuttall: I don't understand how it would give you a friendly error message 😆

bennuttall: It's not documented! Guess it's safe to be removed

lurch: > I don't understand how it would give you a friendly error message

The way I understand it (which I should emphasise I'm still guessing about) is that with a class-prototype of `def __init__(self, left=None, right=None, pwm=True, pin_factory=None):` if you tried to do `r = Robot(2, 3, 4, 5, pwm=True)`, then python would assign left=2, right=3, pwm=True, pin_factory=4 and then get confused about the left-over 5 parameter, and so emit a TypeError. But with the class-prototype of `def __init__(self, left=None, right=None, pwm=True, pin_factory=None, *args):` I think python would then assign left=2, right=3, pwm=True, pin_factory=4, args=[5] and allow the code to continue on to where you check for `if type(left) is not tuple` ?
But I should qualify that all with a massive \*shrug\* ;)

Issue #679, opened 2018-10-08
Add PumpkinPi board


tuftii:
I'm new to developing for gpiozero and thought I'd add one of @modmypi's [Halloween boards](https://www.modmypi.com/raspberry-pi/led-displays-and-drivers-1034/led-boards-1040/halloween-pumpkin-solder-kit) as a starter.

I've got a handful of files at a fork [here](https://github.com/tuftii/python-gpiozero) if you'd like to see where I've got to.

Things I was unsure of:
- Testing, is there a document explaining how to write tests or is it just a matter of copying existing tests?
- Building docs, I've put some code in the docs for this board, but wasn't sure how to build it.
- If this is an appropriate thing to add into gpiozero.

Is there anything else I should do? Happy for comments and pointers :)

bennuttall:
This is awesome, thanks @tuftii.

> Testing, is there a document explaining how to write tests or is it just a matter of copying existing tests?

Not really - you don't need to copy the `LEDBoard` tests, but it might make sense to copy, say `SnowPi` or `StatusBoard` and test the general behaviour. Which pins should be high after `pumpk.left.on()` - that kind of thing. Be sure to test any edge cases.

> Building docs, I've put some code in the docs for this board, but wasn't sure how to build it.

The docs for docs are here: https://gpiozero.readthedocs.io/en/latest/development.html#building-the-docs

Does that make sense? They're pretty tied to Ubuntu, mind. Give me a shout if you have any issues.

> If this is an appropriate thing to add into gpiozero.

Yep :+1: 

> Is there anything else I should do? Happy for comments and pointers :)

Go ahead and open a PR. I'll review it (it looks fine) and merge once any minor issues resolved, and if the tests pass on travis (the PR will kick off a build).

bennuttall:
The only problem is I don't see us shipping v1.5 before halloween - so you might have to bundle a file with this class it in

tuftii:
Great, thanks Ben!

> They're pretty tied to Ubuntu, mind.

I'll have a go on my Ubuntu machine later, I'm on Windoze currently.

> I don't see us shipping v1.5 before halloween

No problem, I put a little class together outside of gpiozero for testing this morning, so I'll add that into the MMP repo temporarily, then I can remove it when 1.5 ships.

Tests - I'll add those in today/tomorrow and try to run them too.

PR #680, opened 2018-10-08
Add PumpkinPi board

tuftii:
First PR for PumpkinPi as mentionned in issue #679.

### DIFF
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -93,6 +93,7 @@ from .boards import (
     PhaseEnableRobot,
     PololuDRV8835Robot,
     Energenie,
+    PumpkinPi,
 )
 from .other_devices import (
     InternalDevice,
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1684,3 +1684,63 @@ class Energenie(SourceMixin, Device):
         Turns the socket off.
         """
         self.value = False
+        
+class PumpkinPi(LEDBoard):
+    """
+    Extends :class: `LEDBoard` for the `ModMyPi Pumpkin Pi` board.
+
+    There are tweleve LEDs connected up to individual pins, so for 
+    the PumpkinPi the pins are fixed.
+
+    Usage:
+        from gpiozero import PumpkinPi
+
+        pumpk = PumpkinPi(pwm=True)
+        pumpk.sides.pulse()
+        pumpk.off()
+
+    :param bool pwm:
+        If ``True``, construct :class:`PWMLED` instances to represent each
+        LED. If ``False`` (the default), construct regular :class:`LED`
+        instances
+
+    :param bool initial_value:
+        If ``False`` (the default), all LEDs will be off initially. If
+        ``None``, each device will be left in whatever state the pin is found
+        in when configured for output (warning: this can be on). If ``True``,
+        the device will be switched on initially.
+
+    :param Factory pin_factory:
+        See :doc:`api_pins` for more information (this is an advanced feature
+        which most users can ignore).
+        
+    .. _ModMyPi PumpkinPi: https://www.modmypi.com/halloween-pumpkin-solder-kit
+
+    """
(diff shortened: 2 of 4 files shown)


gpiozero/boards.py
```diff
@@ -1684,3 +1684,63 @@ def off(self):
...
+    .. _ModMyPi PumpkinPi: https://www.modmypi.com/halloween-pumpkin-solder-kit
+
+    """
+    def __init__(self, pwm=False, initial_value=False, pin_factory=None):
+        super(PumpkinPi, self).__init__(
+            sides=LEDBoard(
+                left=LEDBoard(
+                    bottom=18, midbottom=17, middle=16, midtop=13, top=24,
+                    pwm=pwm, initial_value=initial_value,
+                    _order=('bottom','midbottom','middle','midtop','top'),
```
bennuttall: Can we get a space after each comma in this tuple plz? (sorry)

bennuttall: (same in other places)

gpiozero/__init__.py
```diff
@@ -93,6 +93,7 @@
     PhaseEnableRobot,
     PololuDRV8835Robot,
     Energenie,
+    PumpkinPi,
```
bennuttall: Tests please

bennuttall:
Forgot to say: just push more commits to the same branch and they'll be picked up here

tuftii:
Yeah you're quite right @lurch, sorry I was updating both boards web pages yesterday and copied the wrong link.

tuftii:
Ok, so I've added 3 tests for the PumpkinPi board. I came across something a bit weird.

I'm now setup with a development environment on Ubuntu using the mock pin environment variables. The first two tests (`test_pumpkin_pi` and `test_pumpkin_pi_initial_value`) run without issue.

The third `test_pumpkin_pi_initial_value_pwm` (which I've commented out for now) runs and passes, but causes `test_input_light_sensor`, `test_input_distance_sensor`, `test_mock_pin_frequency_unsupported` and `test_mock_pin_state` to fail. 

I'm a bit puzzled as to why. I've copied the test from the SnowPi pwm test and changed the pins to the PumpkinPi pins, but couldn't work out what would cause these other tests to fail?

gpiozero/__init__.py
tuftii: Tests added :)

bennuttall:
I've uncommented it to see what happens (tests will automatically run on travis). If they're failing on there we can look and see why. If they're passing, but not on your machine, let's look into that.

bennuttall:
Ok got it. Pins 4 and 5 have been set up elsewhere with different functions and I guess the pin objects not closed / reset. If you can figure out how to do that, go for it. Otherwise @waveform80 or I will look when we get chance.

> `ValueError: pin 5 is already in use as a MockTriggerPin`

> `ValueError: pin 4 is already in use as a MockChargingPin`

Thank for the effort :+1:

gpiozero/boards.py
```diff
@@ -1684,3 +1684,63 @@ def off(self):
         Turns the socket off.
         """
         self.value = False
+        
+class PumpkinPi(LEDBoard):
+    """
+    Extends :class: `LEDBoard` for the `ModMyPi Pumpkin Pi` board.
+
+    There are tweleve LEDs connected up to individual pins, so for
```
lurch: tweleve -> twelve

gpiozero/boards.py
```diff
@@ -1684,3 +1684,63 @@ def off(self):
         Turns the socket off.
         """
         self.value = False
+        
+class PumpkinPi(LEDBoard):
```
lurch: Should be 2 blank lines before a new class

tuftii:
Thanks @lurch for the fix! :+1:

lurch:
@bennuttall / @waveform80 Did my 'fix' above to the MockFactory `pin_class` stuff get picked up? Or an equivalent fix committed?

PR #682, opened 2018-10-14
corrects typo for method value

SteveAmor:
Method is value not values :-)


bennuttall:
No it's not :)

https://gpiozero.readthedocs.io/en/stable/source_values.html

SteveAmor:
I did wonder...

PR #698, opened 2018-11-26
Allow source tools to take device object as well as device.values or just values, close #640

### LINKED ISSUE #640: Add connect method to devices with source
I really like how simple [source/values](https://gpiozero.readthedocs.io/en/stable/source_values.html) makes connecting up devices in one line:

```python
led.source = btn.values
```

However, a combination of the magic behind it and the naming convention makes me consider it an advanced feature I wouldn't really teach to kids or beginners.

If we had a way that was as simple as this, but seemed more beginner friendly and approachable, that would be great. This is oddly something I considered in a gpiozero pre-release (before source/values), but something I've reconsidered lately. Here's what I have in mind:

```python
led.connect(btn)
```

which would set the LED's source to the button's values, as above. It could be implemented in `SourceMixin`. We could possibly inspect the "shape" of the value of the two devices to make sure they're compatible.

bennuttall:
Addresses #640 with a different approach:

- Allow source to be set to device object as well as device.values or just values
- Allow source tools to take device object as well as device.values or just values

With this change, you can do:

```python
led.source = btn
led.source = negated(btn)
led.source = all_values(btn1, btn2)
motor.source = scaled(pot, -1, 1)
led.source = booleanized(pot, 0.25, 0.75)
led.source = multiplied(pot1, pot2, pot3)
garden.source = all_values(booleanized(light, 0, 0.1), motion)
```

instead of:

```python
led.source = btn.values
led.source = negated(btn.values)
led.source = all_values(btn1.values, btn2.values)
motor.source = scaled(pot.values, -1, 1)
led.source = booleanized(pot.values, 0.25, 0.75)
led.source = multiplied(pot1.values, pot2.values, pot3.values)
garden.source = all_values(booleanized(light.values, 0, 0.1), motion.values)
```

Documentation branch: https://gpiozero.readthedocs.io/en/connect-devices/

Addresses #640

### DIFF
--- a/docs/examples/button_led_2.py
+++ b/docs/examples/button_led_2.py
@@ -4,6 +4,6 @@ from signal import pause
 led = LED(17)
 button = Button(2)
 
-led.source = button.values
+led.source = button
 
 pause()
--- a/docs/examples/combining_sources.py
+++ b/docs/examples/combining_sources.py
@@ -6,6 +6,6 @@ button_a = Button(2)
 button_b = Button(3)
 led = LED(17)
 
-led.source = all_values(button_a.values, button_b.values)
+led.source = all_values(button_a, button_b)
 
 pause()
--- a/docs/examples/cpu_temperature_bar_graph.py
+++ b/docs/examples/cpu_temperature_bar_graph.py
@@ -4,6 +4,6 @@ from signal import pause
 cpu = CPUTemperature(min_temp=50, max_temp=90)
 leds = LEDBarGraph(2, 3, 4, 5, 6, 7, 8, pwm=True)
 
-leds.source = cpu.values
+leds.source = cpu
 
 pause()
--- a/docs/examples/custom_generator_finite.py
+++ b/docs/examples/custom_generator_finite.py
@@ -2,6 +2,7 @@ from gpiozero import LED
 from signal import pause
 
 led = LED(17)
+led.source_delay = 1
 led.source = [1, 0, 1, 1, 1, 0, 0, 1, 0, 1]
 
 pause()
--- a/docs/examples/internet_status_indicator.py
+++ b/docs/examples/internet_status_indicator.py
@@ -7,8 +7,8 @@ red = LED(18)
 
 google = PingServer('google.com')
 
-green.source = google.values
+green.source = google
 green.source_delay = 60
-red.source = negated(green.values)
+red.source = negated(green)
 
 pause()
--- a/docs/examples/led_button.py
+++ b/docs/examples/led_button.py
@@ -4,6 +4,6 @@ from signal import pause
 led = LED(17)
 button = Button(2)
 
-led.source = button.values
+led.source = button
 
 pause()
--- a/docs/examples/led_button_remote_1.py
+++ b/docs/examples/led_button_remote_1.py
@@ -7,6 +7,6 @@ factory = PiGPIOFactory(host='192.168.1.3')
 button = Button(2)
 led = LED(17, pin_factory=factory)
 
-led.source = button.values
+led.source = button
 
 pause()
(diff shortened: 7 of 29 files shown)


bennuttall:
Needs docs changes

bennuttall:
@waveform80 do you think these tests are reasonable? 701f3223a7187daf3b15db33d8a2abf6b82ee2b6

bennuttall:
It might be worth adding a `zip_values` source tool so that you can replace:

```python
led.source = zip(red_pot.values, green_pot.values, blue_pot.values)
```

with:

```python
led.source = zip_values(red_pot, green_pot, blue_pot)
```

bennuttall:
Pushed docs changes, and they're live on a branch on RTD: https://gpiozero.readthedocs.io/en/connect-devices/source_values.html (and other pages)

I've changed all examples and docstrings where possible. Note the previous comment re: `zip` examples.

I think I'll swap out the diagrams on the source/values page to remove the attributes list, and just have boxes with device names in e.g. LED -> Button. Can probably do it with graphviz too.

bennuttall:
Ping @martinohanlon

bennuttall:
This is BC

bennuttall:
See docs for this at https://gpiozero.readthedocs.io/en/latest/source_values.html - there's a note that says:

> Although this method is still supported, the recommended way is now to set the source to a device object.

SrMouraSilva:
I apologize. I looked at the merge request and did not realize that there would still be support for this mode.

SrMouraSilva:
@bennuttall

For me, `led.source = button` make more sense than `led.value = button.value` and `led.source = pot.values` :)

lurch:
@SrMouraSilva It's a long time since I've actually written (or even run) any GpioZero code, but IIRC
```python
led.value = button.value
```
is a 'one-shot' process, which reads the **current** value of the button, and sets the value of the LED to that value, i.e. it behaves similarly to:
```python
old_value = button.value
led.value = old_value
```

Whereas:
```python
led.source = button.values
```
or
```python
led.source = button
```
is a 'continuous' process, where the value of the LED is continuously being updated (via a background thread) to the value of the button, i.e. it behaves similarly to:
```python
while True:
    old_value = button.value
    led.value = old_value
```

Apologies if you knew that already, but doesn't hurt to clarify :wink:

lurch:
@bennuttall @waveform80 

Slightly OT, but I noticed this example in the docs:
```python
from gpiozero import LED, Button
from signal import pause

red = LED(14)
green = LED(15)
button = Button(17)

red.source = button
green.source = red

pause()
```

I don't have time to test right now, but if that example was slightly tweaked to:
```python
from gpiozero import LED, Button
from signal import pause

red = LED(14)
green = LED(15)
button = Button(17)

red.source = button
green.source = button

pause()
```
does that mean that `red.value` and `green.value` will be set to _alternate_ values read from `button.values` ? Obviously not an issue for this simple example, but might be an issue for some of the source-tools or other custom generators? Is that "defined behaviour" and the user simply needs to be careful with what he's doing?
Will `red.source = button; green.source = red` always be the best way to do things, or might there be value in a built-in sourcetool that allowed you to simultaneously set multiple devices to the same source?

i.e. AIUI `red.source = button; green.source = red` effectively does:
```python
while True:
    old_value1 = button.value
    red.value = old_value1
    old_value2 = red.value
    green.value = old_value2
```
whereas I'm wondering if there might be value in a (new) method which allowed you to effectively do:
```python
while True:
    old_value = button.value
    red.value = old_value1
    green.value = old_value1
```
?  (got a feeling I might not be explaining myself very well!)

SrMouraSilva:
> `led.value = button.value` 

@lurch, my mistake! You are right `led.value = button.value` is other think.
@bennuttall I do not even remember. I think that's what you corrected.

PR #716, opened 2019-01-23
Fix ZeroDivisionError in CPUTemperature class

jee1mr:
- Check for min < max
- Add tests

### DIFF
--- a/gpiozero/other_devices.py
+++ b/gpiozero/other_devices.py
@@ -122,7 +122,6 @@ class CPUTemperature(InternalDevice):
             raise ValueError('min_temp must be less than max_temp')
         self.min_temp = min_temp
         self.max_temp = max_temp
-        self._temp_range = self.max_temp - self.min_temp
         self.threshold = threshold
         self._fire_events()
 
@@ -145,7 +144,8 @@ class CPUTemperature(InternalDevice):
         *max_temp* value). These default to 0.0 and 100.0 respectively, hence
         :attr:`value` is :attr:`temperature` divided by 100 by default.
         """
-        return (self.temperature - self.min_temp) / self._temp_range
+        temp_range = self.max_temp - self.min_temp
+        return (self.temperature - self.min_temp) / temp_range
 
     @property
     def is_active(self):


bennuttall:
Thanks for this. I started working on it last night - I started with what you did, then realised the check should be in the setter, as we support changing it later. But then what happens when you do:

```python
>>> tod = TimeOfDay(time(8), time(10))
>>> tod.start_time = time(10)  # error here
>>> tod.end_time = time(12)
```

You would have to set `end_time` first.

So @waveform80 suggested we make the setter take a tuple:

```python
>>> tod = TimeOfDay(time(8), time(10))
>>> tod.active_period = (time(10), time(12)) # not sure about property name
```

However we'd have to raise a deprecation error on setting one of the old properties.

@waveform80 what should we do for 1.5?

jee1mr:
Ah! Right. I didn't think of that.
I agree with @waveform80's suggestion. I couldn't think of any better alternative.
Let me know if you need my help or if you're going to take it up.

gpiozero/other_devices.py
```diff
@@ -118,8 +118,11 @@ def __init__(self, sensor_file='/sys/class/thermal/thermal_zone0/temp',
             min_temp=0.0, max_temp=100.0, threshold=80.0):
         self.sensor_file = sensor_file
         super(CPUTemperature, self).__init__()
+        if min_temp >= max_temp:
+            raise ValueError('min_temp must be less than max_temp')
         self.min_temp = min_temp
         self.max_temp = max_temp
+        self._temp_range = self.max_temp - self.min_temp
```
bennuttall: Can't do this - it doesn't support changing the values after init. This could be a property instead but that's overkill I think.

bennuttall:
I'm not sure why. We definitely need more tests for those devices though. I'll push up what I did last night shortly.

gpiozero/other_devices.py
jee1mr: Right.
I actually took it from here:
`        self._dc_range = (max_pulse_width - min_pulse_width) / frame_width

    @property
    def max_pulse_width(self):
        """
        The control pulse width corresponding to the servo's maximum position,
        measured in seconds.
        """
        return (self._dc_range * self.frame_width) + self.min_pulse_width
`

jee1mr: Right. I actually took it from here

Servo init:
`self._dc_range = (max_pulse_width - min_pulse_width) / frame_width`
  max_pulse_width property:
`return (self._dc_range * self.frame_width) + self.min_pulse_width`

jee1mr: But that's a read-only property there.

lurch:
Perhaps it's also worth checking that `threshold` is between `min_temp` and `max_temp` ?

> However we'd have to raise a deprecation error on setting one of the old properties.

Not sure I agree with that - IMHO there'd be nothing wrong with doing:
```python
>>> tod = TimeOfDay(time(8), time(10))
>>> tod.end_time = time(11)
```
although I agree that adding a setter (and possibly a matching getter too?) to set both values at once wouldn't hurt.

bennuttall:
> Why wouldn't CPUTemperature initialization run on the build machine?
> Is "/sys/class/thermal/thermal_zone0/tem" specific to raspberrypi?

It's not specific to Rapsberry Pi, that's usually where the temperature can be read in Linux. I'm not sure what would happen in Travis, but I was planning to add a local file to the tests directory with a sample temperature reading in it. @waveform80

bennuttall:
@jee1mr How do you want your name displayed in the list of contributors?

jee1mr:
@bennutttall
"Jeevan M R" would be good. Thanks :D

lurch:
I guess people missed my "Perhaps it's also worth checking that threshold is between min_temp and max_temp ?" comment? :-/

bennuttall:
Is that absolutely necessary?

```python
cpu = CPUTemperature(min_temp=10, max_temp=50)

something.source = cpu.values
```

Here, `threshold` is at the default value of `80` which is outside my value range, but I don't care as I'm not using `threshold`. The device will always show as inactive, but I'm not using `is_active` for anything.

waveform80:
I think I'd at least like a warning that the threshold is outside the specified range? i.e. shouldn't crash the script but it might save someone the headache of wondering why the active/inactive events *never* fire for them.

bennuttall:
pfft

bennuttall:
Done! It should probably be added elsewhere. *checks* - just `LoadAverage`. Possibly elsewhere outside "internal devices"...

bennuttall:
Done

Issue #681, opened 2018-10-10
Add Jam HAT to gpiozero


tuftii:
After chatting with @bennuttall, I've opened up an issue to possibly add [Jam HAT](https://www.modmypi.com/jam-hat) to GPIO Zero.

The board is 6 LEDs, a KC1206 buzzer and two buttons and looks like this:
![jam-hat](https://www.modmypi.com/image/cache/catalog/rpi-products/MMP-1150/DSC_1129-747x569.jpg)

Ben suggested any one of the nested options below to allow control over the lights together:

((red, amber, green), (red, amber, green), button, button, buzzer)
(((red, amber, green), (red, amber, green)), button, button, buzzer)
(((red, amber, green), (red, amber, green)), (button, button), buzzer)
((red, amber, green), (red, amber, green), (button, button), buzzer)
((red, red), (amber, amber), (green, green), (button, button), buzzer)
(red, amber, green, red, amber, green, button, button, buzzer)

(there are more)

The buzzer is 5V so I've been playing with it using the PWMOutputDevice class to try to get notes from it rather than a buzz, but the notes aren't that clean. Here's a short sample of what I'm doing, is there something different I could be doing in setup?

```python
from gpiozero import PWMOutputDevice, Button
from time import sleep

pwm = PWMOutputDevice(20, frequency=2048)
button = Button(19, False)

try:
    while True:
        if button.is_pressed:
            pwm.on()
        else:
            pwm.off()
        sleep(0.1)
except:
    pwm.off()
    pwm.close()
    button.close()
```

@modmypi are producing the board with the York Pi Jam group and some basic examples using RPi.GPIO live [here](https://github.com/modmypi/Jam-HAT). It's available on the website, but we haven't done any publicity around it yet or polished the webpages, I was hoping to add it to GPIO Zero and create some more Jam friendly examples and worksheets to go along with it before we did anything more.

Over the next couple of days I'm going to fork the ModMyPi repo and change the examples over to GPIO Zero just to see what can be done. Any suggestions or pointers as to best approaches for the wider class would be ace :)

bennuttall:
It would be great to have a class for this, we just need to decide the format of the composite device.

I think if I were to choose now, I'd go for: ((red, amber, green), (red, amber, green), button, button, buzzer)

Therefore, the API would be:

```python
hat = JamHat()
hat.lights_1.on()
hat.lights_2.on()
hat.button_1.is_pressed
hat.button_2.is_pressed
[...]

PR #717, opened 2019-01-23
Add JamHat and TonalBuzzer

tuftii:
Here's my first stab at getting the core code in place. It's not complete PR yet, but there's quite a lot of discussion going on in issue #681 so I thought I'd upload what I've got so far so everyone can see what we've got.

Features:
- New PWMBuzzer class for controlling the tonal buzzer, two "note" dictionaries (one piano, one MIDI) and a .play(Hz) function which plays the note.
- New JamHat class.

To Do:
- Implement a sensible way of setting .value in the PWMBuzzer class.
- Documentation
- Testing.

I've come up against a bug which fails on locating the pin factory (I assume this is a setup issue at my end, most likely).

Any other help/comments based on what I've got so far would be much appreciated!


bennuttall:
This doesn't look right. The diff just shows you have added 3 files, it doesn't show what you have added - I'd expect to see just the two class additions, but it's showing every single line as new. How did you manage that?

tuftii:
I have no idea... It's been one of those weeks where *everything* has gone wrong. Shall I kill this off and start again?

gpiozero/output_devices.py
```diff
@@ -503,6 +503,243 @@ def _blink_device(
...
+            'AS7': 3729.31,
+            'B7': 3951.07,
+            'C8': 4186.01,
+        }
+
+        self.midi_notes = {
+            """
+            Dictionary containing MIDI notes which can be played through the PWMBuzzer.
+            """
+            '21': 27.50,
```
lurch: In order for the `play` function to work for midi notes, I think you'll need to remove the quotes from the keys in this dictionary, i.e. `21: 27.50,` etc.

gpiozero/boards.py
```diff
@@ -1686,3 +1687,52 @@ def off(self):
...
+    :param bool pwm:
+    If ``True``, construct :class: PWMLED instances to represent each LED 
+    on the board. If ``False`` (the default), construct regular :class: 
+    LED instances.
+
+    :param bool initial_value:
+    If ``False`` (the default), all LEDs will be off initially. If
+    ``None``, each device will be left in whatever state the pin is found
+    in when configured for output (warning: this can be on). If ``True``,
+    the device will be switched on initially.
```
lurch: Should the initial_value also affect the buzzer?

gpiozero/output_devices.py
```diff
@@ -503,6 +503,243 @@ def _blink_device(
             self._write(value)
             if self._blink_thread.stopping.wait(delay):
                 break
+            
+class PWMBuzzer(PWMOutputDevice):
+    """
+    Extends :class: `PWMOutputDevice` and represents a piezo or tonal buzzer 
+    class.
```
lurch: Isn't this only for tonal buzzers? AIUI piezo buzzers constantly beep whenever a fixed voltage is applied to them?

gpiozero/output_devices.py
tuftii: Removed the quotes around the MIDI notes.

gpiozero/boards.py
tuftii: I guess it should (despite my increasing hatred of the gurgly buzzers being on at all :laughing:)

bennuttall: But to what note?

gpiozero/output_devices.py
```diff
@@ -503,6 +503,243 @@ def _blink_device(
             self._write(value)
             if self._blink_thread.stopping.wait(delay):
                 break
+            
+class PWMBuzzer(PWMOutputDevice):
```
bennuttall: Should this be called `PWMBuzzer` or `TonalBuzzer`?

`PWMLED` works with normal LEDs but this class doesn't work with normal buzzers.

gpiozero/output_devices.py
```diff
@@ -503,6 +503,243 @@ def _blink_device(
...
+
+    def __init__(
+            self, pin=None, active_high=True, initial_value=0, frequency=100,
+            pin_factory=None):
+        super(PWMBuzzer, self).__init__(
+            pin=pin, active_high=active_high, initial_value=initial_value, frequency=frequency,
+            pin_factory=pin_factory
+        )
+
+        self.notes = {
```
bennuttall: Part of me wants these dicts to be declared outside of the file. What do you think @waveform80 @lurch?

gpiozero/output_devices.py
```diff
@@ -503,6 +503,243 @@ def _blink_device(
...
+            104: 3322.44,
+            105: 3520.0,
+            106: 3729.31,
+            107: 3951.07,
+            108: 4186.01,
+        }
+
+        def play(self, note):
+            """
+            PWMBuzzer function to play a given note from the two available dictionaries.
```
bennuttall: It's a method not a function - but you can leave this part out. Start with "play a given note", specify the input options and ranges instead of "from the available dicts", and give two examples of usage. See other docstrings for format.

gpiozero/output_devices.py
lurch: Something like a `notes_data.py`, which `output_devices.py` then imports?

bennuttall: That's what I was thinking

gpiozero/output_devices.py
lurch: Of course the `play` method disappears if we instead decide that setting a non-None value always turns the buzzer on.

gpiozero/boards.py
lurch: Given that a Servo has range -1 to +1, and has an initial value of 0, perhaps the default initial value of a TonalBuzzer that ranges from 0 to 1 should be 0.5 ?

gpiozero/boards.py
```diff
@@ -1686,3 +1687,52 @@ def off(self):
...
+
+    :param Factory pin_factory:
+    See :doc:`api_pins` for more information (this is an advanced feature
+    which most users can ignore).
+
+    .. _ModMyPi JamHat: https://www.modmypi.com/jam-hat
+
+    """
+    def __init__(self, pwm=False, initial_value=False, pin_factory=None):
```
lurch: Perhaps the `__init__` for JamHat should also have an option for the frequency to get passed into the buzzer's init?

gpiozero/boards.py
bennuttall: `TrafficHat` doesn't have an initial value...

gpiozero/boards.py
```diff
@@ -1686,3 +1687,52 @@ def off(self):
...
+
+    Usage:
+        from gpiozero import JamHat
+
+        hat = JamHat()
+        hat.lights_1.on()
+        hat.buzzer.play(2000)
+        hat.off()
+
+    :param bool pwm:
```
lurch: Given that the JamHat has more than just LEDs, but this argument applies only to the LEDs, I wonder if something like `pwm_leds` might be a better name? :man_shrugging:

bennuttall: It's tricky. Though LEDs are the only variable device. I think `pwm` is fine and the docstring is clear enough.

gpiozero/boards.py
bennuttall: I think that's complicated, and it's easy enough to set immediately after.

lurch: ...unless you want to turn on the buzzer initially to a specific frequency (i.e. `initial_value=True`) without having an audible glitch :wink:

gpiozero/boards.py
lurch: True - TrafficLightsBuzzer doesn't have an `initial_value` so maybe JamHat shouldn't either.

tuftii: OK, so I'm going to leave this and not add an `initial_value` for JamHat.

bennuttall: I think that's a good call. I wonder if we should remove `initial_value` from composite devices other than `LEDBoard`. a) it gets complicated b) we only really turn LEDs on c) we waste time discussing what it should do every time

gpiozero/output_devices.py
tuftii: I'm happy to go with `TonalBuzzer`. 

As an alternative, would `PWMTonalBuzzer` be acceptable? Just to make it clear that's the type of behaviour we should expect from this class?

tuftii:
The two base classes now work on the JamHAT, the last commit was me (I've no idea how Git has assigned it to someone else?! I'm having a right hassle with it!).

For some reason, the comment at the top of `self.midi_notes` was causing the dictionary to have a syntax error, but I couldn't see anything weird with it.

I'm going to move out these two dictionaries to the notes_data.py as suggested above. So I'll add comments back in then.

gpiozero/output_devices.py
tuftii: OK, I've moved the two dicts into `notes_data.py` and added the import to `output_devices.py`

gpiozero/output_devices.py
```diff
@@ -503,6 +504,53 @@ def _blink_device(
...
+    """
+    Extends :class: `PWMOutputDevice` and represents a piezo or tonal buzzer 
+    class.
+
+    :param: int pin:
+        The GPIO pin which the buzzer is attached to. See :ref:`pin-numbering` for
+        valid pin numbers.
+    
+    :param bool active_high:
+        If ``True`` (the default), the :meth:`on` method will set the GPIO to
```
lurch: Hmmm, if the buzzer functions by means of pwm, do we even need to specify whether it's active-high or active-low?

gpiozero/boards.py
bennuttall: Which classes (other than LEDBoards and LEDCollections) have `initial_value`?

lurch:
@tufti Seems you forgot to add `notes_data.py` ?

tuftii:
Added now, thanks @lurch.

gpiozero/boards.py
lurch: Hmmm, looks like you're right - I must have been getting confused with something else, sorry!

gpiozero/output_devices.py
bennuttall: I'd go for `TonalBuzzer`

gpiozero/output_devices.py
bennuttall: Good point. We can get rid of that option.

bennuttall:
I have been playing with this and now I see the light. We don't want to subclass `PWMOutputDevice` at all. Look at `Servo`. We want to subclass `CompositeDevice` (and `SourceMixin`) - and just create a pwm device within it. We don't want on/off/blink/etc. We want to define our own properties and methods, and disguise the way value converts 0-1 into a note.

I'll try to push something up when I've got it working.

bennuttall:
Ok I've pretty much made it work this way. I think this is on the right track. I've not pushed here to overwrite your work, I've pushed to another branch.

I wrote some tests for `TonalBuzzer` (and they're all passing): https://github.com/RPi-Distro/python-gpiozero/blob/jamhat/tests/test_outputs.py#L1509:L1565

You can see the new implementation here: https://github.com/RPi-Distro/python-gpiozero/blob/jamhat/gpiozero/output_devices.py#L508

One other thing I'd like to do is create a method to convert a note to a value (which depends on the object's min-max value range), so that you can set source to a series of notes:

```python
tune = ('G5', 'G5', 'A5', 'G5', 'C6', 'B5', None,
        'G5', 'G5', 'A5', 'G5', 'D6', 'C6', None, 
        'G5', 'G5', 'GS6', 'E6', 'C6', 'B5', 'A5', None,
        'F6', 'F6', 'E6', 'C6', 'D6', 'C6', None)
bz.source_delay = 0.5
bz.source = (bz.note_value(note) for note in tune)
```

bennuttall:
I've merged it now, and added docs for both classes, which are viewable here:
https://gpiozero.readthedocs.io/en/jamhat/api_output.html#tonalbuzzer
https://gpiozero.readthedocs.io/en/jamhat/api_boards.html#jam-hat

Still need tests for the JamHat. And I think we need to implement `off()` which also calls `buzzer.stop()` as well as `super()`.

I don't think the source setting example works yet, I need to take another look.

bennuttall:
Well, it didn't push to this branch for some reason. It's here: https://github.com/RPi-Distro/python-gpiozero/tree/jamhat

gpiozero/notes_data.py
```diff
@@ -1,7 +1,4 @@
 notes = {
-    """
```
lurch: Or you could have moved the comments above the `notes = {` line, instead of deleting them entirely :wink:

bennuttall: I didn't think they were necessary :P

waveform80:
Merged after rebase in 0baa34ea1c05837956bc455d75faf02cda9eb0e1

PR #718, opened 2019-01-24
Add tests for internal devices

bennuttall:


gpiozero/internal_devices.py
```diff
@@ -279,45 +286,40 @@ def __init__(self, start_time, end_time, utc=True):
...
+        self._utc = utc
         self._fire_events(self.pin_factory.ticks(), None)
 
     def __repr__(self):
         return '<gpiozero.TimeOfDay active between %s and %s %s>' % (
                 self.start_time, self.end_time, ('local', 'UTC')[self.utc])
 
+    def _validate_time(self, value):
+        if isinstance(value, datetime):
+            print("dt", value)
```
lurch: Leftover debug print?

gpiozero/internal_devices.py
```diff
@@ -279,45 +286,40 @@ def __init__(self, start_time, end_time, utc=True):
...
-        self._start_time = value
-
     @property
     def end_time(self):
         """
         The time of day after which the device will be considered inactive.
         """
         return self._end_time
 
-    @end_time.setter
```
lurch: With the removal of the `start_time`, `end_time` and `utc` setters, I guess TimeOfDay objects are now effectively "readonly"?

bennuttall: Yes - decided it was easier than validating changes.

gpiozero/internal_devices.py
bennuttall: removed, thanks

lurch:
Did the internal-devices ever get fixed up to run a background thread, so that e.g. the when_active / when_inactive events get fired as appropriate?

bennuttall:
Not yet!

gpiozero/internal_devices.py
lurch: I guess the only validation needed (in this case) is that start_time != end_time? :man_shrugging: 

Nice catch with the tests that work with the 'active' period crossing the midnight boundary BTW.

bennuttall: https://github.com/RPi-Distro/python-gpiozero/blob/b8b7a49d782af56f5f153a187826da3cd2e38a59/gpiozero/internal_devices.py#L289

lurch: Yeah I saw that, I was just wondering why you couldn't do that in the setters too?

Although now that I think about it, I guess you ought to convert both the start_time and end_time parameters to `time` instances, _before_ checking if they're equal, in case somebody passes in a `time` for one parameter and a `datetime` for another parameter? (and add unit-tests for this too :slightly_smiling_face: )

gpiozero/exc.py
```diff
@@ -166,3 +166,5 @@ class PinFactoryFallback(PinWarning):
 class PinNonPhysical(PinWarning):
     "Warning raised when a non-physical pin is specified in a constructor"
 
+class ThresholdOutOfRange(GPIOZeroWarning):
```
lurch: Hmmm, interesting. Should the `ValueError`s that you're raising elsewhere in this PR also be subclasses of `GPIOZeroError`? ping @waveform80

gpiozero/internal_devices.py
bennuttall: Added tests and fixed!

gpiozero/exc.py
bennuttall: Probably - yes. I have a feeling there are more non-gpiozero exceptions lying around, which we should clean up.

Issue #675, opened 2018-09-20
Performance issues with pkg_resources


BuddhaOhneHals:
I want to use your library in one of my projects with a Raspberry Pi Zero. Until now I used RPi.GPIO but your API seems just great. But one issue that bothers me is the duration when I'm importing the gpiozero module.
I use a freshly installed raspbian lite (`2018-06-27-raspbian-stretch-lite`) with the gpiozero package installed via `apt`. The import of gpiozero in the python shell needs 5 seconds or more.
```
$ python -c "from timeit import default_timer as t;s=t();import gpiozero;print(t()-s)"
5.40878105164
```
It seems like the import of the `pkg_resources` package in the `devices.py` file is responsible for this:
```
$ python -c "from timeit import default_timer as t;s=t();import pkg_resources;print(t()-s)"
4.59635186195
```
It would be really great and helpful if there's a possibility to skip the importing of the `pkg_resources` package.

bennuttall:
My first impression is that surely the bottleneck isn't `pkg_resources` but just how big gpiozero is.

We've noticed long import delays, especially on Pi 1/Zero. It's purely because there are so many classes provided in the module's init file. We've considered moving imports out of here but that would change the way imports work: `from gpiozero import LED` would become `from gpiozero.outputs import LED`.

You can do this now, using the filenames the classes are in. Can you try `from gpiozero.output_devices import LED` (or similar - for the stuff you're importing - just [look at the files they're in](https://github.com/RPi-Distro/python-gpiozero/tree/master/gpiozero)) - and see if that makes it any faster?

BuddhaOhneHals:
Unfortunately no differences:
```
$ python -c "from timeit import default_timer as t;s=t();from gpiozero.output_devices import LED;print(t()-s)"
4.41987299919
```

BuddhaOhneHals:
It seems to be the usage of the `pkg_resources`-package.

I removed the import of the package from the `devices.py` file and replaced the body of `_default_pin_factory` with the following  two lines:
```
def _default_pin_factory(name=os.getenv('GPIOZERO_PIN_FACTORY', None)):
    from gpiozero.pins.rpigpio import RPiGPIOFactory
    return RPiGPIOFactory()
```
And now:
```
$ python -c "from timeit import default_timer as t;s=t();from gpiozero.output_devices import LED;print(t()-s)"
0.619349002838
```

BuddhaOhneHals:
I see two possible solutions:

[...]

PR #722, opened 2019-01-27
I said import, damn it!

### LINKED ISSUE #600: "import gpiozero" shouldn't fail without a valid pin factory
The following example is a valid method of setting the pin factory to mock:

```python
import gpiozero
from gpiozero.pins.mock import MockFactory
gpiozero.Device.pin_factory = MockFactory()
```

but it only works if `import gpiozero` is successful, having found a valid pin factory.

If you're on a Raspberry Pi, that's easy - you just install rpi.gpio (even if you don't intend to use it).

If you're on a PC, the following will happen:

- rpi.gpio will fail (even if it's installed - can only be run on a Pi)
- rpio will fail (even it it's installed - can only be run on a Pi)
- pigpio will fail (even if it's installed - pigpiod not running on localhost)
- native will fail (can't access /dev/mem)

And it runs out of options.

The only way to get it to work is to set the pin factory in an environment variable.

It would be nice if you could get past `import gpiozero`, and if no pin factory is found, it's set to `None` and fails when you try to create a device. Would that be that possible?

waveform80:
Permit gpiozero to be imported without trying to load a pin factory; only load a pin factory on first Device construction. While we're at it re-work the default pin factory loader to avoid pkg_resources import in as many cases as possible (while keeping it around for third-party pin implementations). This is just because pkg_resources takes *ages* to load (especially on an I/O limited platform like the Pi).

Fixes #600 and #675

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ develop-eggs
 
 # Installer logs
 pip-log.txt
+get-pip.py
 
 # Unit test / coverage reports
 coverage
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -11,12 +11,10 @@ import os
 import atexit
 import weakref
 import warnings
-from collections import namedtuple
+from collections import namedtuple, OrderedDict
 from itertools import chain
 from types import FunctionType
 from threading import Lock
-
-import pkg_resources
 
 from .pins import Pin
 from .threads import _threads_shutdown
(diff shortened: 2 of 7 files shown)


bennuttall:
I skimmed through the changes and saw you'd removed the PWM tests tuple and thought "I wonder if we can do that with a decorator" and saw that you had! That's really much nicer.

So I can now `import gpiozero` when no pin factory is set. Awesome. It seems to fall back nicely.

@waveform80 what do you think about failing tests on 3.2/3.3?

waveform80:
The tests are because Travis evidently has an ancient version of pytest for its 3.2 and 3.3 tarballs. I changed the pytest fixtures to use the newer generator format (where it uses yield to return the fixture value and everything after the yield is the fixture cleanup). Unfortunately, while this passed locally under tox with 3.2 and 3.3 it didn't on Travis so they apparently have a downlevel version of pytest that they install implicitly. I'll try and tweak the dependencies in setup.py to fix it, because I'd like to stick with the generator fixtures if possible (they're just cleaner and easier to read).

lurch:
Is there / should there / could there be a test for/using `ZombieThread` ?

PR #507, opened 2016-12-12
Make close() nethods more exception-safe

lurch:
If an exception is thrown during init, it's possible for a close() method (or
one of the methods called by close) to try accessing an attribute that hasn't
been created yet, which would throw an AttributeError exception. This commit
fixes that by ensuring that the close() methods (and methods called by them)
always use hasattr() or getattr() first, before trying to access the
attributes directly.

Fixes #503 and supersedes #494

PR #725, opened 2019-02-01
Coverage boost

waveform80:
Our coverage has been suffering a bit lately and needs a boost. This also fixes several bugs found along the way (I'd broken ButtonBoard two years ago and not noticed, and there was a *very* subtle bug in LocalFactory's hack of pin reservations), and tidies up the test suite to get rid of the ugly PWM-required lists (now there's just a pytest "pwm" fixture that can be added to any test requiring PWM).

Finally, this adds full support for later tox versions to permit parallel tests of various Python versions. It even supports this when testing physical pins by taking out a file-system level lock to prevent multiple interpreters hitting the pins at the same time. To try out the tox support, you can just run "tox" assuming you've got all the required Python interpreters installed (e.g. via the deadsnakes PPA on Ubuntu). If you haven't got all them installed, "tox -s" will skip missing interpreters. "tox -p auto -s" will run tests in parallel skipping missing interpreters. At the end you can run "coverage combine --rcfile coverage.cfg && coverage report --rcfile coverage.cfg" to receive a combined coverage report (unfortunately this seems to skip python 3.2 as the coverage file format changed from the ancient version in use with that interpreter ... oh well).

### DIFF
--- a/gpiozero/mixins.py
+++ b/gpiozero/mixins.py
@@ -68,14 +68,8 @@ class SourceMixin(object):
         super(SourceMixin, self).__init__(*args, **kwargs)
 
     def close(self):
-        try:
-            self.source = None
-        except AttributeError:
-            pass
-        try:
-            super(SourceMixin, self).close()
-        except AttributeError:
-            pass
+        self.source = None
+        super(SourceMixin, self).close()
 
     def _copy_values(self, source):
         for v in source:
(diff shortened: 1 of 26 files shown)


gpiozero/mixins.py
```diff
@@ -348,13 +342,10 @@ def __init__(self, *args, **kwargs):
         self._hold_thread = HoldThread(self)
 
     def close(self):
-        if getattr(self, '_hold_thread', None):
+        if self._hold_thread is not None:
```
lurch: Hmmm, so is #507 no longer an issue then? :man_shrugging:

bennuttall: What do you mean? #507 is a merged PR - do you mean the comment about `hasattr`?

lurch: I meant that #507 changed this from `if self._hold_thread:` to `if getattr(self, '_hold_thread', None):`, and this PR now changes it back to `if self._hold_thread is not None:`
(i.e. it might explode if `close()` gets called on an object which doesn't have a `_hold_thread` attribute yet)

Issue #642, opened 2018-04-09
Consider reimplementing blink/pulse by setting source


bennuttall:
As suggested in #449, perhaps `blink` and `pulse` should be implemented by setting `.source` rather than in a blink thread.

Also consider suggestion #397 - adding `is_blinking` property.

lurch:
Would we _really_ need a `.is_blinking` property, or would it be sufficient to just check `.source != None` ?
Otherwise would we also get requests for `.is_pulsing` or `.is_fading` properties?

bennuttall:
@martinohanlon what do you think?

martinohanlon:
I like the idea of using `source` internally to manage changes in value (i.e. blinking, fading), its self contained and can be implemented consistently across components.

I dont think an `.is_blinking` property is needed, it feels very specific with little reuse and as @lurch says where do you stop... If a user really needed to keep track of whether a component was blinking its an easy implementation in the calling program.

lurch:
> I dont think an `.is_blinking property` is needed

Should #397 be closed then?

martinohanlon:
Let no one say I am not prepared to re-evaluate my views... ;)

normanr:
Actually I was thinking of writing a helper (in the form of a decorator), that you could add to a generator function that would allow you to introspect the source function once the generator has started.

```python
def smart_generator(fn):

  class inner(object):
  
    def __init__(self, *args, **kwargs):
      self.func = wrapper
      self.args = args
      self.kwargs = kwargs
  
    def __iter__(self):
      return fn(*self.args, **self.kwargs)
  
    def __repr__(self):
      return '<smart_generator %s(%s) at 0x%08x>' % (
        fn.__name__,
        ', '.join(
          ([', '.join(repr(a) for a in self.args)] if self.args else []) +
          ([', '.join('%s=%r' % i for i in self.kwargs.items())] if self.kwargs else [])),
        id(self))

  @functools.wraps(fn)
  def wrapper(*args, **kwargs):
    return inner(*args, **kwargs)

  return wrapper

@smart_generator
def foo(a, b, d=8, e=9, c=7):
  for x in range(5):
    yield x

f = foo(1, '2', c='3', d=4)

print(f)  # <smart_generator foo(1, '2', c='3', d=4) at 0xf78d9a8c>
assert f.func == foo
```

Then this allows a few things:
[...]

PR #729, opened 2019-02-09
Tones

bennuttall:
Dave's `TonalBuzzer` reimplementation, which includes a separate `Tone` class.

Docs branch can be found at https://gpiozero.readthedocs.io/en/tones/api_tones.html


bennuttall:
TonalBuzzer's repr was broken so I've added a test and fixed it.

I think we should add a repr test for every class!

bennuttall:
This is cool...

```python
>>> from gpiozero import TonalBuzzer
>>> from gpiozero.tools import since_values
>>> bz = TonalBuzzer(20)
>>> bz.source = sine_values()
```

[video](https://photos.google.com/share/AF1QipPij0H3MV4fzllV_s0bTKNvYJNcSb1iLPGXP7arCAGNSRaw0VufptuwIBPUt8MGyA/photo/AF1QipNAne3Hf9GLAAuXEgtHn-CM07l5AvZv2XkHneN_?key=Wkg1aWlIbEhiZXExOWNrbUVycW9UQXlFTWcwTENR)

and again with 2 octaves:

[video](https://photos.google.com/share/AF1QipPij0H3MV4fzllV_s0bTKNvYJNcSb1iLPGXP7arCAGNSRaw0VufptuwIBPUt8MGyA/photo/AF1QipN8zyCLxEIvrZSt2BNrCUjwoBV_RLrWHsY4pbML?key=Wkg1aWlIbEhiZXExOWNrbUVycW9UQXlFTWcwTENR)

lurch:
@bennuttall Those video links just give me a 404 ?

gpiozero/output_devices.py
```diff
@@ -535,109 +536,123 @@ class TonalBuzzer(SourceMixin, CompositeDevice):
...
+            except ValueError:
+                raise ValueError(
+                    '%r is too low for %d octaves' %
+                    (self._mid_tone, self._octaves))
+            try:
+                self.max_tone.note
+            except ValueError:
+                raise ValueError(
+                    '%r is too high for %d octaves' %
+                    (self._max_tone, self._octaves))
```
lurch: Should this be `(self._mid_tone, self._octaves))` ?

bennuttall:
@lurch does this [shared album link](https://photos.google.com/share/AF1QipPij0H3MV4fzllV_s0bTKNvYJNcSb1iLPGXP7arCAGNSRaw0VufptuwIBPUt8MGyA/photo/AF1QipNAne3Hf9GLAAuXEgtHn-CM07l5AvZv2XkHneN_?key=Wkg1aWlIbEhiZXExOWNrbUVycW9UQXlFTWcwTENR) work?

bennuttall:
We should probably add a `beep` method which would be like [`Buzzer.beep`](https://gpiozero.readthedocs.io/en/stable/api_output.html#gpiozero.Buzzer.beep) (i.e. `blink` for buzzers), so it can "beep" a particular tone.

```python
>>> bz = TonalBuzzer(20)
>>> bz.beep()  # defaults: on_time=1, off_time=1, tone=self.mid_tone)
>>> bz.buzz(on_time=0.5, off_time=0.5, tone=Tone('C4'))
```

(also needs `n` and `background` args)

We could even add `pulse`, as we have seen `sin_values` works well with this type of buzzer.

We could leave this for now and add it in v1.6. I think it would be much easier if `blink` worked by setting `source` rather than maintaining a blink thread, as suggested in #642, again something to consider for v1.6. @waveform80 @lurch thoughts?

gpiozero/output_devices.py
```diff
@@ -535,109 +536,123 @@ class TonalBuzzer(SourceMixin, CompositeDevice):
...
+        also be set to play the specified tone.
+        """
+        if self.pwm_device.pin.frequency is None:
+            return None
+        else:
+            return Tone.from_frequency(self.pwm_device.pin.frequency)
+
+    @tone.setter
+    def tone(self, value):
+        self.play(value)
```
lurch: I'm not sure if it'd be more consistent with the rest of the GpioZero API to simply get rid of the `play()` method and just have a `tone` property-setter?

gpiozero/output_devices.py
```diff
@@ -535,109 +536,123 @@ class TonalBuzzer(SourceMixin, CompositeDevice):
...
+
     @property
     def value(self):
         """
         Represents the state of the buzzer as a value between 0 (representing
         the minimum note) and 1 (representing the maximum note). This can also
```
lurch: Seems to be some confusion in the docstrings as to whether value is 0 -> 1 or -1 -> 1 ? :confused:

bennuttall: It started as 0->1 but should be -1->1 now.

gpiozero/tones.py
```diff
@@ -0,0 +1,200 @@
...
+    variety of ways. For example as a straight frequency in `Hz`_ (which is the
+    internal storage format), as an integer MIDI note, or as a string
+    representation of a musical note.
+
+    All the following constructors are equivalent ways to construct the typical
+    tuning note, `concert A`_ at 440Hz which is MIDI note #69):
+
+        >>> from gpiozero.tones import Tone
+        >>> Tone(440.0)
+        >>> Tone(69)
```
lurch: Euw, I'm afraid I don't like this at all. Given that GpioZero is aimed at newcomers to Python, I think they'd find the fact that `Tone(60)` creates a midi-note of 60 and not a frequency of 60Hz very confusing.

gpiozero/tones.py
```diff
@@ -0,0 +1,200 @@
...
+    @classmethod
+    def from_note(cls, note):
+        """
+        Construct a :class:`Tone` from a musical note which must consist of
+        a capital letter A through G, followed by an optional semi-tone
+        modifier ("b" for flat, "#" for sharp, or their Unicode equivalents),
+        followed by an octave number (0 through 9).
+
+        For example `concert A`_, the typical tuning note at 440Hz, would be
+        represent as "A4". One semi-tone above this would be "A#4" or
```
lurch: represent -> represented ?

gpiozero/tones.py
```diff
@@ -0,0 +1,200 @@
...
+    @property
+    def note(self):
+        """
+        Return the (nearest) note to the tone's frequency. This will be a
+        string in the form accepted by :meth:`from_note`. If the frequency is
+        outside the range represented by this format ("A0" is approximately
+        27.5Hz, and "G9" is approximately 12.5Khz) a :exc:`ValueError`
+        exception will be raised.
+        """
+        offset = self.midi - 60  # self.midi - A4_midi + Tone.tones.index('A')
```
lurch: I wonder if it'd be better to do this calculation directly from `self.frequency`, instead of relying on `self.midi` ?

gpiozero/tones.py
```diff
@@ -0,0 +1,200 @@
...
+                str(octave)
+            )
+        raise ValueError('%f is outside the notation range' % self.frequency)
+
+    def up(self, n=1):
+        """
+        Return the :class:`Tone` *n* semi-tones above this frequency (*n*
+        defaults to 1).
+        """
+        return Tone.from_midi(self.midi + n)
```
lurch: I have absolutely zero musical background, but again I'm wondering why this is being done by converting to and from `.midi` instead of just using `.frequency` ?  (at the very least, presumably it introduces quantisation steps which the user might not expect from reading the docstring?)

lurch:
> @lurch does this shared album link work?

Nope, that also gives me a 404
(top tip - you can sometimes use a private / incognito browsing window to check that links that you're sharing are actually publicly visible)

SteveAmor:
> > @lurch does this shared album link work?
> 
> Nope, that also gives me a 404
> (top tip - you can sometimes use a private / incognito browsing window to check that links that you're sharing are actually publicly visible)

All the URLs work for me :-)

bennuttall:
I think you must need to be logged in to a Google account (sorry, just a quick and easy way to share videos without uploading anywhere)

SteveAmor:
> I think you must need to be logged in to a Google account (sorry, just a quick and easy way to share videos without uploading anywhere)

Ah, yes, that could be it.  Sent from my Chromebook :-)

gpiozero/output_devices.py
waveform80: Yup - whoops!

gpiozero/output_devices.py
waveform80: I partly left `play()` in because it was in Claire's original design and I didn't want to alter it too much, and also because a fair number of tests depended on it and I'm lazy :). Finally I think `b.play("A4")` is a little more obvious to new coders than `b.tone = Tone("A4")` which is a bit less imperative (not that that's a bad thing, but it's nice to have the choice for newbies).

That said, I agree entirely that I shouldn't be messing with the PWM device's frequency (directly) in two different places. Given `play()` is staying, its setter should probably stay as a simple alias of `play()`, but either `play()` should set `value`, or `value` setter should call `play()`. I'll have a look at what leads to the simpler code (might be the latter as there'll be less messing with logs).

gpiozero/output_devices.py
waveform80: Whoops - I'll fix that in a mo!

gpiozero/output_devices.py
lurch: Oh, could `b.tone = "A4"` not be made to work?
The only reason I brought this up is because e.g. https://gpiozero.readthedocs.io/en/master/api_output.html#angularservo only has an `angle` property and not a `move_to(angle)` method :wink: 
And it's obviously easier to add methods to an API later, than it is to add them now and remove them later. :man_shrugging:

gpiozero/tones.py
waveform80: Hmm, it's a tricky one. My reasoning is that this is a bit like colorzero's Color constructor(s). The main constructor is essentially a convenience and part of being convenient is a trade-off regarding ambiguities. Color's ambiguity is admittedly tiny (does `Color(1, 0, 0)` mean full red, or a very very dim red?), and Tone's is *much* larger but then there's a much larger overlap in the domains it's trying to represent (frequencies and MIDI notes here).

The alternate constructors mitigate this to a degree by allowing the user to explicitly specify what they expect (i.e. `Tone.from_frequency(60)` will always be 60Hz, and `Tone.from_midi(60)` will always be middle C). Still, that doesn't answer the question of why I choose to prioritize MIDI for the 0..127 range.

This was mostly down to how much of the range could be represented. Let's say we prioritize MIDI (as I did). In this case 0..127 covers the entire MIDI range, while the frequency domain only gets 128Hz..20Khz, losing the bottom range. However, the alternative (prioritize the frequency domain) means MIDI gets ... erm ... note 0, and frequency gets 1Hz..20Khz. An alternative I did consider was permitting 0..20 for MIDI notes because as a rough rule of thumb, tones below 20Hz are inaudible. However, this still leaves the MIDI system almost unusable because the concert range of A0..G9 is represented by MIDI notes 21..127. So to leave the MIDI system even vaguely useful in the constructor means chopping out a (small) portion of the lower end of the frequency range.

I did, however, leave one get-out clause: floating-point numbers are *always* treated as frequencies. So `Tone(60)` is MIDI note 60, `Tone(60.0)` is 60Hz. Frankly though, I'm not sure if that's an even more egregious ambiguity (which is why, partly out of embarrassment, it's not prominently documented!).

That leads finally to the question: do we care about MIDI? Is it useful? I'd argue for new programmers it's *much* easier to deal with than the alternative systems. Each note is a simple integer, not a string that needs constructing or parsing. There's no worrying about the octave number changing between B and C, rather than the wrapping point of A and G (as with the otherwise human friendly notation system<sup>1</sup>), and there's no powers or logs to deal with in calculating the frequency of the next or prior semi-tone (as there is with frequencies). It's just simple addition and subtraction: perfect for programming without getting into the heavy maths.

So that in the end is why I went with prioritizing MIDI over the frequency range - I'd be most interested in any counter-arguments though (I frequently enjoy @lurch's reviews - they often lead me to consider my justifications for things or, as in this case, to formalize some of the choices I'd already run through in my head).

<sup>1</sup> Actually, now I think about it, the notation system *isn't* particularly human friendly; it's quite weird in some respects. It's actually *musician* friendly (a fraternity I still somewhat count myself among the ranks of, having passed grade 8 theory *many* years ago ... don't ask me to explain the plagal cadence though; I've forgotten most of it!)

bennuttall: How about using kwargs like `Tone(frequency=440)` over `Tone.from_frequency(440)`?

gpiozero/tones.py
waveform80: It's possible but it'd just end up repeating the logarithm in the MIDI conversion. At the end of the day, both MIDI and the notation system are very similar: they both quantize the continuous frequency domain into a discrete "notes" domain. Looking at it another way: the MIDI system performs the quantization step, representing the result as a (limited) range of integers. The notation system then goes one step beyond this to a mod 12 representation (the octaves being the "major" step, and A..G, or rather C..B, being the "minor" steps).

That's without getting into the horror of different tuning systems though...

gpiozero/tones.py
waveform80: You're absolutely right that it involves quantization, but then that's the basis of MIDI (and indeed musical notes) generally. Quick bit of musical background (which I was astonished to realize I remember the other day!):

Take a note, say concert A (A4) which is 440Hz. If you want the same note an octave up you double the frequency, so A5 is 880Hz. Conversely the same note an octave down is half the frequency, so A3 is 220Hz. An octave is divided into 12 semi-tones (think of a semi-tone as the difference between one key and the next key on a piano keyboard - that's including the black keys, BTW). So, if 12 semi-tones (one octave) increases/decreases the frequency by a power of 2, what's the frequency ratio between one note and the next (or prior)? It's the twelfth root of 2<sup>*</sup>. So to go up a note, you multiple the frequency by the twelfth root of 2, to go down a note you divide by the same. This explains why we're using the MIDI system here: it keeps it simple. Going up a note is +1, going down a note is -1 and you don't start getting into twelfth roots and negative powers :)

<sup>*</sup> Or at least that's what we've mostly standardized on these days - the so called "[Equal temperament system](https://en.wikipedia.org/wiki/Equal_temperament)". I'd avoid looking at that link if you're not into musical theory - it's a *major* rabbit hole! Though it is quite interesting hearing music played in different tuning systems, given that some classical pieces were written prior to its invention and thus if played today on an equal temperament tuning, aren't necessarily played "how they were written".

gpiozero/tones.py
waveform80: > How about using kwargs like Tone(frequency=440) over Tone.from_frequency(440)?

Ah, knew I'd forgotten something. Yes, I should add those back in (although it's barely less typing than just using the explicit constructor). Incidentally, what do you think of the (horrid) `Tone(60)` vs `Tone(60.0)`. Worth excising? I'd say yes, in light of keyword-args.

lurch: ...and/or we could borrow inspiration from #470 and allow `Tone('MIDI25')` and `Tone('60Hz')`? :wink: 

As already mentioned, I have **no** musical background, but having `Tone(127)` construct a midi-note and having `Tone(128)` construct a frequency, sounds like a big disconnect / ambiguity to me. If the current behaviour is to be kept, IMHO it ought to be much better documented in the docstring.

gpiozero/output_devices.py
waveform80: Hmm, I like it - but to make it consistent really ought to accept anything that could be used to construct a `Tone` which leads back to the ambiguity question below re: ints in the range 0..127 (which Ben's suggestion of keyword args may mitigate with the explicit `Tone` constructor).

 I could raise a `Warning` in the constructor if an int in the ambiguous range (1..127) was used but it's worth bearing in mind that would trigger for pretty much *any* MIDI note, if not explicitly constructed.

gpiozero/tones.py
lurch: Fair enough, I guess I was just worried about the different notation systems (possibly) having different min/max values.

gpiozero/tones.py
lurch: Ahhh, I didn't realise it was so complex :laughing: :notes: 
My concern was that if a user had a frequency which was part-way between two midi notes, they might expect the `up` or `down` to shift their original frequency, and not jump to the midi frequency nearest to their original frequency first?
But it sounds like it's complicated enough already, so I won't ask to make it more complicated! :grin:

gpiozero/tones.py
lurch: ...I guess another alternative might be to _only_ allow construction from kwargs, and treat `Tone(60)` as an error? :man_shrugging:

gpiozero/tones.py
waveform80: Doh! Will be fixed in the next commit.

gpiozero/output_devices.py
lurch: Would it make sense to expose the various Tone constructors as property-setters in the TonalBuzzer? i.e. have `b.note = "A4"` be equivalent to `b.tone = Tone(note="A4")` (or `b.tone = Tone.from_note("A4")`), have `b.midi = 60` be equivalent to `b.tone = Tone(midi=60)` etc. Or does that become too wasteful of temporary objects?

LOL, maybe there needs to be a `TonesSetterMixin` :rofl:

gpiozero/tones.py
bennuttall: Yes, I think that's what Dave suggested earlier ^ if I'm reading it correctly. If so, I think we're all in agreement.

gpiozero/tones.py
waveform80: It's a reasonable concern, but if they're playing with frequencies which aren't already a note they're either playing with "pure" frequencies (i.e. not concerned with the musical aspect at all - which is fine, Tone derives from float after all so it can just be treated mathematically as such), in which case I'd *guess* they're not going to be interested in methods dealing with semi-tones, *or* they're playing in different tuning systems - which is sufficiently complex that I've not included any support for it.

... yet ;)

(actually I wouldn't know where to start with that stuff - it's something I *used* to know a tiny bit about 20+ years ago having studied it at a very high level without much detail - but I'd need to do a huge amount of reading before confidently designing an API to support it)

lurch: > if they're playing with frequencies which aren't already a note they're playing with "pure" frequencies, in which case I'd _guess_ they're not going to be interested in methods dealing with semi-tones

Fair point, I hadn't considered that. Told you I wasn't musical :laughing:

gpiozero/tones.py
waveform80: Actually I just suggested a warning in that case rather than a full-blown exception. My problem with an exception for ambiguous cases is that it'll blow up for almost *all* MIDI notes (1..127 as it's only 0 that can't be a frequency). A warning would have the same issue but would only report once (given warnings default filter), and wouldn't just terminate the user's script for wanting to use MIDI (or perhaps frequency!).

Still, lurch's concern is a reasonable one: what about users who simply want to play with frequencies (an interesting thing to do whether one is musical or not)?

Okay ... decision time:

1. I'll remove the really horrid int/float ambiguity handling. There's too much danger of someone doing arithmetic on an int, accidentally turning it into a float and suddenly getting a pure frequency instead of the MIDI note they expected and then having no clue why.

2. I'll introduce the warning for the ambiguous cases (numeric values between 1 and 127). The exception's too harsh (basically bans the use of MIDI notes entirely) and the warning, although it'll trigger for basically *all* MIDI notes, will only trigger on the first use due to the default suppression of repeat warnings. The warning message will include instructions for explicitly specifying the frequency and MIDI systems in case the user got the wrong one, and/or just wishes to suppress the warning.

3. I'll tweak the `tone` setter to permit any value that would validly construct a `Tone` instance. From the above, that'll also raise a warning in ambiguous cases.

4. I won't be adding in the extra `note`, `midi`, and `frequency` properties - partly because it makes for a rather bloated API, but also partly because I have some plans/vague musings for "note" (hint: what's the difference between a tone and a note?).

5. I'm still undecided on the removal of `play()`. So much for decision time ... Ben, you want to make a call on this?

bennuttall: `play` should stay in play :)

gpiozero/output_devices.py
```diff
@@ -535,109 +536,123 @@ class TonalBuzzer(SourceMixin, CompositeDevice):
...
 
     def __repr__(self):
         try:
-            return '<gpiozero.%s object on pin %r, is_active=%s>' % (
-                self.__class__.__name__, self.pwm_device.pin, self.is_active)
+            if self.value is None:
+                return '<gpiozero.TonalBuzzer on pin %r, silent>' % (
+                    self.pwm_device.pin,)
+            else:
+                return '<gpiozero.TonalBuzzer on pin %r, playing %s>' % (
```
lurch: Looks like you need to add ` object` into these `repr` strings to get the Travis tests passing @bennuttall

bennuttall: :+1:

gpiozero/internal_devices.py
```diff
@@ -483,8 +483,12 @@ def usage(self):
         """
         # XXX Use os.statvfs?
         df = subprocess.Popen(['df', '--output=pcent', self.filesystem], stdout=subprocess.PIPE)
-        output = df.communicate()[0].split(b'\n')[1].split()[0]
-        return float(output.decode('UTF-8').rstrip('%'))
+        try:
```
lurch: There's no `with` style wrapper for this?

gpiozero/tones.py
```diff
@@ -36,8 +39,17 @@ class to easily represent musical tones. The class can be constructed in a
...
-    :meth:`from_frequency`, :meth:`from_midi`, or :meth:`from_note`.
+    end of the frequencies, from 128Hz down), you can use one of the explicit
+    constructors, :meth:`from_frequency`, :meth:`from_midi`, or
+    :meth:`from_note`, or you can specify a keyword argument when
+    constructing::
+
+        >>> Tone.from_frequency(440.0)
+        >>> Tone.from_midi(69)
+        >>> Tone.from_note('A4')
+        >>> Tone(frequency=440.0)
```
lurch: IIRC you said you were getting rid of the int-vs-float ambiguity, so I guess the `440.0`s here could just be `440`?

gpiozero/tones.py
```diff
@@ -61,16 +73,37 @@ class to easily represent musical tones. The class can be constructed in a
...
+        if value is None:
+            if len(kwargs) != 1:
+                raise TypeError('expected precisely one keyword argument')
+            key, value = kwargs.popitem()
+            try:
+                return {
+                    'frequency': cls.from_frequency,
+                    'midi': cls.from_midi,
+                    'note': cls.from_note,
+                }[key](value)
```
lurch: I'm beginning to wonder if you deliberately write obfuscated code :rofl:

gpiozero/tones.py
```diff
@@ -61,16 +73,37 @@ class to easily represent musical tones. The class can be constructed in a
...
+            except KeyError:
+                raise TypeError('unexpected keyword argument %r' % key)
+        else:
+            if kwargs:
+                raise TypeError('cannot specify keywords with a value')
+            if isinstance(value, (int, float)):
+                if 0 <= value < 128:
+                    warnings.warn(
```
lurch: Don't we only need the warning `if value > 0` ?

gpiozero/output_devices.py
waveform80: Yup, done :)

gpiozero/internal_devices.py
waveform80: There is ... but not for 2.7, annoyingly. However, I've just pushed a better fix for this in another branch (uses statvfs instead of a sub-process). I'll merge that a little later to fix this "properly"

gpiozero/tones.py
waveform80: Yup, certainly could be - I'll change it.

gpiozero/tones.py
waveform80: Good point! Yup, will change that too.

gpiozero/tones.py
waveform80: Just calling the result of a looking up a method in a dict; that's not obscure ... by my standards *cough*

PR #732, opened 2019-02-10
Add repr tests, close #730

### LINKED ISSUE #730: Ensure every class's repr is tested
It's possible for all tests to be passing but a device's repr fail, as I just found with `TonalBuzzer`.

I added a call to `repr(tb)` at the top of the first test for `TonalBuzzer` and since calling it raised an exception.

I know some tests actually inspect the contents of repr. Do we think this is necessary for all classes? Maybe. I think there should be at least a "test" that repr doesn't fall over.

@waveform80 @lurch thoughts?

bennuttall:
Add repr tests and make internal device repr uniform (and use super when it fails)

Fixes #730

### DIFF
--- a/coverage.cfg
+++ b/coverage.cfg
@@ -8,7 +8,6 @@ ignore_errors = True
 show_missing = True
 exclude_lines =
     pragma: no cover
-    def __repr__
     if self\.debug
     raise AssertionError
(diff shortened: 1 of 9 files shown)


lurch:
Very nice! :smiley:

PR #735, opened 2019-02-11
Bump coverage a tiny amount, and fix a genuine bug

lurch:
(the bug is that `spi_devices.py` wasn't importing `InputDeviceError`)

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -114,7 +114,7 @@ class InputDevice(GPIODevice):
         else:
             if active_state is not None:
                 raise PinInvalidState(
-                    'Pin %d is not floating, but "active_state" it not None' %
+                    'Pin %d is not floating, but "active_state" is not None' %
                     self.pin.number)
             self._active_state = False if pull_up else True
         self._inactive_state = not self._active_state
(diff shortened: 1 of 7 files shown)


lurch:
Hmmm, Travis is telling me that the tests all pass on Py2.x and Py3.x but fail on pypy and pypy3 with:
```
GPIOPinInUse: pin 4 is already in use by <gpiozero.InputDevice object on pin GPIO4, pull_up=None, is_active=False>
```

Any ideas @waveform80 ?

tuftii:
Is this the same GPIO pin in use thing we saw in #680?

lurch:
@tuftii It's similar, but I don't think it's the same thing. IIRC the bug you found affected all versions of python, where the bug here is only affecting pypy.

waveform80:
@lurch yup - the issue is the pypy GC is too slow at destroying the first InputDevice on GPIO4. It's fine in other tests where "with" is used as the context manager will explicitly close the device (no waiting for the GC). But the first test in `test_input_invalid_external_pull_configuration` is constructing a bare InputDevice outside a "with" section (which is entirely correct BTW, but in this case it'll cause this issue with pypy). I've just pushed a simple fix which is to separate it into two tests; the mock_factory cleanup will explicitly close all attached devices which should deal with the issue.

Issue #589, opened 2017-08-21
Booleanized .values?


bennuttall:
I realised while answering #587 that while some devices have linear value scales, they also present a boolean state, for example the light sensor `.value` might be `0.5` but its `.light_detected` is `True`.

In the procedural code solution to the problem, the light sensor value was unused, and just the boolean state was used:

```python
while True:
    if pir.motion_detected and not ldr.light_detected:
        led.on()
    else:
        led.off()
```

However, when using source/values, the `.values` property yields float values, not booleans, so I had to manually booleanise the `.value` using the same logic that's found in `light_detected` (`is_active`):

```python
led.source = all_values(pir.values, booleanized(ldr.values, 0, ldr.threshold))
```

So I was thinking, it would be much neater if we provided a `.values`-like iterator for the boolean state, for example:

```python
@property
def bool_values(self):
    while True:
        yield self.is_active
```

Light and dark on their own would become:

```python
ldr.bool_values  # light
negated(ldr.bool_values)  # dark
```

Making my solution neater:

```python
led.source = all_values(pir.values, negated(ldr.bool_values))
```

Like many problems, this is easy to solve, but the hard bit is *naming it*. Any ideas? @waveform80 @lurch @martinohanlon

lurch:
```python
ldr_bool_values = booleanized(ldr.values, 0, ldr.threshold)
led.source = all_values(pir.values, negated(ldr_bool_values))
```
is quite close to your last example? :wink: 

But yeah, for those devices which have an 'implicit' boolean state like `light_detected` it _might_ be nice to offer a simpler way to use that with source-values.
Instead of naming it `bool_values` what about `active_values`? So that `values` returns an infinite generator of `value`, and `active_values` returns an infinite generator of `is_active`?  Perhaps the LDR class could even then alias `lights` to `active_values`? But again, this becomes a tricky naming problem... :confused: 
```python
led.source = all_values(pir.values, negated(ldr.lights))
```

Of course the advantage of explicitly using `booleanized` is that it makes it easy to use the funky hysteresis option I added, to prevent 'flickering' when the LDR is hovering around the threshold value.

bennuttall:
Ok, just a thought.

If you wanted an LED to come on when the CPU temperature exceeds 50 degrees, the following would currently work:

```python
led = LED(2)
cpu = CPUTemperature(min_temp=50, max_temp=100)

led.source = booleanized(cpu, 0, 1)
```

[...]

PR #737, opened 2019-02-11
Expand source/values docs and use graphviz for diagrams, close #733

### LINKED ISSUE #733: Use graphviz for source/values diagrams
https://gpiozero.readthedocs.io/en/latest/source_values.html

bennuttall:
I thought I'd unstaged the changes to the hierarchy images but they seem to have made their way here.

Note: I've changed the docs makefile as I put the source/values images in a separate directory.

I've extended the contents of the source/values docs page you can see here: https://gpiozero.readthedocs.io/en/graphvis-source-values-docs/source_values.html

(seeing it on RTD for the first time makes me wish there was more padding beneath the images. Is that possible?

So you can ignore most files from the review, just take a look at the docs page, and @waveform80 are you happy with the .dot files?


docs/examples/garden_light.py
```diff
@@ -0,0 +1,11 @@
+from gpiozero import LED, MotionSensor, LightSensor
+from gpiozero.tools import booleanized, all_values
+from signal import pause
+
+garden = LED(2)
+motion = MotionSensor(4)
+light = LightSensor(5)
+
+garden.source = all_values(booleanized(light, 0, 0.1), motion)
```
lurch: Wouldn't it be better to use `negated(light)` rather than `booleanized(light, 0, 0.1)` ?

docs/source_values.rst
```diff
@@ -178,12 +181,106 @@ In this example, the LED is lit only when the button is not pressed:
...
+a range between -1 and 1 (e.g. :class:`Motor`, :class:`Servo` and
+:class:`TonalBuzzer`). Some source tools output values between -1 and 1, which
+are ideal for these devices, for example passing :func:`~tools.sin_values` in:
+
+.. image:: images/source_values/sin_values.*
+    :align: center
+
+.. literalinclude:: examples/sin_values.py
+
+In this example, all three devices are following the `sine wave`_. The motor
```
lurch: I wonder if it might be better to use cosine here instead of sine, as it might be easier to explain (in words) going from the highest value, down to the mid value, and then down to the lowest value; than it is to explain starting at the mid value, going up to the max value, then back to the mid value, then down to the lowest value ? :man_shrugging:

docs/source_values.rst
```diff
@@ -178,12 +181,106 @@ In this example, the LED is lit only when the button is not pressed:
...
+
+.. image:: images/source_values/disk_usage_bar_graph.*
+    :align: center
+
+.. literalinclude:: examples/disk_usage_bar_graph.py
+
+Demonstrating a garden light system whereby the light comes on if it's dark and
+there's motion is simple enough, but it requires using the
+:func:`~tools.booleanized` source tool to convert the light sensor from a float
+value into a boolean:
```
lurch: Ahhh, this explains my earlier comment :wink: I'd forgotten that the light sensor returns a continuous 0->1 value and not a boolean 0 or 1 value.

Although it raises an interesting point for future discussion - for those devices like LightSensor that have an `is_active` property, I wonder if there'd be any easy way to use `is_active` (boolean) with source-tools instead of `value` (float) ?

bennuttall: I suggested that in #589 and you disagreed! :P

docs/source_values.rst
```diff
@@ -38,12 +38,17 @@ GPIO Zero provides a method of using the declarative programming paradigm to
...
+.. image:: images/source_values/led_button.*
+    :align: center
+
 .. literalinclude:: examples/led_button.py
 
 which is equivalent to:
 
 .. literalinclude:: examples/led_button_loop.py
 
+except that the former is updated in a background thread.
```
lurch: Might be worth expanding to e.g. "...the former is updated in a background thread (which enables you to do other things at the same time)". ?

docs/source_values.rst
```diff
@@ -38,12 +38,17 @@ GPIO Zero provides a method of using the declarative programming paradigm to
...
+
 Every device has a :attr:`~Device.value` property (the device's current value).
 Input devices can only have their values read, but output devices can also have
```
lurch: This might be clearer as "Input devices (like buttons) can only have their values read, but output devices (like LEDs) can also ..." ?

docs/source_values.rst
lurch: LOL, I'd completely forgotten about that! :joy: 
Looks like I _didn't_ disagree though...
> But yeah, for those devices which have an 'implicit' boolean state like `light_detected` it _might_ be nice to offer a simpler way to use that with source-values.
Instead of naming it `bool_values` what about `active_values`? So that `values` returns an infinite generator of `value`, and `active_values` returns an infinite generator of `is_active`?

Why was that issue closed?

bennuttall: I think I agreed with you that using `booleanized` was easy enough and avoiding API bloat. Also trying to keep the number of unresolved issues down, closing any I think are dead suggestions.

docs/source_values.rst
```diff
@@ -38,12 +38,17 @@ GPIO Zero provides a method of using the declarative programming paradigm to
...
 their value set to alter the state of the device:
```
lurch: Perhaps the code-block just below here should use a LED rather than a PWMLED? Start simple and work upwards :wink:

bennuttall: I start with LED at the top of the page

docs/source_values.rst
```diff
@@ -65,13 +70,16 @@ device will iterate over the values of the device provided, setting the
 device's value to each element at a rate specified in the
```
lurch: I wonder if the paragraph above should have links (perhaps to the python docs) explaining the terms "generator" and "iterator" ?

docs/source_values.rst
bennuttall: Done - in next push

docs/source_values.rst
bennuttall: Done - in next push

docs/source_values.rst
bennuttall: Done - in next push

docs/source_values.rst
```diff
@@ -65,13 +70,16 @@ device will iterate over the values of the device provided, setting the
...
-the values of an input device, like the example above. A more interesting
+match the values of an input device, like the example above. A more interesting
 example would be a potentiometer controlling the brightness of an LED:
 
+.. image:: images/source_values/pwmled_pot.*
+    :align: center
+
 .. literalinclude:: examples/pwmled_pot.py
 
 The way this works is that the device's :attr:`~ValuesMixin.values` property is
```
lurch: "The way this works is that the device's `values` property is used to feed values into the device."
sounds a little bit confusing (two mentions of "device"). Perhaps
"The way this works is that the pot device's `values` property is used to feed values into the led device's `source` property." ?

docs/source_values.rst
```diff
@@ -89,25 +101,38 @@ used to feed values into the device. Prior to v1.5, the
 It is also possible to set an output device's :attr:`~SourceMixin.source` to
 another output device, to keep them matching:
 
+.. image:: images/source_values/matching_leds.*
+    :align: center
+
```
lurch: Perhaps add a sentence explaining what the example does? i.e. pressing the button will simultaneously light both the red and green LEDs.

docs/source_values.rst
```diff
@@ -178,12 +185,106 @@ In this example, the LED is lit only when the button is not pressed:
...
+While most devices have a :attr:`~Device.value` range between 0 and 1, some have
+a range between -1 and 1 (e.g. :class:`Motor`, :class:`Servo` and
+:class:`TonalBuzzer`). Some source tools output values between -1 and 1, which
+are ideal for these devices, for example passing :func:`~tools.sin_values` in:
+
+.. image:: images/source_values/sin_values.*
+    :align: center
+
+.. literalinclude:: examples/sin_values.py
```
lurch: Alternatively, I guess this could do:
```python
motor.source = sin_values()
servo.source = motor
buzzer.source = servo
```
which means the sin_values only get calculated once, not three times :slightly_smiling_face:

docs/source_values.rst
```diff
@@ -178,12 +185,106 @@ In this example, the LED is lit only when the button is not pressed:
...
+
+In this example, all three devices are following the `sine wave`_. The motor
+ramps up from 0 (stopped) to 1 (full speed forwards), then back down to 0 and on
+to -1 (full speed backwards) in a cycle. Similarly, the servo moves from its mid
+point to the right, then towards the left; and the buzzer starts with its mid
+tone, gradually raises its frequency, to its highest tone, then down towards its
+lowest tone. Note that setting :attr:`~SourceMixin.source_delay` will alter
+the speed at which the device iterates through the values.
```
lurch: Might be worth including an example here, so that people can see how / where the `source_delay` is set? I guess a nice example might be setting `source_delay` to three different values (and ignoring my previous comment about setting `servo.source = motor` :wink: ) so that the three different devices ramp at three different speeds?

docs/source_values.rst
```diff
@@ -178,12 +185,106 @@ In this example, the LED is lit only when the button is not pressed:
...
+Internal devices
+----------------
+
+GPIO Zero also provides several :doc:`internal devices <api_internal>` which
+represent facilities provided by the operating system itself. These can be used
+to react to things like the time of day, or whether a server is available on the
+network. These classes include a :attr:`~ValuesMixin.values` property which can
+be used to feed values into a device's :attr:`~SourceMixin.source`. For example,
+a lamp connected to an :class:`Energenie` socket can be controlled by a
+:class:`TimeOfDay` object so that it is on during the hours of 8am and 8pm:
```
lurch: during -> between?

lurch:
Just a general comment wrt the graphviz diagrams - for the diagrams that illustrate what the code examples are doing, there seems to be a bit of inconsistency between whether the object-labels in the graphviz diagrams are the name of the class (e.g. `LED`), the name of the variable (e.g. `Lamp`) or something else (e.g. `Button A` when the variable is named `button_a`, or `Green LED` when the variable is named `green`).
IMHO consistency is always good, and personally I think it might work well to always use the variable name (i.e. `led`, `lamp`, `button_a`, `green`, etc.).

docs/source_values.rst
```diff
@@ -178,12 +185,106 @@ In this example, the LED is lit only when the button is not pressed:
 
 Some tools combine the values of multiple sources:
 
-.. image:: images/combining_sources.*
+.. image:: images/source_values/combining_sources.*
     :align: center
 
 In this example, the LED is lit only if both buttons are pressed (like an
 `AND`_ gate):
```
lurch: Might be worth repeating this example again but using `any_values`, just to illustrate the different kinds of thing that can be done?

docs/source_values.rst
lurch: Hmmm, looking at https://gpiozero.readthedocs.io/en/master/api_input.html it seems both MotionSensor and LightSensor inherit from SmoothedInputDevice - does that mean that they both have a float-value, and so **both** need to be booleanized?

docs/source_values.rst
```diff
@@ -178,12 +185,106 @@ In this example, the LED is lit only when the button is not pressed:
...
+    >>> tuple(robot.value)
+    (-1.0, -1.0)
+    >>> robot.value = (1, 1)  # robot is now driven forwards
+
+Use two potentiometers to control the left and right motor speed of a robot:
+
+.. image:: images/source_values/robot_pots_1.*
+    :align: center
+
+.. literalinclude:: examples/robot_pots_1.py
```
lurch: Given that `left` and `right` are used in the robot's constructor, this might be clearer if you used `left_pot` and `right_pot` variables instead of `left` and `right`? (and also in the next example)

docs/source_values.rst
```diff
@@ -178,12 +185,106 @@ In this example, the LED is lit only when the button is not pressed:
...
+
+To include reverse direction, scale the potentiometer values from 0->1 to -1->1:
+
+.. image:: images/source_values/robot_pots_2.*
+    :align: center
```
lurch: It's not clear why the first example uses `zip_values` but the second example uses `zip`? A bit of explanation would be great.

docs/source_values.rst
bennuttall: I've changed "device" to "input device" and "output device". Much better.

docs/source_values.rst
bennuttall: Done

docs/source_values.rst
bennuttall: Yes, that's true, but I thought it would make the diagram more confusing when the point I'm trying to make is that each device is being set by `sin_values()`. I'll add a note to point that out though.

bennuttall: Actually, it might be better if I do:

```python
sin = sin_values()
buzzer.source = sin
servo.source = sin
motor.source = sin
```

That means they stay in sync and the diagram can still be simple.

bennuttall: No - maybe that won't work...

bennuttall: I'll settle for:

```python
motor.source = sin_values()
servo.source = motor
buzzer.source = motor
```

docs/source_values.rst
bennuttall: The candle example sets source_delay, that'll do for now.

docs/source_values.rst
bennuttall: I'll add a note.

docs/source_values.rst
lurch: @waveform80 I've always wondered... given the alternatives of
```python
motor.source = sin_values()
servo.source = motor
buzzer.source = servo
```
or
```python
motor.source = sin_values()
servo.source = motor
buzzer.source = motor
```

is one version more preferable or more efficient than the other?

docs/source_values.rst
lurch: Ahh, fair enough. But if that's the only place where source_delay is demonstrated, it might be worth pointing it out? (e.g. "Note that this example also uses `source_delay` to slow down the speed of the flickering")

Should https://gpiozero.readthedocs.io/en/master/source_values.html#source-tools and https://gpiozero.readthedocs.io/en/master/api_tools.html#gpiozero.tools.random_values be using the same code snippet?

docs/source_values.rst
bennuttall: Ok

docs/source_values.rst
bennuttall: I'll add a note

docs/source_values.rst
bennuttall: I've added a note

docs/source_values.rst
bennuttall: Added a note

bennuttall:
> Just a general comment wrt the graphviz diagrams - for the diagrams that illustrate what the code examples are doing, there seems to be a bit of inconsistency between whether the object-labels in the graphviz diagrams

I've updated them (about to push) to be more consistent.

bennuttall:
https://gpiozero.readthedocs.io/en/latest/source_values.html

docs/source_values.rst
bennuttall: @lurch I've figured out why they're different. `MotionSensor` has a default `queue_len` of 1 and `LightSensor` has a default `queue_len` of 5, therefore LS's value is averaged and MS's effectively isn't.

However, setting MS's `queue_len` to >1 gives it a boolean value:

```python
>>> pir = MotionSensor(14)
>>> pir.value
0
>>> pir = MotionSensor(15, queue_len=2)
>>> pir.value
0.0
```

lurch: Yup, see my comment here https://github.com/RPi-Distro/python-gpiozero/pull/744#discussion_r256206318

Issue #323, opened 2016-05-10
Energenie socket activation behaviour


tombennet:
This issue concerns the [Energenie class](https://gpiozero.readthedocs.io/en/v1.2.0/api_boards.html#energenie) added in v1.2.0. 

Let's say we're using an instance of this class to activate or deactivate a socket based on a condition, via a script which runs frequently on a schedule, using `socket.on()` / `socket.off()`. If the condition for activation is met and the socket in question is already on, the socket switches off then on in quick succession, with two clicks.

This is in contrast to the [standalone Energenie module](https://github.com/RPi-Distro/python-energenie) - when an already-activated socket is switched on with `energenie.switch_on(1)`, it simply stays on. In cases where the script needs to run frequently, the unnecessary off/on cycles caused by this behaviour may reduce the longevity of the device.

I've tested this issue thoroughly with more than one physical Energenie socket. The problem does not happen in reverse (i.e. deactivating a socket which is off does not cause an on/off cycle, it just stays off). In case it's useful for anyone, the script which first alerted me to this behaviour is [available here](https://github.com/tombennet/raspberry-pi-terrarium/blob/dd509b2791ec6df0e62f0cb6d83b959ad68d5681/monitor.py).

I suspect this issue stems from the fact that these sockets provide no means of reading their state, as mentioned in the documentation. Either way, the ideal outcome would be for activation of a socket which is already on to leave the state of the device unchanged. 

Any input / suggestions much appreciated. Thanks.

waveform80:
The reason for this is related to the fact that the sockets provide no means of reading their state. In order to ensure that the state of the socket is "known" at construction time there's an `initial_state` value in the constructor which defaults to `False` (i.e. off) which is applied at construction time.

So, if you're running your script every time to set a state (as compared to a long-running script which uses the same object) you probably want to just set the initial_state (rather than calling on/off), e.g.

```python
from gpiozero import Energenie
lamp = Energenie(1, initial_value=True)
```

waveform80:
Incidentally, there is a later version of these sockets (rather more expensive) which do permit reading their state. We haven't got support for those (yet) though.

tombennet:
Thanks Dave, that makes perfect sense now. 

[...]

PR #739, opened 2019-02-11
We can't read the value of an Energenie socket, so we shouldn't allow…

lurch:
… a None initial_value

See https://github.com/RPi-Distro/python-gpiozero/issues/323#issuecomment-218543961


lurch:
Hmmm, for both this PR and #734 the Travis builds are failing on Python2.7 because `No output has been received in the last 10m0s` :confounded: (even though all the Python 3.x builds take less than 2 minutes each!)

PR #743, opened 2019-02-11
Use statvfs instead of a subprocess

waveform80:
Simplify DiskUsage to avoid relying upon a sub-process


gpiozero/internal_devices.py
```diff
@@ -461,8 +461,7 @@ class DiskUsage(InternalDevice):
     """
     def __init__(self, filesystem='/', threshold=90.0, pin_factory=None):
         super(DiskUsage, self).__init__(pin_factory=pin_factory)
-        if not os.path.ismount(filesystem):
-            raise ValueError('invalid filesystem')
+        os.statvfs(filesystem)
```
lurch: Might be worth adding a comment in here explaining that this is just checking that the `filesystem` exists, which is why the return value from statvfs is being thrown away? Looks mighty weird otherwise ;)

gpiozero/internal_devices.py
```diff
@@ -481,22 +480,23 @@ def usage(self):
...
-        return self.usage / 100.0
+        # This slightly convoluted calculation is equivalent to df's "Use%";
+        # it calculates the percentage of FS usage as a proportion of the
+        # space available to *non-root users*. Technically this means it can
+        # exceed 100% (when FS is filled to the point that only root can write
+        # to it), hence the clamp.
+        vfs = os.statvfs(self.filesystem)
+        used = vfs.f_blocks - vfs.f_bfree
+        total = used + vfs.f_bavail
+        return min(1.0, used / total)
```
lurch: Could `total` ever be 0 ?

gpiozero/internal_devices.py
waveform80: Good point - and yes, that's exactly what it's doing :)

gpiozero/internal_devices.py
waveform80: Good question - I took a brief wade thru df and coreutils' code when writing this (to make sure my method of matching df's Use% output was equivalent) but I didn't notice any particular handling for div-by-0 (mind you I wasn't specifically looking so I may've missed it). I did note (from comments in that code) that under weird circumstances they can be negative but frankly I filed that under "weird enough that I don't care until someone complains"!

lurch: I found this the other day, which also seems to corroborate things https://stackoverflow.com/questions/4965355/converting-statvfs-to-percentage-free-correctly

PR #744, opened 2019-02-12
Fix #379 (again ;)

### LINKED ISSUE #379: value property documentation
I just noticed that _many_ of the classes in the [documentation](http://gpiozero.readthedocs.io/en/stable/index.html) don't document their `value` property. Given that this is fairly essential knowledge in order to be able to utilise the [sourcetools](http://gpiozero.readthedocs.io/en/stable/api_tools.html) functions properly, IMHO it ought to be added.

waveform80:
Fixes #379


gpiozero/input_devices.py
```diff
@@ -515,6 +525,15 @@ def __init__(
             self.close()
             raise
 
+    @property
+    def value(self):
+        """
+        Returns a value representing the average of the queued values. This
+        is nearer 0 for black under the sensor, and nearer 1 for white under
+        the sensor.
```
lurch: Apologies for being a bit blunt (I'm tired), but "nearer 0" and "nearer 1" sounds a bit vague and wishy-washy?
I'm probably missing something obvious, but surely the sensor is either detecting a line, or not detecting a line? It's not probabilistic, is it? :confused: 

Or if it _is_ probabilistic, maybe this would be clearer if phrased more explicitly like "A value less than 0.5 means there's probably black under the sensor, and a value above 0.5 means there's probably white under the sensor"?
IMHO "a value less than 0.5" seems easier to understand than "a value nearer 0" :man_shrugging: 

BTW does black being ~0 and white being ~1 mean you changed your mind again about the "logic" behind `.value`? (that you were discussing in another issue)

I googled and found https://www.instructables.com/id/Using-IR-Sensor-TCRT-5000-With-Arduino-and-Program/ which suggests turning on and off the emitter, to help remove ambient noise from the receiver? I guess that would be hardware-incompatible with this current LineSensor class, but maybe it'd be worth adding a DenoisedLineSensor in future which did that? :smiley:

gpiozero/input_devices.py
```diff
@@ -598,6 +617,16 @@ def __init__(
             self.close()
             raise
 
+    @property
+    def value(self):
+        """
+        With the default *queue_len* of 1, this is effectively boolean where 0
+        means no motion detected and 1 means motion detected. If you specify
+        a *queue_len* greater than 1, this will be an averaged value where
+        values closer to 1 imply motion detection.
```
lurch: Ahhhh, so this explains my confusion at https://github.com/RPi-Distro/python-gpiozero/pull/737#discussion_r255509849 :grinning: 

If it's only "effectively boolean" and not "actually boolean" does that mean the GardenLight example in https://gpiozero.readthedocs.io/en/master/source_values.html#internal-devices **should** be booleanizing the MotionSensor too?

gpiozero/input_devices.py
waveform80: Yeah, it is wishy-washy and the logic is indeed (still) backwards. Haven't changed my mind: it's wrong - but it's a non-backwards-compatible change (if anyone's relying on the current "polarity" of value) so we need to be a bit careful changing it. The nearer 0/1 thing is because it's a "smoothed" input device. The sensor does indeed just return 0/1 but we queue up those values as fast as we can and return an average of them. Historically this was a mean (hence the nearer) - now I think it's a actually a median so with a pure digital input it *should* also be a digital output (assuming an odd-length queue).

Anyway, for the next point release, this is something to come back to and clean up properly :)

gpiozero/input_devices.py
```diff
@@ -889,6 +925,16 @@ def distance(self):
         """
         return self.value * self._max_distance
 
+    @property
+    def value(self):
+        """
+        Returns a value between 0, indicating the reflector is either touching
+        the sensor or is sufficiently near that the sensor can't tell the
+        difference, and 1, indicating the reflector is at or beyond the
```
lurch: This is the only place that "the reflector" is mentioned, and (IMHO) it makes "the reflector" sound like a specific thing (e.g. a bike reflector) rather than "the distance measured by the sensor" that the rest of the docstrings talk about.

gpiozero/output_devices.py
```diff
@@ -134,9 +134,8 @@ def toggle(self):
     @property
     def value(self):
         """
-        Returns :data:`True` if the device is currently active and
-        :data:`False` otherwise. Setting this property changes the state of the
-        device.
+        Returns 1 if the device is currently active and 0 otherwise. Setting
+        this property changes the state of the device.
```
lurch: Does this change mean that any examples that read or write a boolean `.value` ought to be updated to read or write 0 or 1 instead?

bennuttall: Yes I think they ought to be :+1:

PR #745, opened 2019-02-12
Update Diagrams

waveform80:
Use Arial instead of Sans because unlike Linux and Windows, Mac OS X thinks a good substitute for "Sans" is ... erm ... a tiny serifed font. Also make all the circuit diagrams in fritzing (for consistency) and add some of the missing diagrams (robot, fix motion sensor component, etc. etc.)


gpiozero/tones.py
```diff
@@ -53,7 +53,15 @@ class to easily represent musical tones. The class can be constructed in a
 
     Several attributes are provided to permit conversion to any of the
     supported construction formats: :attr:`frequency`, :attr:`midi`, and
-    :attr:`note`.
+    :attr:`note`. Methods are provided to step :meth:`up` or :meth:`down` to
+    adjacent MIDI notes.
```
lurch: Oh, interesting. Probably worth adding this much-clearer explanation to the 'up` and `down` doscstrings too, as IMHO (as a non-musician) talking about moving up or down adjacent MIDI notes is much easier to understand than moving a specific number of semi-tones above/below a frequency...?

lurch:
I just picked a random example and looked at `button_robot_bb.png`. I suspect it'll be much easier for students to follow the wiring diagrams if wires which are overlapping or very close to each other (e.g. the blue and green wires in this diagram) have different colours?
But I have no educational background, so maybe @bennuttall 's experience from Picademey etc. will be useful here.

PR #750, opened 2019-02-14
Expand JamHat buzzer range

tuftii:
Just a little change to allow the buzzer to play from A0 to A8, rather than the narrow A3 to A5 default.


bennuttall:
Was talking to @waveform80 about this - we really ought to provide a way to set this at init - might be part of a bigger change to allow a dict of params for each device within the composite device.

We ought to leave JamHat's buzzer as TonalBuzzer's default, but provide a way to set it. I'm going to close this, and I'll open an issue to discuss the wider solution.

Thanks though

tuftii:
Ah, thanks @bennuttall.

PR #752, opened 2019-02-16
Fix buzzer-class in JamHat's docstring

lurch:


lurch:
LOL, I've got no idea how or why this _decreases_ coverage :laughing:

Issue #144, opened 2016-01-10
Add Steppter Motor


bennuttall:
Here's someone using `OutputDevice`: https://github.com/topshed/PiClubSushi/blob/master/17-Stepper.pdf

bennuttall:
More info here: http://www.winkleink.com/2013/04/raspberry-pi-unipolar-stepper-motors.html

lreading:
Does anyone have strong opinions on how the API for the stepper motor should look?

Full disclosure... Most of my knowledge of stepper motors comes from this video: https://www.youtube.com/watch?v=Dc16mKFA7Fo and the little bit I've picked up while using one.  I plan to contribute a class for StepperMotor if this is still wanted, but thought it would be good to solicit early feedback.

I propose the following methods, each of which would accept a number for the total rotations, eg .75, 1, 1.12, -2, etc, where negative numbers would rotate the stepper motor in a "backwards" direction (if possible, which I assume it is):
 - single_step
 - full_step
 - half_step

The constructor accept the following arguments:
 - pins: list of pin numbers to use
 - steps_per_rev: the number of steps required for a full revolution *without gearing*
 - gear_ratio: an integer representing the denominator of the ratio (1/64 gearing = 64)

Is there anything else anyone would like to see as part of this? Is there anything that should be different from the proposal?

fas3r:
Hello, 


I know it's an old topic but I use a stepper motor NEMA 17 JK42HM34-1334 with a L298N controller. I started a class to control forward / backward as follow :

```
import time
import sys
import threading
from gpiozero import OutputDevice
from pprint import pprint


class stepMotors:
  def __init__(self, pins=[6,13,19,26]):
    self.motorBase = []
    self.pins = pins
    for pin in self.pins:
      self.motorBase.append(OutputDevice(pin))

    self.seq = [
        [1,0,1,0],
        [1,0,0,1],
        [0,1,0,1],
        [0,1,1,0]
    ]
    self.state = False

  def cleanup(self):
    if self.thread.is_alive():
      self.state = False
      self.thread.join()

    for GpioOutputDevice in self.motorBase:
      GpioOutputDevice.off()

    return True


  def forward(self):
    if self.state:
      self.cleanup()

    self.direction = 1
    self.stepCounter = 0
    self.state = True
    self.thread = threading.Thread(target=self.run, args=())
    self.thread.daemon = True
    self.thread.start()


  def backward(self):
    if self.state:
      self.cleanup()

[...]

PR #757, opened 2019-03-17
Add StepperMotor

m-alzam:
Hello:
This is a suggestion to add a StepperMotor class to the library. It is based on [this code](https://github.com/RPi-Distro/python-gpiozero/issues/144#issuecomment-454507250) with some improvements and tested using [this motor](http://www.savagehomeautomation.com/projects/raspberry-pi-stepper-motor-control-breakout-board.html).

Some examples:

```python
from time import sleep
from gpiozero import StepperMotor

motor = StepperMotor(pins=(18,22,17,27), sequence=((1, 0, 1, 0),(0, 1, 1, 0),(0, 1, 0, 1),(1, 0, 0, 1)), 
delay=4/1000)
motor.forward(100) # Move 100 steps
motor.backward() # Move unlimited time
sleep(3)
motor.cleanup() # Stop motor and shut down pins

with StepperMotor(pins=(18,22,17,27), sequence=((1, 0, 1, 0),(0, 1, 1, 0),(0, 1, 0, 1),(1, 0, 0, 1)), delay=4/1000) as mt:
    mt.forward()
    sleep(1)
    mt.backward()
    sleep(1)
```


lurch:
Thanks. Just some quick feedback (without having looked at the underlying code, which I'll try to do later...)

IMHO having an `__init__` method that requires the user to specify the `pins` as a list and the `sequence` as a list-of-lists isn't very user-friendly and doesn't feel like the same level of abstraction as the rest of GpioZero. I don't know much about stepper motors but when/why would the user need to provide a different `sequence`? Would it make sense to also provide e.g. a `HMISoftwareStepperDriver` class as a wrapper around this `StepperMotor` class, which has e.g. `pinX1_0`, `pinX1_1`, `pinX1_2`, `pinX1_3` as init-params? (if you're feeling adventurous you could implement this as a `CompositeDevice` of two `StepperMotor`s)

I'm not sure about a `.forward(steps)` method as it conflicts with the `.forward(speed)` method of the existing [Motor class](https://gpiozero.readthedocs.io/en/stable/api_output.html#motor). Would it be possible (or make sense) to implement a `.forward(speed)` method for compatibility, and add a separate method for specifying the number of steps? :man_shrugging:

m-alzam:
Hello

I've rename the `forward` and `backward` functions so you can set the speed of the motor relative to its initial velocity. Now, if you want to move a certain number of steps, you need to call the `forward_steps` function.

For a user-friendly `__init__` method, I've added the `BasicStepperMotor` class with predefined attributes. You can create another classes for others motors that uses another configurations.

Now it is the example code:

```python
from time import sleep
from gpiozero import BasicStepperMotor

motor = BasicStepperMotor() # Set pins=() if you are using another.
motor.forward_steps(100) # Move 100 steps
motor.backward(2) # Move unlimited time at x2 speed
sleep(3)
motor.cleanup() # Stop motor and shut down pins.
```

lurch:
Thanks.
```python
motor.backward(2)
```
If you look at the docs for the existing Motor class you'll see that the speed ranges from 0 (stopped) to 1 (full speed), with values inbetween indicating different speeds (e.g. 0.5 for half maximum speed). Would it be possible for you to tweak your code to match that "API" too?
And I don't think you need an explicit `cleanup()` method, IIRC GpioZero automatically calls any `close()` method on your class when it garbage-collects the object.

m-alzam:
Hello
I was thinking about the `speed` value when I wrote the code. To control de speed of the motor, you change the `delay` attribute. But if you want to move the motor at its half maximum speed, you need to know the maximum delay the motor can have. If you use a huge delay, the motor won't move. I can add a new attribute to the `StepperMotor` class to define its min and max delay time. Is it a good option? What do you think? Thanks.

m-alzam:
Hello
`StepperMotor` class now match "API". That's the code:

```python
from time import sleep
from gpiozero import StepperMotor, BasicStepperMotor

motor = BasicStepperMotor()
motor.forward(speed=0.1)
sleep(2)
motor.backward()
sleep(2)
motor.stop() # Stop motor
```

> And I don't think you need an explicit cleanup() method

Sorry, it's for the first example I wrote, which uses later the same pins for other test.

PR #765, opened 2019-04-24
Should work with revision 100000e

DominikHerold:
I have a PI with revision _100000e_. Which does not work with the current version, because I get following exception:
`unknown old-style revision "100000e"` 👉 https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/pins/data.py#L1071
Therefor I added an additional line with the key `16777230` because of 👉 https://github.com/RPi-Distro/python-gpiozero/blob/master/gpiozero/pins/data.py#L1338
I looks a bit weird, but it _should_ ™️ work


lurch:
Hmmm, looks like a duplicate of #390 i.e. this seems to be a regression. Should hopefully be an easy fix so I'll take a look at it tonight or tomorrow.

Bonno:
Just to add extra information. Also breaks for rev 1000002. Looks like the overvolt check is removed from data.py somewhere in time.

lurch:
Finally got round to digging out an old Pi which has been overvolted, and figured out what was going wrong. I believe https://github.com/RPi-Distro/python-gpiozero/commit/433023903acd2a3b68f2f818b48f6b968e15ee17 is what broke things (by checking `/proc/device-tree` _before_ `/proc/cpuinfo`).
So I'm going to close this PR in favour of #769

PR #771, opened 2019-06-02
Don't allow NativePin to have a 'reserved' pull-up

lurch:
This prevents this (IMHO wrong) behaviour:
```python3
>>> from gpiozero import Device, Button
>>> from gpiozero.pins.native import NativeFactory
>>> Device.pin_factory = NativeFactory()
>>> b = Button(16)
>>> b.pin.pull
'up'
>>> b.pin.pull = 'reserved'
>>> b.pin.pull
'reserved'
```

### DIFF
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -359,7 +359,6 @@ class NativePin(LocalPiPin):
         'up':       0b10,
         'down':     0b01,
         'floating': 0b00,
-        'reserved': 0b11,
         }
 
     GPIO_FUNCTION_NAMES = {v: k for (k, v) in GPIO_FUNCTIONS.items()}


bennuttall:
@lurch what does reserved mean in this context? Why shouldn't NativePin allow it?

lurch:
Apologies for the lack of explanation.
`reserved` is an invalid pull-state value, and doesn't mean anything. `.pull` should only ever be up, down or floating.
I _think_ it was only added so that `0b11` had an entry in the dictionary (?), but I think (IMHO) that this is invalid behaviour.
(IIRC I noticed this when I was working on the enhanced version of #777 that we've been discussing over email)

lurch:
Closing, as this fix is included in #900

PR #777, opened 2019-06-24
Add support for RPi4 to native library

TheEnbyperor:


bennuttall:
Thank you. LGTM. @waveform80 will take a look tomorrow :+1:

lurch:
While this is a noble effort, I'm afraid it won't actually work properly.
A working patch is already being discussed internally...

Misiu:
I think Pi4 is supported (https://github.com/gpiozero/gpiozero/commit/615400ae7ffefb660668d2414a7cf9ae363670dd#diff-d51cee55d16c76a12beddd2e360b8367).
If yes then maybe this can be closed?

bennuttall:
@lurch does native pin work on Pi 4, or need some changes?

lurch:
The pull-up / pull-down won't work for NativePin on the Pi 4. I've sent you an email...

Misiu:
Having support for PI4 would allow many users to replace RPI.GPIO for doorbell scenarios, as suggested here: https://github.com/home-assistant/core/issues/10498#issuecomment-607899990

Misiu:
@lurch @bennuttall any updates?
My Pi4 is on the way so I can help with testing.

bennuttall:
@lurch can you PR your patch when you get chance?

lurch:
> @lurch can you PR your patch when you get chance?

Okay, I'll try to remember to do that this weekend.

Misiu:
@lurch any updates on this? thanks :)

stuartsoft:
@lurch @Misiu any updates? I'd like to use the native pin factory on my RPI4

waveform80:
Thanks very much for this PR, and my apologies it's taken so long for me to get to. Looking at this and the changes in lurch's subsequent PR in #900 unfortunately I do think I'll have to close this without merging, as the situation looks a bit more complicated than this patch currently allows for and #900 appears to have the correct logic (though I need to do some proper testing on it).

Even so, patches like this are still useful as they give me another perspective on the implementation, and at the very least indicate that the native pin-driver seems to have garnered more use than I thought it would! I honestly thought it'd be the driver of last resort that nobody would ever willingly use, but it seems having a pure Python driver has actually been a useful tool in some circumstances! Anyway, many thanks for the patch - and sorry I have to close it without merging, but it was much appreciated nonetheless!

PR #779, opened 2019-06-27
Clearer error message for invalid bounce values

kyrogon:
The factory.connection.set_glitch_filter function from pigpio has a maximum steady parameter of [0.3 seconds.](https://abyz.me.uk/rpi/pigpio/pdif2.html#set_glitch_filter) Enforce that limit here to eliminate the more confusing "PI_BAD_FILTER" error from the pigpio library.


bennuttall:
LGTM. @waveform80?

PR #781, opened 2019-07-13
Warn users when using default pin factory for Servos and Distance Sensors, close #780

### LINKED ISSUE #780: Warn users when using default pin factory for servos
When someone uses the Servo class with gpiozero using the default settings (i.e. default pin factory RPi.GPIO) they get a bad experience and don't know that it's easily fixable, by using the pigpio pin factory.

We should use a warning (like we do with e.g. spi fallback) when someone tries to use Servo without pigpio.

@evildmp

SofiiaKosovan:
#780

### DIFF
--- a/gpiozero/exc.py
+++ b/gpiozero/exc.py
@@ -192,7 +192,10 @@ class SPIWarning(GPIOZeroWarning):
     "Base class for warnings related to the SPI implementation"
 
 class SPISoftwareFallback(SPIWarning):
-    "Warning raised when falling back to the software implementation"
+    "Warning raised when falling back to the SPI software implementation"
+
+class PWMSoftwareFallback(SPIWarning):
+    "Warning raised when falling back to the PWM software implementation"
 
 class PinWarning(GPIOZeroWarning):
     "Base class for warnings related to pin implementations"
@@ -211,3 +214,4 @@ class CallbackSetToNone(GPIOZeroWarning):
 
 class AmbiguousTone(GPIOZeroWarning):
     "Warning raised when a Tone is constructed with an ambiguous number"
+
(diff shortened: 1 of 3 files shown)


bennuttall:
Thank you @SofiiaKosovan!

Assigning to @waveform80 to review this

gpiozero/exc.py
```diff
@@ -192,7 +192,10 @@ class SPIWarning(GPIOZeroWarning):
     "Base class for warnings related to the SPI implementation"
 
 class SPISoftwareFallback(SPIWarning):
-    "Warning raised when falling back to the software implementation"
+    "Warning raised when falling back to the SPI software implementation"
+
+class PWMSoftwareFallback(SPIWarning):
```
lurch: I'm not sure if it (logically) makes sense for `PWMSoftwareFallback` to be a subclass of `SPIWarning` ? :man_shrugging:

gpiozero/output_devices.py
```diff
@@ -1528,6 +1538,13 @@ def __init__(
             ),
             pin_factory=pin_factory
         )
+        
+        if PiGPIOFactory is None or not isinstance(self.pin_factory, PiGPIOFactory):
+            warnings.warn(PWMSoftwareFallback(
+                'To reduce servo jitter, use the pigpio pin factory.'
+                'See https://gpiozero.readthedocs.io/en/stable/api_output.html#servo for more info'
```
lurch: It seems that the documentation being pointed to doesn't mention PiGpio at all? Maybe this PR should also update the associated documentation?

gpiozero/exc.py
bennuttall: True

gpiozero/output_devices.py
bennuttall: A note was added to the Servo docstring in this PR :)

Issue #381, opened 2016-06-16
Add __version__ attribute to gpiozero


SteveAmor:
Is it sensible to add the `__version__` attribute to gpiozero so you can use `gpiozero.__version__` to return the current version?

bennuttall:
We used to have one ([see?](https://github.com/RPi-Distro/python-gpiozero/commit/a07f89c5619654356d8101be19edbde2e9f3cff9)) but @waveform80 removed it in favour of the method described [here](http://picamera.readthedocs.io/en/release-1.10/faq.html#how-can-i-tell-what-version-of-picamera-i-have-installed).

lurch:
Maybe that "How can I tell what version of XXXX I have installed?" could be added to the gpiozero documentation?
Re-opening this issue as a 'reminder'.

bennuttall:
I agree

lurch:
Hmmm, I've just tried that on my Pi, and both the apt-get installed version and the latest-from-github version report `1.2.0`, which isn't very helpful :-/

( `[gpiozero 1.2.0 (/usr/lib/python2.7/dist-packages)]` and `[gpiozero 1.2.0 (/home/pi/github/lurch/python-gpiozero)]` )

bennuttall:
True. We could introduce a policy of changing the version number to `1.4.0-dev` or something as soon as v1.3 is released?

SteveAmor:
Yes please. A documentation update would be useful.  I'm happy to create a PR (under Notes?)

SteveAmor:
PR #383 added

waveform80:
My reason for not bothering (at least in my other projects like picamera, picraft, etc.) goes something like this (my apologies if this comes across as blunt - I don't mean it to be, but it's late and my diplomacy filters aren't working :)

* If you're using a GitHub master clone, I assume you're working in a virtualenv, in which case you know when you're using the master clone because the bash prompt tells you you're in the virtualenv
* If you're using a GitHub master clone and you're *not* in a virtualenv ... you need to learn virtualenv and start using it
* If you're using a GitHub master clone and you *refuse* to use a virtualenv ... you probably deserve all the versioning pain you're encountering ;)

Still, you're quite right that bumping the version after release isn't that big of a deal; it's easy to do and it'll work around this problem, so this gets a thumbs up from me.

lurch:
Closing, as the associated PR as now been merged.

PR #787, opened 2019-08-03
Embed pep-0008 / pep-0396 version number

clach04:
Sample:

    Python 2.7.10 (default, May 23 2015, 09:40:32) [MSC v.1500 32 bit (Intel)] on win32
    >>> import gpiozero
    >>> gpiozero.__version__
    u'1.5.1'
    >>> gpiozero.version
    u'1.5.1'

### DIFF
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -144,3 +144,6 @@ from .internal_devices import (
     TimeOfDay,
     DiskUsage,
 )
+
+version_tuple = (1, 5, 1)
+version = version_string = __version__ = '%d.%d.%d' % version_tuple
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,8 @@ elif sys.version_info[0] == 3:
 else:
     raise ValueError('Unrecognized major version of Python')
 
+import gpiozero
+
 HERE = os.path.abspath(os.path.dirname(__file__))
 
 # Workaround <http://www.eby-sarna.com/pipermail/peak/2010-May/003357.html>
@@ -24,7 +26,7 @@ except ImportError:
     pass
 
 __project__      = 'gpiozero'
-__version__      = '1.5.1'
+__version__      = gpiozero.__version__
 __author__       = 'Ben Nuttall'
 __author_email__ = 'ben@raspberrypi.org'
 __url__          = 'https://github.com/RPi-Distro/python-gpiozero'


lurch:
Out of curiosity, how does this interact with https://gpiozero.readthedocs.io/en/stable/faq.html#how-can-i-tell-what-version-of-gpiozero-i-have-installed ?

lurch:
Ahhh, I _thought_ I remembered this cropping up before... see #381

bennuttall:
We tend to do it this way in projects now:

https://github.com/piwheels/piwheels/blob/master/setup.py#L18
https://github.com/piwheels/piwheels/blob/master/piwheels/__init__.py#L70

So maybe we should move to that style here.

clach04:
> Out of curiosity, how does this interact with https://gpiozero.readthedocs.io/en/stable/faq.html#how-can-i-tell-what-version-of-gpiozero-i-have-installed ?

No change. You can still use `setuptools` and have that meta data available. The differences are :

  * that this also works without requiring setup tools
  * for maintenance; edit the module rather than the setup.py (which means when the module changes, the version number is changed inside the same file, rather than needing to open another file

clach04:
> We tend to do it this way in projects now:
> 
> https://github.com/piwheels/piwheels/blob/master/setup.py#L18
> https://github.com/piwheels/piwheels/blob/master/piwheels/__init__.py#L70
> 
> So maybe we should move to that style here.

I only moved version information, I wanted to see if there was an appetite for this approach. I don't feel strongly that the other meta data should move (note I've moved most meta data into modules I personally maintain so its all in one place).

I do feel strongly that both version string and version tuple should be available, I noticed that in the link above the Python version tuple was made use of :) https://github.com/piwheels/piwheels/blob/master/setup.py#L6 (note this is called `version_info`, I used `version_tuple`). Using a tuple means that:

    In [4]: '1.0.2' > '1.0.10'
    Out[4]: True

    In [5]: (1, 0, 2) > (1, 0, 10)
    Out[5]: False

behaves sanely, as does:

    In [12]: (1, 0, 1) > (1, 0, 0, 'dev')
    Out[12]: True

    In [13]: (1, 0, 1) > (1, 0, 1, 'dev')
    Out[13]: False

    In [14]: (1, 0, 0) >= (1, 0, 1, 'dev')
    Out[14]: False

    In [15]: (1, 0, 1) >= (1, 0, 1, 'dev')
    Out[15]: False

    In [16]: (1, 0, 2) > (1, 0, 1, 'dev')
    Out[16]: True

whilst avoiding shenanigans with:

    # from https://www.python.org/dev/peps/pep-0440/
    tuple(map(int, release_segment.split(".")))

which then fail if non-numeric version entries are present.

ukBaz:
For reference I looked up the python packaging documentation:
https://packaging.python.org/guides/single-sourcing-package-version/

It does post the following warning about this technique

> Warning
> 
> Although this technique is common, beware that it will fail if sample/__init__.py imports packages from install_requires dependencies, which will very likely not be installed yet when setup.py is run.

bennuttall:
> It does post the following warning about this technique


Yes I found that out on the first round of building for piwheels - many projects failed to run setup.py for this reason! I think it's safe if there's only metadata in your `__init__.py`

bennuttall:
@waveform80 do you have any thoughts? Should we move to the piwheels way?

bennuttall:
I have a feeling this won't work because running `__init__` with `setup.py` will inadvertently import stuff that might make it fail, due to missing pin factory and things.

waveform80:
> I have a feeling this won't work because running __init__ with setup.py will inadvertently import stuff that might make it fail, due to missing pin factory and things.

Precisely - there's a lot of stuff that implicitly happens when gpiozero is imported, which is why I really don't want to do that in setup.py. It is possible to work around this by "parsing" the `__init__.py` from `setup.py` instead of importing/executing it but firstly this results in a complex `setup.py` (something I also tend to avoid) and secondly this usually restricts what can go in `__init__.py` (which can come back to bite you later).

(Aside: incidentally, I *do* tend to include the version in this way when building applications (as opposed to libraries) but that's because the root `__init__.py` in those tend to be entirely empty because setuptools' entry_points are used to launch things.)

Anyway, I'm afraid I don't see the trade-off of having `__version__` as useful enough to go for such a work-around in the case of gpiozero.

bennuttall:
We *could* add `__version__` to `__init__.py` but we *can't* import it into `setup.py`, so there's no guarantee it's correct. However, it could be part of the release procedure, or maybe the tests.

Or... could we create, say `meta.py` in the module dir, add `__version__` and things there, and import them into `setup.py` instead of `__init__.py`?

ukBaz:
I use a little utility called bumpversion that updates the version string in the files I specify. This might be helpful if you go with the above idea:
https://pypi.org/project/bumpversion/
some recommend bump2version:
https://medium.com/@williamhayes/versioning-using-bumpversion-4d13c914e9b8

waveform80:
> We _could_ add `__version__` to `__init__.py` but we _can't_ import it into `setup.py`, so there's no guarantee it's correct. However, it could be part of the release procedure, or maybe the tests.
> 
> Or... could we create, say `meta.py` in the module dir, add `__version__` and things there, and import them into `setup.py` instead of `__init__.py`?

Nope, importing any module under gpiozero will implicitly import __init__.py too - that's just how the importer machinery works (traverses the hierarchy of modules above the one you're importing). For now, this won't work unless we want to incorporate some parsing machinery into setup.py which I'm not terribly keen on. Once we ditch 2.7 support in 2.x, there are actually some ways around this (using some of the built-in parsing facilities in setup.cfg), but for now this'll have to wait.

septatrix:
For newer version this is not required anymore as the `importlib` module from the standard library provides this functionality since Python 3.8. For gpiozero the following already works:

```python3
>>> import importlib.metadata
>>> importlib.metadata.version("gpiozero")
'1.5.1'
>>> tuple(int(x) for x in importlib.metadata.version("gpiozero").split("."))
(1, 5, 1)
```

PR #796, opened 2019-09-17
Correct usage of warnings!

bennuttall:
Turns out we've been doing warnings wrong this whole time...

What we've been doing:

```python
warnings.warn(SomeWarning(some_message))
```

Correct usage:

```python
warnings.warn(some_message, SomeWarning)
```

https://docs.python.org/3/library/warnings.html#warnings.warn
https://docs.python.org/2/library/warnings.html#warnings.warn


lurch:
Wow, how come this was never noticed before?

bennuttall:
No idea! Guess a warning was still being raised, so the tests were catching them.

I only noticed when I realised the output was coming out duplicated...

@waveform80

bennuttall:
Dave points out that the docs page says:

> Alternatively message can be a Warning instance

And so I think it's fine to leave it as it is.

lurch:
I _did_ find it surprising that @waveform80 would make such an egregious error :wink:

bennuttall:
I still think there's something up though. I've noticed some of our warnings don't look right, and I've realised it's when the warning message spans multiple lines.

To be honest, when the warning message is repeated, it doesn't look right.

See these examples: https://gist.github.com/bennuttall/a461009c39a779e4361e2a3881708233

I think my preferred option is this:

```
w4.py:8: MyWarning: This is a warning
  warnings.warn(message, MyWarning)
```

lurch:
Isn't that just a side-effect of the way that the Python interpreter reports exceptions? :man_shrugging:

bennuttall:
Yes, but seeing here that we have a choice of how we write them to optimise for what's reported, I thought it would be a good opportunity to clean it up and make it nicer.

lurch:
Ahhh, so I guess you're basically asking for e.g.
```python
                except Exception as e:
                    warnings.warn(
                        PinFactoryFallback(
                            'Falling back from %s: %s' % (name, str(e))))
```
to be re-written as:
```python
                except Exception as e:
                    warnings.warn(PinFactoryFallback('Falling back from %s: %s' % (name, str(e))))
```
(i.e. temporarily ignoring the line-length limits in order to have Python's line-based exception reporting  produce nicer output?)

Works for me :+1:  (if I've understood correctly)

bennuttall:
Not specifically that, I just think we need to look at these as examples and see if we can improve what's printed in the exception or warnings in general, and try to choose wisely :)

I actually don't like the way they come out, but that's just how Python does it.

lurch:
I've not tested it, but I believe the change I suggested above would change your example from:
```
/home/pi/.virtualenvs/gpiozero/lib/python3.7/site-packages/gpiozero/devices.py:282: PinFactoryFallback: Falling back from rpigpio: No module named 'RPi'
'Falling back from %s: %s' % (name, str(e))))
```
to
```
/home/pi/.virtualenvs/gpiozero/lib/python3.7/site-packages/gpiozero/devices.py:282: PinFactoryFallback: Falling back from rpigpio: No module named 'RPi'
warnings.warn(PinFactoryFallback('Falling back from %s: %s' % (name, str(e))))
```

Maybe I'm still misunderstanding what you're suggesting?

PR #797, opened 2019-09-17
Add pulse_width setter to Servo class

### LINKED ISSUE #795: Set pulse_width on Servo / PWMOutputDevice
It's currently not possible to set the pulse width of a servo or PWM device. There's a `pulse_width` property in `Servo`, but only a getter, not a setter. I added one and tested it with @evildmp and it seemed to work:

```python
@property
def pulse_width(self):
    if self.pwm_device.pin.frequency is None:
        return None
    else:
        return self.pwm_device.pin.state * self.frame_width

@pulse_width.setter
def pulse_width(self, value):
    self.pwm_device.pin.state = value / self.frame_width
```

I think this could be added lower down, in `PWMOutputDevice`.

bennuttall:
Partially fixes #795 - adds pulse_width setter to Servo class. Could we add a getter/setter to `PWMOutputDevice` too? It's not immediately obvious to me how.

@evildmp

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -1486,13 +1486,13 @@ class Servo(SourceMixin, CompositeDevice):
         servo = Servo(17)
 
         servo.value = 0.5
-    
+
     .. note::
 
         To reduce servo jitter, use the pigpio pin driver rather than the default
         RPi.GPIO driver (pigpio uses DMA sampling for much more precise edge
         timing). See :ref:`changing-pin-factory` for further information.
-    
+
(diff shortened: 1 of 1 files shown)


lurch:
> Could we add a getter/setter to PWMOutputDevice too?

I wouldn't think so, given that `._frame_width` is _only_ in Servo? :man_shrugging:  But I'm probably misunderstanding the question...

bennuttall:
> But I'm probably misunderstanding the question...

No, you are correct (in that that's the reason it's not straightforward)

PR #773, opened 2019-06-18
Add Pibrella

ForToffee:
Using JamHat as template, added Pibrella so can be used remotely via PiZero GPIOExtender

PR #798, opened 2019-09-17
Add Pibrella class

bennuttall:
Building on the excellent start made by @fortoffee in #773...

- lights is now a `TrafficLights` instance
- Removed `LEDBoard` & `ButtonBoard` instances
- Provided input and output pins as namedtuples of pins
- Provided example of using input and output pins

New example usage:

```python
from gpiozero import Pibrella

pb = Pibrella()

pb.button.wait_for_press()
pb.lights.on()
pb.buzzer.play('A4')
pb.off()
```

```python
from gpiozero import Pibrella, LED, Button

pb = Pibrella()
btn = Button(pb.inputs.a)
led = LED(pb.outputs.e)

btn.when_pressed = led.on
```

TODO:

- improve docstrings e.g. link namedtuple to python docs
- add tests

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -2313,7 +2313,7 @@ class Pibrella(CompositeOutputDevice):
 
         .. attribute:: e, f, g, h
     """
-    def __init__(self, pwm=True, pin_factory=None):
+    def __init__(self, pwm=False, pin_factory=None):
         super(Pibrella, self).__init__(
             lights=TrafficLights(red=27, amber=17, green=4, pwm=pwm,
                               pin_factory=pin_factory),
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -2273,9 +2273,9 @@ class Pibrella(CompositeOutputDevice):
         btn.when_pressed = led.on
 
     :param bool pwm:
-        If :data:`True` (the default), construct :class:`PWMLED` instances to
-        represent each LED on the board. If :data:`False`, construct regular
-        :class:`LED` instances.
+        If :data:`True`, construct :class:`PWMLED` instances to represent each
+        LED on the board, otherwise if :data:`False` (the default), construct
+        regular :class:`LED` instances.
 
     :type pin_factory: Factory or None
     :param pin_factory:
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -2181,7 +2181,7 @@ class JamHat(CompositeOutputDevice):
         hat.off()
 
     :param bool pwm:
-        If :data:`True`, construct :class: PWMLED instances to represent each
+        If :data:`True`, construct :class:`PWMLED` instances to represent each
         LED on the board. If :data:`False` (the default), construct regular
         :class:`LED` instances.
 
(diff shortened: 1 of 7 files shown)


gpiozero/boards.py
```diff
@@ -2239,3 +2239,104 @@ def off(self):
...
+
+        .. attribute:: a, b, c, d
+
+    .. attribute:: outputs
+
+        A namedtuple of the output pin numbers
+
+        .. attribute:: e, f, g, h
+    """
+    def __init__(self, pwm=True, pin_factory=None):
```
lurch: Similar classes default `pwm` to False - why does this one default it to True?

gpiozero/boards.py
```diff
@@ -2239,3 +2239,104 @@ def off(self):
...
+            lights=TrafficLights(red=27, amber=17, green=4, pwm=pwm,
+                              pin_factory=pin_factory),
+            button=Button(11, pull_up=False, pin_factory=pin_factory),
+            buzzer=TonalBuzzer(18, pin_factory=pin_factory),
+            _order=('lights', 'buzzer', 'button'),
+            pin_factory=pin_factory
+        )
+        InputPins = namedtuple('InputPins', ['a', 'b', 'c', 'd'])
+        OutputPins = namedtuple('OutputPins', ['e', 'f', 'g', 'h'])
+        self.inputs = InputPins(a=9, b=7, c=8, d=10)
```
lurch: I _think_ you might be able to create the actual `Pin` objects here (whilst using the appropriate `pin_factory`), rather than just storing the gpio numbers?

bennuttall: I did think of that, but couldn't remember if a Pin object is a valid pin parameter any more.

gpiozero/boards.py
bennuttall: Good spot, should be `False`

gpiozero/boards.py
lurch: ping @waveform80

bennuttall: Looks like it's not. You need to pass in a number or string, and `pin_factory` (optional) so I don't think `pb.outputs.e` being an instance of `RPiGPIOPin` is useful, as you can't create a device object with it.

lurch: Fair enough. The "pins" stuff has gone through so many changes, it's hard to remember what is and isn't supported. And I just tried having a quick look at the sourcecode just now, and there's so many abstract classes, factory classes, super method calls, etc. etc. that I'm still not able to follow exactly what's going on :-)

gpiozero/boards.py
lurch: Don't forget to fix this before merging...

bennuttall: Thanks - done

lurch: Need to update the docstring too :stuck_out_tongue_winking_eye:

bennuttall: Oh nearly missed that - done

bennuttall:
I think that's everything now - I'll merge once the tests are done.

PR #800, opened 2019-09-19
Pin factory docs

bennuttall:
I noticed the docs on the default pin factory and changing the pin factory were outdated. `Device.pin_factory` is now `None` until construction of a device, so that `import gpiozero` works without a valid pin factory. This means the device pin flowchart changes from [this](https://gpiozero.readthedocs.io/en/v1.5.1/_images/device_pin_flowchart.svg) to this:

device_pin_flowchart

Full docs page changes here: https://gpiozero.readthedocs.io/en/pin-factory-docs/api_pins.html


lurch:
Looking at https://github.com/gpiozero/gpiozero/blob/master/gpiozero/devices.py#L241 your flowchart seems to be wrong... if `Device.pin_factory is None` it does `Device.pin_factory = Device._default_pin_factory()` (and then `self.pin_factory = Device.pin_factory`); whereas your flowchart says `self.pin_factory = Device._default_pin_factory()` - I think it's worth correcting the flowchart to illustrate that `Device.pin_factory` only gets set once.

There's something wrong with the grammar in "The majority of the documentation focuses on devices as pins are below the level that most users are concerned with." but I'm not quite sure what it's _supposed_ to be saying.

"an BadPinFactory" should probably be "a BadPinFactory exception" ?

`>>> Device.pin_factory()` probably doesn't want brackets on the end?

Some examples have
```bash
pi@raspberrypi:~ $ python3
Python 3.7.3 (default, Apr  3 2019, 05:39:12)
```
and others have
```
ben@magicman:~ $ python3
Python 3.6.8 (default, Aug 20 2019, 17:12:48)
```
Should they be made consistent?

In the example where you do `pi@raspberrypi:~ $ GPIOZERO_PIN_FACTORY=native python3` why does it then say `RPiGPIOFactory` ??

I've not tested it myself, but in the example where you do `export GPIOZERO_PIN_FACTORY=native` shouldn't `gpiozero.Device.pin_factory` still be `None`, based on earlier comments?


That'll do for now, I can hear my bed calling :wink:  I may read through the rest of it tomorrow.

bennuttall:
> I think it's worth correcting the flowchart to illustrate that Device.pin_factory only gets set once.

Ah yes, good point. Like this?

device_pin_flowchart

> "an BadPinFactory" should probably be "a BadPinFactory exception" ?

Corrected

> Device.pin_factory() probably doesn't want brackets on the end?

Corrected

> Some examples have pi@raspberrypi:~ $ python3 and others have ben@magicman:~ $ python3 Should they be made consistent?

There's one example where it's on a PC, that's all.

> In the example where you do pi@raspberrypi:~ $ GPIOZERO_PIN_FACTORY=native python3 why does it then say RPiGPIOFactory ??

Corrected

> I've not tested it myself, but in the example where you do export GPIOZERO_PIN_FACTORY=native shouldn't gpiozero.Device.pin_factory still be None, based on earlier comments?

I don't see any instances of that. I use `Device._default_pin_factory()` not `Device.pin_factory`.

> That'll do for now, I can hear my bed calling wink I may read through the rest of it tomorrow.

Thanks!

PR #824, opened 2020-01-06
Update migrating_from_rpigpio.rst

damosurfer:
The example for PWM migration should be led.value = b / 100.0 instead of led.value = b / 100 since range(100) yields integers and dividing by the literal int 100 will always result in 0.

It would also be best to change range(100) to range(101) to obtain the full range from 0.0 to 1.0 since the last step with range(100) would be 99/100.

Thank you for this great library!

### DIFF
--- a/docs/migrating_from_rpigpio.rst
+++ b/docs/migrating_from_rpigpio.rst
@@ -304,8 +304,8 @@ In GPIO Zero::
 
     led = PWMLED(2)
 
-    for b in range(100):
-        led.value = b / 100
+    for b in range(101):
+        led.value = b / 100.0
         sleep(0.01)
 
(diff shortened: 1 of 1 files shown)


bennuttall:
Thanks

> The example for PWM migration should be led.value = b / 100.0 instead of led.value = b / 100 since range(100) yields integers and dividing by the literal int 100 will always result in 0.

Only in Python 2, which is EOL.

> It would also be best to change range(100) to range(101) to obtain the full range from 0.0 to 1.0 since the last step with range(100) would be 99/100.

Correct

lurch:
Should the `for dc in range(100):` line in the RPi.GPIO example also be using `range(101)` for the same reason? :thinking:

bennuttall:
I think so

PR #831, opened 2020-01-20
Add is_lit alias to LEDCollection, close #822

### LINKED ISSUE #822: is_lit equivalent for LEDBoard?
**Short description of change:** Add _**is_lit**_ type attribute to LEDBoard class
**Where would this change be made:** LEDBoard

**Purpose of change:**
As I've been reading the docs, I see that in the LED class, there's a _**is_lit**_ attribute which can be used to check conditions, but is there an equivalent for LEDBoard? 

Context is I've got a series of lights in a LEDBoard and I I'd like to be able to check if it is active (lit) so I can control on/off. 

I can see that LED and PWMLED supports _**is_lit**_ 
But I'm using the pulse setting in the LEDBoard LEDs so I can simulate certain effects across the LEDs

But it seems if I want to be able to control the pulse, and check it's state, I need to use LED/PWMLED vs LEDBoard as LEDBoard does not have an equivalent. 

Please provide example usage of gpiozero with your suggested change: 

```python
body_leds = LEDBoard(6, 5, 13, 19, 16, 26, 20, 21, pwm=True)
hull_led = LEDBoard(18, pwm=True)
left_red_thruster = LEDBoard(27, pwm=True)
left_green_thruster = LEDBoard(22, pwm=True)
right_red_thruster = LEDBoard(4, pwm=True)
right_green_thruster = LEDBoard(17, pwm=True)
boosters = LEDBoard(24, 25, pwm=True)

while True:
	if static_button.is_pressed:
		if body_leds.is_lit:
			leds_off()
		else:
			static_display()
```
Thanks

bennuttall:
#822

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -358,6 +358,9 @@ class LEDCollection(CompositeOutputDevice):
     @property
     def active_high(self):
         return self[0].active_high
+    
+    
+LEDCollection.is_lit = LEDCollection.is_active
 
 
 class LEDBoard(LEDCollection):


bennuttall:
Oh god, no. This fix is correct but it looks like `is_active` is broken...

```python
In [1]: from gpiozero import *                                                  

In [2]: sp = SnowPi()                                                           

In [3]: sp.is_active                                                            
Out[3]: True
```

bennuttall:
Merged because there's nothing wrong with the PR, just that it needs to be fixed underneath.

Opened a new bug #832

lurch:
Why not make LEDCollection.is_lit a new property and use the `any(led.is_lit for led in self.leds)` approach you suggested in #822 ?

bennuttall:
Surely that's what `is_active` should be?

lurch:
Yeah, perhaps the fix does **indeed** belong in CompositeDevice (since there's nothing stopping CompositeDevices being nested) and https://github.com/gpiozero/gpiozero/blob/master/gpiozero/devices.py#L504 needs to be fixed to be recursive ? Perhaps _something_ like:
```python3
    @property
    def is_active(self):
        any_active = False
        for device in self:
            if isinstance(device, CompositeDevice):
                active = device.is_active
            else:
                active = bool(device.value)
            if active:
                any_active = True
                break
        return any_active
```
(that's written entirely off the top of my head, so may or may not work!)

PR #835, opened 2020-02-03
callback after DigitalOutputDevice.blink ends

### LINKED ISSUE #827: callback after DigitalOutputDevice.blink ends
Currently there is no indication, when a DigitalOutputDevice stops blinking because n blinks are done or another state was set.
In order to build complex blinking sequences or simply to set the value after blinking a callback would be helpful.
This callback should be called with a parameter indicating whether the blinking stops after the cycle was completed or the cycle was terminated by another command.

MNeum:
close #827

enhance blink method in DigitalOutputDevice, PWMOutputDevice and RGBLED:
- additional parameter on_end: callback function to call after blink ended
- additional parameter end_state / end_value / end_color: will be set, if a blink sequence was finished uninterrupted

new blink_sequence method in DigitalOutputDevice, PWMOutputDevice and RGBLED runs blinking sequences

additional feature: wait delay is time corrected, so that multiple independently blinking devices stay in sync

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -236,7 +236,7 @@ class DigitalOutputDevice(OutputDevice):
         self._stop_blink()
         self._write(False)
 
-    def blink(self, on_time=1, off_time=1, n=None, background=True):
+    def blink(self, on_time=1, off_time=1, end_state=None, n=None, background=True, on_end=None):
         """
         Make the device turn on and off repeatedly.
 
@@ -245,6 +245,11 @@ class DigitalOutputDevice(OutputDevice):
 
         :param float off_time:
             Number of seconds off. Defaults to 1 second.
+
+        :type end_state: bool or None
+        :param end_state:
+            State to set after a full uninterrupted blink.
+            Defaults to None, on which no state will be set.
 
         :type n: int or None
         :param n:
@@ -255,10 +260,18 @@ class DigitalOutputDevice(OutputDevice):
             continue blinking and return immediately. If :data:`False`, only
             return when the blink is finished (warning: the default value of
             *n* will result in this method never returning).
+
+        :type on_end: Callable[[DigitalOutputDevice, bool], None] or None
+        : param on_end:
+            callback function to call after blink ended with parameters
+                self: the
+                uninterrupted:
+
         """
         self._stop_blink()
+        sequence = [(True, on_time), (False, off_time)]
         self._blink_thread = GPIOThread(
-            target=self._blink_device, args=(on_time, off_time, n)
+            target=self._blink_device, args=(sequence, n, end_state, on_end)
         )
         self._blink_thread.start()
         if not background:
@@ -273,15 +286,20 @@ class DigitalOutputDevice(OutputDevice):
             self._blink_thread.stop()
         self._blink_thread = None
 
-    def _blink_device(self, on_time, off_time, n):
+    def _blink_device(self, sequence, n, end_state, on_end):
         iterable = repeat(0) if n is None else repeat(0, n)
+        uninterrupted = True
         for _ in iterable:
-            self._write(True)
-            if self._blink_thread.stopping.wait(on_time):
-                break
-            self._write(False)
-            if self._blink_thread.stopping.wait(off_time):
-                break
+            for state, wait_time in sequence:
+                self._write(state)
+                if self._blink_thread.stopping.wait(wait_time):
+                    uninterrupted = False
+                    break
+
+        if uninterrupted and end_state is not None:
+            self._write(end_state)
+        if on_end:
+            on_end(self, uninterrupted)
 
 
 class LED(DigitalOutputDevice):
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -40,17 +40,19 @@ from __future__ import (
 )
 str = type('')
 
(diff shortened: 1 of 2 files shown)


MNeum:
Looking through the travis build log, I see that the pypy3 build failed because travis was unable to download pypy3 archive

I hope, this is not serious...

gpiozero/output_devices.py
```diff
@@ -236,7 +240,8 @@ def off(self):
         self._stop_blink()
         self._write(False)
 
-    def blink(self, on_time=1, off_time=1, n=None, background=True):
+    def blink(self, on_time=1, off_time=1, end_state=None,
```
lurch: I think for backwards-compatibility reasons, we prefer to not reorder the args-list. So here you'd need to move `end_state` to after `background` .

gpiozero/output_devices.py
```diff
@@ -324,7 +384,11 @@ class LED(DigitalOutputDevice):
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
     """
-    pass
+    def __init__(self, pin=None, active_high=True, initial_value=False, pin_factory=None):
+        self._blink_thread = None
+        self._controller = None
+        super(LED, self).__init__(pin, active_high, initial_value, pin_factory=pin_factory)
```
lurch: Why is this change needed, when `DigitalOutputDevice`'s `__init__` already sets `_blink_thread` and `_controller` to None? :man_shrugging:

gpiozero/output_devices.py
```diff
@@ -517,6 +583,22 @@ def blink(
...
+        :param float off_value:
+            value during off_time.
+            Defaults to 0 (full off).
+
+        :type end_value: float or None
+        :param end_value:
+            value to set after a full uninterrupted blink.
+            Defaults to None, on which no value will be set.
+
+        :param int fps:
```
lurch: Why add this as an extra parameter? Wouldn't it make more sense to just (internally) use `self.frequency` ?

bennuttall:
Thanks for putting the effort in but I'm not convinced it's necessary. You could subclass to add this feature - I'd consider it API bloat.

gpiozero/output_devices.py
```diff
@@ -526,41 +608,125 @@ def blink(
...
+        self._blink_thread.start()
+        if not background:
+            self._blink_thread.join()
+            self._blink_thread = None
+
+    @staticmethod
+    def interpolate_value(start_value, end_value, delta_t, t):
+        return (start_value * (delta_t - t) + end_value * t) / delta_t
+
+    def _calculate_sequence(self, on_value, fade_in_time, on_time, off_value, fade_out_time, off_time, fps):
```
lurch: Could this also be a `@staticmethod` ?

gpiozero/output_devices.py
```diff
@@ -889,23 +1055,44 @@ class RGBLED(SourceMixin, Device):
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
 
-    .. _colorzero: https://colorzero.readthedocs.io/
+    .. _colorzero: https://colorzero.readtheocs.io/
```
lurch: ???

lurch:
Why is the `end_state` parameter necessary - surely it's easy enough to just set that from the callback function?
(I started reviewing this before realising how big the changes were, so I only got part-way through the review... :wink: )

gpiozero/output_devices.py
MNeum: I didn't think about that. No problem!

gpiozero/output_devices.py
MNeum: It would be possible to use self.frequency. I borrowed this from the _blink_device, where it was a default parameter, but never set on the call.
fps defines, how smooth the transition of the fade will be. 
A value of 25 means, that every 40 ms a new value will be set.  
On a PWM frequency of 100 Hz that means every 4th zero crossing the pulse width will be changed. That is a heavy load already. In most cases a lower value is enough for a smooth eye pleasing transition.

gpiozero/output_devices.py
MNeum: Yes, you are right.

gpiozero/output_devices.py
MNeum: Definitely a bad error...

MNeum:
> Why is the `end_state` parameter necessary - surely it's easy enough to just set that from the callback function?
In my application this is a frequent used feature, whereas the callback is seldom needed (only for sync purposes). 
But you are right, end_state can be set by a callback and the end_set parameter could be removed.

lurch:
> Maybe blinking functionality could be a Mixin?

I had that idea too a few years ago :wink:  I started working on it but then I ran into some problem (I can't remember now what it was - perhaps something to do with RGBLED) and I didn't have enough enthusiasm to pursue the idea further.

> the same problem I try to handle here ... complex rules for switching on leds ... sync several led timings

Sounds to me like you'd probably be better off just creating your own timing threads, and only using gpiozero to turn the LEDs `.on()` and `.off()` as needed, rather than trying to modify the (deliberately limited) `blink_thread` to do everything you need it to? :man_shrugging:

MNeum:
Since there seems no desire to implement this in the core library, I suppose we can close this PR

PR #849, opened 2020-03-30
remove print("here") from def all_values()

darton:
print ("here") is not needed


bennuttall:
😱 

git blame

it was me

oh dear

Thanks!

PR #852, opened 2020-04-18
Change average of MotionSensor to geometric mean instead of median

roberterdin:
To make the `MotionSensor.threshold` configuration work, the average of the sensor values can't be the median, as the value is always either 0 or 1.
For example `[0,0,0,1,1]` is 0 and `[0,0,1,1,1]` is 1. Every threshold value larger than 0 will trigger as soon as the median is 1 and there is no granularity.
There are two potential fixes:
- change the average function of the internal queue from median to geometric mean
- remove the threshold configuration entirely and rely on the queue length

The first option, i.e. using the geometric mean to calculate the average is chosen in the PR to give more control to the user.

### DIFF
--- a/gpiozero/input_devices.py
+++ b/gpiozero/input_devices.py
@@ -45,7 +45,7 @@ import warnings
 from time import sleep, time
 from threading import Event, Lock
 try:
-    from statistics import median
+    from statistics import median, mean
 except ImportError:
     from .compat import median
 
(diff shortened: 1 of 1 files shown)


lurch:
Is this an issue specifically with `MotionSensor`, or does it apply to all `SmoothedInputDevice` s?
ping @waveform80

roberterdin:
Anything that has continuous values should not be affected. 
As a user, you might still want to think about which of the two (mean or median) makes more sense for you but at least you're not under the impression that you are adjusting a threshold when actually you are not.

waveform80:
Good catch - the change to median was originally to benefit things like DistanceSensor that occasionally (due to timing glitches) wind up with ridiculous outliers (that then throw off the mean), but I'd forgotten there was a sub-class with discrete values. Thanks!

PR #860, opened 2020-05-10
Update add-on boards to use BOARD numbering, close #349

### LINKED ISSUE #349: Addon boards compatibility question
Something that occurred to me while I was adding #333 ...

As we all know, GPIOzero _only_ deals with GPIO / Broadcom / BCM numbers, rather than physical pin numbers. This is all fine and helps prevent confusion.

However (by their very nature) add-on boards always attach to fixed physical pin numbers. And some add-ons, such as https://www.piborg.org/ledborg/specs connect to a physical pin that has a different GPIO number on the original Rev1 pinout than [later revisions](https://www.raspberrypi.org/blog/upcoming-board-revision/).
This means that if a Rev1 ModelB Pi owner connects an LedBorg and tries using GpioZero to control it, it won't work properly.

With the ['pins database'](http://gpiozero.readthedocs.io/en/latest/api_pins.html#utilities) @waveform80 added, we could fix this so that [add-on boards](http://gpiozero.readthedocs.io/en/latest/api_boards.html) with fixed physical pins could lookup (at runtime) the appropriate GPIO numbers for the physical pins they actually use.
Alternatively, we could say that the proportion of people still using Rev1 ModelB Pis is so small that it's not worth bothering with? I see for example that http://gpiozero.readthedocs.io/en/latest/recipes.html#pin-numbering doesn't bother mentioning the GPIO numbers of the [Rev1 Pis](http://elinux.org/RPi_Low-level_peripherals#P1_Header).

Comments / opinions welcomed... :-)

bennuttall:
As suggested by @lurch in #349, this branch updates add-on boards to use BOARD numbering so that the right BCM numbers are elected regardless of Pi model.

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1999,10 +1999,10 @@ class _EnergenieMaster(SharedMixin, CompositeOutputDevice):
         super(_EnergenieMaster, self).__init__(
             *(
                 OutputDevice(pin, pin_factory=pin_factory)
-                for pin in (17, 22, 23, 27)
+                for pin in ('BOARD11', 'BOARD15', 'BOARD16', 'BOARD13')
             ),
-            mode=OutputDevice(24, pin_factory=pin_factory),
-            enable=OutputDevice(25, pin_factory=pin_factory),
+            mode=OutputDevice('BOARD18', pin_factory=pin_factory),
+            enable=OutputDevice('BOARD22', pin_factory=pin_factory),
             _order=('mode', 'enable'), pin_factory=pin_factory
         )
 
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -889,9 +889,10 @@ class LedBorg(RGBLED):
     """
 
     def __init__(self, initial_value=(0, 0, 0), pwm=True, pin_factory=None):
-        super(LedBorg, self).__init__(red=17, green=27, blue=22,
-                                      pwm=pwm, initial_value=initial_value,
-                                      pin_factory=pin_factory)
+        super(LedBorg, self).__init__(
+            red='BOARD11', green='BOARD13', blue='BOARD15',
+            pwm=pwm, initial_value=initial_value, pin_factory=pin_factory
+        )
 
 
 class PiLiter(LEDBoard):
@@ -928,9 +929,11 @@ class PiLiter(LEDBoard):
     """
 
     def __init__(self, pwm=False, initial_value=False, pin_factory=None):
-        super(PiLiter, self).__init__(4, 17, 27, 18, 22, 23, 24, 25,
-                                      pwm=pwm, initial_value=initial_value,
-                                      pin_factory=pin_factory)
+        pins = ('BOARD7', 'BOARD11', 'BOARD13', 'BOARD12',
+                'BOARD15', 'BOARD16', 'BOARD18', 'BOARD22')
+        super(PiLiter, self).__init__(
+            *pins, pwm=pwm, initial_value=initial_value, pin_factory=pin_factory
+        )
 
 
 class PiLiterBarGraph(LEDBarGraph):
@@ -964,7 +967,8 @@ class PiLiterBarGraph(LEDBarGraph):
     """
 
     def __init__(self, pwm=False, initial_value=0.0, pin_factory=None):
-        pins = (4, 17, 27, 18, 22, 23, 24, 25)
+        pins = ('BOARD7', 'BOARD11', 'BOARD13', 'BOARD12',
+                'BOARD15', 'BOARD16', 'BOARD18', 'BOARD22')
         super(PiLiterBarGraph, self).__init__(
             *pins, pwm=pwm, initial_value=initial_value, pin_factory=pin_factory
         )
@@ -1107,9 +1111,10 @@ class PiTraffic(TrafficLights):
     .. _Low Voltage Labs PI-TRAFFIC: http://lowvoltagelabs.com/products/pi-traffic/
     """
     def __init__(self, pwm=False, initial_value=False, pin_factory=None):
-        super(PiTraffic, self).__init__(9, 10, 11,
-                                        pwm=pwm, initial_value=initial_value,
-                                        pin_factory=pin_factory)
+        super(PiTraffic, self).__init__(
+            'BOARD21', 'BOARD19', 'BOARD23',
+            pwm=pwm, initial_value=initial_value, pin_factory=pin_factory
+        )
 
 
 class PiStop(TrafficLights):
@@ -1150,12 +1155,12 @@ class PiStop(TrafficLights):
     .. _location: https://github.com/PiHw/Pi-Stop/blob/master/markdown_source/markdown/Discover-PiStop.md
     """
     LOCATIONS = {
-        'A': (7, 8, 25),
-        'A+': (21, 20, 16),
-        'B': (10, 9, 11),
-        'B+': (13, 19, 26),
-        'C': (18, 15, 14),
-        'D': (2, 3, 4),
+        'A': ('BOARD26', 'BOARD24', 'BOARD22'),
+        'A+': ('BOARD40', 'BOARD38', 'BOARD36'),
+        'B': ('BOARD19', 'BOARD21', 'BOARD23'),
+        'B+': ('BOARD33', 'BOARD35', 'BOARD37'),
+        'C': ('BOARD12', 'BOARD10', 'BOARD8'),
+        'D': ('BOARD3', 'BOARD5', 'BOARD7'),
     }
 
     def __init__(
@@ -1221,9 +1226,9 @@ class StatusZero(LEDBoard):
 
     def __init__(self, *labels, **kwargs):
         pins = (
-            (17, 4),
-            (22, 27),
-            (9, 10),
+            ('BOARD11', 'BOARD7'),
+            ('BOARD15', 'BOARD13'),
+            ('BOARD21', 'BOARD19'),
         )
         pin_factory = kwargs.pop('pin_factory', None)
         if len(labels) == 0:
@@ -1306,11 +1311,11 @@ class StatusBoard(CompositeOutputDevice):
 
     def __init__(self, *labels, **kwargs):
         pins = (
-            (17, 4, 14),
-            (22, 27, 19),
-            (9, 10, 15),
-            (5, 11, 26),
-            (13, 6, 18),
+            ('BOARD11', 'BOARD7', 'BOARD8'),
+            ('BOARD15', 'BOARD13', 'BOARD35'),
+            ('BOARD21', 'BOARD19', 'BOARD10'),
+            ('BOARD29', 'BOARD23', 'BOARD37'),
+            ('BOARD33', 'BOARD31', 'BOARD12'),
         )
         pin_factory = kwargs.pop('pin_factory', None)
         if len(labels) == 0:
@@ -1398,12 +1403,12 @@ class SnowPi(LEDBoard):
         super(SnowPi, self).__init__(
             arms=LEDBoard(
                 left=LEDBoard(
-                    top=17, middle=18, bottom=22,
+                    top='BOARD11', middle='BOARD12', bottom='BOARD15',
                     pwm=pwm, initial_value=initial_value,
                     _order=('top', 'middle', 'bottom'),
                     pin_factory=pin_factory),
                 right=LEDBoard(
-                    top=7, middle=8, bottom=9,
+                    top='BOARD26', middle='BOARD24', bottom='BOARD21',
                     pwm=pwm, initial_value=initial_value,
                     _order=('top', 'middle', 'bottom'),
                     pin_factory=pin_factory),
@@ -1411,12 +1416,12 @@ class SnowPi(LEDBoard):
                 pin_factory=pin_factory
                 ),
             eyes=LEDBoard(
-                left=23, right=24,
+                left='BOARD16', right='BOARD18',
                 pwm=pwm, initial_value=initial_value,
                 _order=('left', 'right'),
                 pin_factory=pin_factory
                 ),
-            nose=25,
+            nose='BOARD22',
             pwm=pwm, initial_value=initial_value,
             _order=('eyes', 'nose', 'arms'),
             pin_factory=pin_factory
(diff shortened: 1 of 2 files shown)


martynwheeler:
Hi,

Thank you for this. How do I use this branch?

ukBaz:
Hi @martynwheeler 

Are you asking for the git commands to get this branch? e.g:
https://stackoverflow.com/questions/1783405/how-do-i-check-out-a-remote-git-branch
`git checkokut 349-board-numbering-addons`

Or are you seeking clarification about the different pin Pin Numbering schemes? e.g:
https://gpiozero.readthedocs.io/en/stable/recipes.html#pin-numbering

martynwheeler:
Hi,
Sorry for not being clear.  It was the former that I was looking for.  Is it likely that these changes will be merged into the master?

Thanks, Martyn

martynwheeler:
Sorry for the questions,
Do I have to clone the git repository and then install the branch?  I am not very au fait with using git and am slightly confused??  Or is it easier to wait until this branch is merged?

lurch:
There's quite good docs at https://gpiozero.readthedocs.io/en/stable/development.html#development-installation :wink:

martynwheeler:
Thanks,
Not really sure that I understand the process enough, I don't want to develop and not sure how to use a virtual environment.  I think I will just have to wait until it is merged and just use something else in the meantime.

martynwheeler:
Hmm, I tried the first couple of things and then got this:

$ git clone https://github.com/RPi-Distro/python-gpiozero.git
Cloning into 'python-gpiozero'...
remote: Enumerating objects: 111, done.
remote: Counting objects: 100% (111/111), done.
remote: Compressing objects: 100% (77/77), done.
error: RPC failed; curl 56 GnuTLS recv error (-9): Error decoding the received TLS packet.
fatal: the remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed

bennuttall:
That looks like your internet connection.

I've made a release you can try:

```
sudo pip3 install https://bennuttall.com/files/gpiozero-1.5.2b349-py3-none-any.whl
```

martynwheeler:
Thank you, I have to go and teach a lesson now.  Will try later.

martynwheeler:
@bennuttall - thanks for the pointer (Virgin Media's internet safety feature was blocking git).  Okay, I have followed the instructions and installed this branch in a virtual environment and it works perfectly!  Thank you.  Will this branch be merged into the master in the near future or should I install this branch on my system?

Once again, @lurch and @ukBaz thanks for guiding me through the process.

bennuttall:
Good to know it works. @lurch is this what you had in mind? I'll try and check with Dave if he has a mo.

Seems like a reasonable approach to take for all physical boards really.

lurch:
> @lurch is this what you had in mind?

I can't quite remember, given that I originally opened that issue 4 years ago :wink:  But yeah the change here looks reasonable.
Thanks to @martynwheeler for testing and confirming that it works on your old Pi.

> Seems like a reasonable approach to take for all physical boards really.

Yeah IMHO if you're gonna be doing this it makes sense to do it for **all** add-on boards (for consistency).

bennuttall:
> Yeah IMHO if you're gonna be doing this it makes sense to do it for all add-on boards (for consistency).

Well, that was a dull way to spend the last hour. Done!

One thing that annoys me is that the way the input/output pins are made available with the new `Pibrella` class:

```python
In [1]: from gpiozero import *                                                                           

In [2]: pb = Pibrella()                                                                                  

In [3]: pb.inputs.a                                                                                      
Out[3]: 'BOARD21'
```

But I guess they still work as intended:

```python
In [4]: btn = Button(pb.inputs.a, pull_up=False) 
   ...: led = LED(pb.outputs.e)
```

So it's only really if somebody prints them out that it could confuse people... bit of an edge case though!

One question: what does `LED('BOARD2')` or god forbid `Energenie()` do on a compute module? Does it matter? :laughing:

gpiozero/boards.py
```diff
@@ -592,7 +592,7 @@ def pulse(self, fade_in_time=1, fade_out_time=1, n=None, background=True):
             on_time, off_time, fade_in_time, fade_out_time, n, background)
 
     def _blink_device(
-            self, on_time, off_time, fade_in_time, fade_out_time, n, fps=25):
```
bennuttall, line 595: Oh jesus... find an replace

gpiozero/boards.py
```diff
@@ -1814,11 +1826,12 @@ class CamJamKitRobot(Robot):
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
 
-    .. _CamJam #3 EduKit: http://camjam.me/?page_id=1035
+    .. _CamJam #3 EduKit: http://camjam.me/?page_id='BOARD19'35
```
lurch: Whoops

lurch:
I wonder if it might be worth leaving the tests using the GPIO numbers, as that provides a vague sanity-check that the translations between BOARD numbers and GPIO numbers is working correctly?

> One thing that annoys me is that the way the input/output pins are made available with the new Pibrella class

Hmmm, that's a shame - I kinda assumed that specifying a BOARD number when constructing an object would automatically convert it back into the internal GPIO numbers. But thinking about it, perhaps constructing e.g. a `LED` with a BOARD number and having it print as a GPIO number, would be even more confusing?

Oh... now having looked at the code in more detail :mag: I see that it's because pb.inputs is just a namedtuple, and a GpioZero object hasn't been created yet. But I think if you printed out btn.pin in your example then you'd see the GPIO number?

> One question: what does LED('BOARD2') or god forbid Energenie() do on a compute module?

Looking at https://github.com/gpiozero/gpiozero/blob/master/gpiozero/pins/data.py#L1272 and https://github.com/gpiozero/gpiozero/blob/master/gpiozero/pins/data.py#L320 I _think_ that 'BOARD2' on a compute module would get converted to 'EMMC DISABLE N' and then "raise PinInvalidPin('%s is not a valid pin spec' % spec)"

> Does it matter? :laughing: 

Probably not ;-)

bennuttall:
Eugh it seems silly to worry about compute module but if you think about a non-industrial hacker type playing around with one - what would they expect? Wire up e.g. en Energenie using the right BCM pins, they would expect it to work.

Let's say we make it possible for that to work - some workaround that we document in the FAQs - what would that look like?

My initial thought would be allowing an override to pins in the constructor:

```python
class SomeHat(LEDBoard):
    def __init__(self, pins=None, pwm=False, pin_factory=None):
        if pins is None:
            pins = (2, 3, 4, 5)
        elif len(pins) != 4:
            raise GPIOPinErrorWhatever()
        super().__init__(*pins, pwm=pwm, pin_factory=pin_factory)
```

Instead of what this class would look like now:

```python
class SomeHat(LEDBoard):
    def __init__(self, pwm=False, pin_factory=None):
        pins = (2, 3, 4, 5)
        super().__init__(*pins, pwm=pwm, pin_factory=pin_factory)
```

An entirely different approach to all of this would be: instead of using BOARD numbers in classes, we call a function that converts physical pin numbers to BCM numbers for the original Pi, but leaves them alone for everything else, e.g:

```python
def pin(n):
    # where n is the board number
    rev = pi_info().revision
    if rev == ORIGINAL_PI_REV:
        return ORIGINAL_PI_PINS[n]
    return n
```

This is ridiculous 😆

lurch:
`def __init__(self, pins=None, pwm=False, pin_factory=None):`
Provided that the `pins` parameter is moved to be last in the list (for backwards compatibility) I think an approach like that might work. This would also be beneficial for people using things like https://shop.pimoroni.com/products/mini-black-hat-hack3r to connect multiple add-on boards to their Pi at once :smiley: 

Although perhaps if a `pins` parameter is given it should be a `dict` instead of a `tuple` ?

bennuttall:
It would be better all-round if pwm was keyword-only :)

Good point re: BHH

In a way, adding `pins` could be a separate solution to this. I think it's worth considering regardless. Gets a bit tricky with the "shape" of the value...

What do you think about the other approach? We could do both 😆

lurch:
> It would be better all-round if pwm was keyword-only :)

I think I once remember seeing @waveform80 say that Python3 allows keyword-only arguments (?), but that we couldn't use this yet because GpioZero supports py2 and py3. But if we're going to be dropping py2...

> What do you think about the other approach? We could do both laughing

Given how complex this is getting, I think we ought to stick to just one way of doing this :wink:

> Gets a bit tricky with the "shape" of the value...

Do you mean checking that the required numbers of `pins` are provided? IMHO it makes sense for this to be an all-or-nothing thing - either the `pins` is left at `None` (the default) or _all_ pins must be provided, with anything else being an error.
We could have e.g. `hat = TrafficHat(pins={'red': 2, 'amber': 3, 'green': 4, 'button': 5, 'buzzer': 6})`. And extending this back to the [ComputeModule](https://github.com/gpiozero/gpiozero/blob/master/gpiozero/pins/data.py#L320) question I think that'd even work with `hat = TrafficHat(pins={'red': 'BOARD9', 'amber': 'BOARD11', 'green': 'BOARD15', 'button': 'BOARD17', 'buzzer': 'BOARD21'})` :laughing:

bennuttall:
> I think I once remember seeing @waveform80 say that Python3 allows keyword-only arguments (?), but that we couldn't use this yet because GpioZero supports py2 and py3. But if we're going to be dropping py2...

Yeah, it would be good to enforce keyword-only arguments (where appropriate) post- dropping Python 2.

> Given how complex this is getting, I think we ought to stick to just one way of doing this

I see this as two separate things:
- specifying pins in board definitions so that they work on original pis too
- allowing people to override pins for boards (for any reason)

> Do you mean checking that the required numbers of pins are provided?

I meant where the board is complex enough to have a nested "shape" (i.e. `.value` is not just a flat tuple or dict) like the JamHat whose value shape is:

```
JamHatValue(
    lights_1=LEDBoardValue(red=0, yellow=0, green=0),
    lights_2=LEDBoardValue(red=0, yellow=0, green=0),
    button_1=0,
    button_2=0,
    buzzer=None
)
```

lurch:
> I see this as two separate things:

Yeah, I meant that I don't think there's much value in investigating your "instead of using BOARD numbers in classes, we call a function that converts physical pin numbers to BCM numbers for the original Pi, but leaves them alone for everything else" suggestion - IMHO it feels far too special-case and error-prone. (And isn't "a function that converts physical pin numbers to BCM numbers" exactly what BOARD numbering does anyway? :wink: )

> I meant where the board is complex enough to have a nested "shape" (i.e. .value is not just a flat tuple or dict) like the JamHat

Ah, so you're asking whether a board with a "shaped" `.value`, should also have a "shaped" `pins` parameter in its constructor? I personally don't think there's any value in that - given that if a custom `pins` parameter (if used) will only ever be a one-off thing, IMHO it makes sense to keep `pins` as a flat dict? Remember that all it's telling us is how the user has re-wired the connections to their add-on board (using either the ComputeModule or something like a BlackHatHacker).
So I'd suggest _something_ like:
```
hat = JamHat(pins={red1=2, yellow1=3, green1=4, red2=5, yellow2=6, green2=7, button1=8, button2=9, buzzer=10})
```
(with the JamHat constructor than checking that `pins is None or pins.keys() == set(('red1', 'yellow1', 'green1', 'red2', 'yellow2', 'green2', 'button1', 'button2', 'buzzer'))` )

But perhaps it's worth keeping this PR as just "specifying pins in board definitions so that they work on original pis too" and then creating a separate issue / PR for "allowing people to override pins for boards (for any reason)" ?

Hmmmm2, perhaps we could allow a "special value" to be supplied to the `pins` init-parameter to allow a user to indicate that although they're using a ComputeModule, they've actually wired up their GPIO numbers the same as on a regular Pi B+ ? (which I guess is probably the case for a number of ComputeModule "host boards"?) `hat = JamHat(pins='USE_BPLUS_MAPPING')` ?

P.S. Slightly off-topic question, but is there a reason why `JamHat` uses two `LEDBoard` instances rather than two `TrafficLights` instances?

bennuttall:
> I personally don't think there's any value in that - given that if a custom pins parameter (if used) will only ever be a one-off thing

True. We could use the dict declaration ourselves, so instead of:

```
super.__init__(lights=TrafficLights(red=2, amber=3, green=4), button=5)
```

we do:

```
if pins is None:
    pins = {red: 2, amber=3, green=4, button=5}
super.__init__(lights=TrafficLights(red=pins['red'], amber=pins['amber'], green=pins['green']), button=pins['button'])
```

That way we pave the way for copy/paste/edit.

> Hmmmm2, perhaps we could allow a "special value" to be supplied to the pins init-parameter to allow a user to indicate that although they're using a ComputeModule

That makes sense :)

> P.S. Slightly off-topic question, but is there a reason why JamHat uses two LEDBoard instances rather than two TrafficLights instances?

I don't know! I'd have to check the PR. Probably a mistake by overthinking it. I know what you're thinking, that sounds really out of character, but I guess sometimes these things happen.

lurch:
```
if pins is None:
    pins = {'red': 2, 'amber': 3, 'green': 4, 'button': 5}
super.__init__(lights=TrafficLights(red=pins['red'], amber=pins['amber'], green=pins['green']), button=pins['button'])
```

Nice! Except we'd obviously be using 'BOARD' numbers rather than GPIO numbers in the second line :wink:

martynwheeler:
Hi,
Just wondering when this will be merged into the main branch?  Thank you,  Martyn

bennuttall:
Not sure yet - think it needs a rethink. And we're in the middle of another project at the moment.

martynwheeler:
No problem, thanks for the update. Sent from Mail for Windows 10 From: Ben NuttallSent: 24 June 2020 13:29To: gpiozero/gpiozeroCc: martynwheeler; MentionSubject: Re: [gpiozero/gpiozero] Update add-on boards to use BOARD numbering, close #349 [WIP] (#860) Not sure yet - think it needs a rethink. And we're in the middle of another project at the moment.—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

gpiozero/boards.py
```diff
@@ -961,9 +964,10 @@ class PiLiterBarGraph(LEDBarGraph):
     """
 
     def __init__(self, pwm=False, initial_value=0.0, pin_factory=None):
-        pins = (4, 17, 27, 18, 22, 23, 24, 25)
         super(PiLiterBarGraph, self).__init__(
-            *pins, pwm=pwm, initial_value=initial_value, pin_factory=pin_factory
+            'BOARD7', 'BOARD11', 'BOARD13', 'BOARD12',
```
waveform80, line 968: These could just go in pins?

waveform80:
Okay, read through the thread: yes, this will break on compute modules (1, 3, and 3+). However, I'm not fussed. Given CM4 is now out and (due to the various needs of the high-speed interfaces) has reverted to a basically "classic" GPIO layout (at least on the dev board), and that CM1/3/3+ users will be a *tiny* niche of gpiozero users, I'm fine with merging this as is. Any objections @bennuttall ?

Issue #707, opened 2019-01-15
Allow users to disable implicit cleanup


bennuttall:
Sometimes, you want to write a script which turns on a device, and leaves it on. Let's say you have a cron job to check something periodically, and control a device accordingly. Rather than have the script continuously running, you want to run it regularly. With gpiozero, your script will turn the device on, exit and turn it back off when running implicit cleanup.

I saw on this [SO question](https://stackoverflow.com/questions/53618198/using-gpiozero-on-raspberry-pi-to-control-pins-but-output-pins-are-reset-up) recently that someone wanted to do this, and they ended up using RPi.GPIO instead.

It should be possible in gpiozero. If we could either have it per-device:

```python
led = LED(2, cleanup=False)
```

or just in the `Device` base class:

```python
Device.cleanup = False

led = LED(2)
```

or provide both?

I'm not certain that's the right API, but you get the idea.

This was originally raised in #323. In fact I almost suggested this at the end (though my suggestion doesn't quite make sense...)

lurch:
My gut feeling is that `cleanup=False` is potentially dangerous, so it should definitely be a per-device setting rather than a global `Device` setting?

bennuttall:
Per-device makes sense.

lurch:
Having just read through #323 again, I wonder if we need _both_ `cleanup` and `final_value` properties?
(with `final_value` being needed for those boards like Energenie where we can't read the device's value)

For an LED, if it's on before we run our gpiozero script and we specify `initial_value=None, cleanup=False`, should the final state of the LED be on or off after the script exits? And similarly if it's off before we run our gpiozero script and we specify `initial_value=None, cleanup=False`, should the final state of the LED be on or off after the script exits? And if we specify an explicit `initial_value`, how should that affect the final state of the LED with `cleanup=False`?
Or am I overcomplicating things, and `cleanup` will simply leave the pin at it's current state? Might users want a `return_to_initial_state` setting? :man_shrugging:   ( @lurch runs away from the can of worms he just opened :running_man: :wink: )

CAM-Gerlach:
@lurch @bennuttall Its been exactly a year since the last comment on this. Is there any realistic plans for its implementation? To me, this seems like a very basic feature, given how seemingly common scripts that toggle or turn on/off devices (e.g. LEDs) are _especially_ among beginners, RPI.GPIO's primary target audience.

[...]

PR #865, opened 2020-05-31
Energenie sockets preserve state

NotBobTheBuilder:
At present when you initialise the `Energenie` class, sockets are either switched on or off according to `initial_value`.

Sometimes it is useful to initialise the class without setting a state, for example starting a service to control a socket which respects its initial state.

This PR allows specifying `None` as an `initial_value`, in which case no message is sent to the socket. 

The default behaviour of switching off the socket remains unchanged

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -58,7 +58,6 @@ from .exc import (
     GPIOPinMissing,
     EnergenieSocketMissing,
     EnergenieBadSocket,
-    EnergenieBadInitialValue,
     OutputDeviceBadValue,
     CompositeDeviceBadDevice,
     )
@@ -2054,7 +2053,8 @@ class Energenie(SourceMixin, Device):
         means of reading their state, you may provide an initial state for
         the socket, which will be set upon construction. This defaults to
         :data:`False` which will switch the socket off.
-        Specifying :data:`None` will not switch the socket on construction.
+        Specifying :data:`None` will not set any initial state nor transmit any
+        control signal to the device.
 
     :type pin_factory: Factory or None
     :param pin_factory:
@@ -2068,15 +2068,13 @@ class Energenie(SourceMixin, Device):
             raise EnergenieSocketMissing('socket number must be provided')
         if not (1 <= socket <= 4):
             raise EnergenieBadSocket('socket number must be between 1 and 4')
-        if initial_value is None:
-            raise EnergenieBadInitialValue("initial value can't be None")
         self._value = None
         super(Energenie, self).__init__(pin_factory=pin_factory)
         self._socket = socket
         self._master = _EnergenieMaster(pin_factory=pin_factory)
         if initial_value:
             self.on()
-        elif initial_value == False:
+        elif initial_value is not None:
             self.off()
 
     def close(self):
@@ -2107,11 +2105,16 @@ class Energenie(SourceMixin, Device):
         """
         Returns :data:`True` if the socket is on and :data:`False` if the
         socket is off.  Setting this property changes the state of the socket.
+        Returns :data:`None` only when constructed with :data:`initial_value`
+        set to :data:`None` and neither :data:`on()` nor :data:`off()` have
+        been called since construction.
         """
         return self._value
 
     @value.setter
     def value(self, value):
+        if value is None:
+            raise TypeError('value cannot be None')
         value = bool(value)
         self._master.transmit(self._socket, value)
         self._value = value
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -2048,11 +2048,13 @@ class Energenie(SourceMixin, Device):
         Which socket this instance should control. This is an integer number
         between 1 and 4.
 
-    :param bool initial_value:
+    :type initial_value: bool or None
+    :param initial_value:
         The initial state of the socket. As Energenie sockets provide no
-        means of reading their state, you must provide an initial state for
+        means of reading their state, you may provide an initial state for
         the socket, which will be set upon construction. This defaults to
         :data:`False` which will switch the socket off.
+        Specifying :data:`None` will not switch the socket on construction.
 
     :type pin_factory: Factory or None
     :param pin_factory:
@@ -2074,7 +2076,7 @@ class Energenie(SourceMixin, Device):
         self._master = _EnergenieMaster(pin_factory=pin_factory)
         if initial_value:
             self.on()
-        else:
+        elif initial_value == False:
             self.off()
 
     def close(self):
--- a/tests/test_boards.py
+++ b/tests/test_boards.py
@@ -1264,19 +1264,21 @@ def test_energenie_bad_init(mock_factory):
         Energenie(0)
     with pytest.raises(ValueError):
         Energenie(5)
-    with pytest.raises(ValueError):
-        Energenie(1, initial_value=None)
 
 def test_energenie(mock_factory):
     pins = [mock_factory.pin(n) for n in (17, 22, 23, 27, 24, 25)]
     with Energenie(1, initial_value=True) as device1, \
-         Energenie(2, initial_value=False) as device2:
+         Energenie(2, initial_value=False) as device2, \
+         Energenie(3, initial_value=None) as device3:
         assert repr(device1) == '<gpiozero.Energenie object on socket 1>'
         assert repr(device2) == '<gpiozero.Energenie object on socket 2>'
+        assert repr(device3) == '<gpiozero.Energenie object on socket 3>'
         assert device1.value
         assert not device2.value
+        assert device3.value is None
         assert device1.socket == 1
         assert device2.socket == 2
+        assert device3.socket == 3
         [pin.clear_states() for pin in pins]
         device1.on()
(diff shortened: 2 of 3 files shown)


bennuttall:
Thanks Jack. I seem to remember @waveform80 intentionally not allowing this, because you can't read the initial state, so if you tried to read the value after init, before explicitly turning it on or off, it might "lie" to you.

A compromise may be to allow `initial_value` to be `None`, but then let `value` be `None` until it's been set by turning the device on or off.

I'll try and get a decision from Dave.

bennuttall:
Sounds good to me. Push another commit and I'll try it out.

Will defer to Dave to make the final call, since we wrote it in the first place. But this seems a reasonable way of dealing with it.

bennuttall:
Maybe look at adding tests for this edge case if you can.

lurch:
It's actually the "other side of the coin" but you may find the discussions in #707 and #323 of interest too :wink: 

I wondered if having `value == None` might be a weird thing for the GpioZero API to support, but then I remembered that we use it in servo https://gpiozero.readthedocs.io/en/stable/api_output.html#gpiozero.Servo.value
"This can also be the special value None indicating that the servo is currently “uncontrolled”, i.e. that no control signal is being sent."

So I guess it might make sense that for Energenie "This can also be the special value None indicating that the state of the socket is unknown"? And obviously while you could read a `None` value from an Energenie, you should be prevented from setting `.value` to None.

gpiozero/boards.py
```diff
@@ -2045,11 +2045,13 @@ class Energenie(SourceMixin, Device):
...
 
-    :param bool initial_value:
+    :type initial_value: bool or None
+    :param initial_value:
         The initial state of the socket. As Energenie sockets provide no
-        means of reading their state, you must provide an initial state for
+        means of reading their state, you may provide an initial state for
         the socket, which will be set upon construction. This defaults to
         :data:`False` which will switch the socket off.
+        Specifying :data:`None` will not switch the socket on construction.
```
lurch, line 2054: "will not alter the socket's state" might be less ambiguous than "will not switch the socket on construction" ?

("not switch the socket on" could be misinterpreted as "switch the socket off" ? :wink: )

gpiozero/boards.py
```diff
@@ -2063,15 +2065,15 @@ def __init__(self, socket=None, initial_value=False, pin_factory=None):
...
+        if initial_value not in (True, False, None):
+            raise EnergenieBadInitialValue("initial value must be bool or None")
         self._value = None
         super(Energenie, self).__init__(pin_factory=pin_factory)
         self._socket = socket
         self._master = _EnergenieMaster(pin_factory=pin_factory)
         if initial_value:
             self.on()
-        else:
+        elif initial_value is False:
```
lurch, line 2076: I don't think we should be checking explicitly for True or False, but instead simply checking for None, and then just using duck-typing otherwise? See e.g. https://github.com/gpiozero/gpiozero/blob/master/gpiozero/input_devices.py#L112

I think this also means the `if initial_value not in (True, False, None):` line would need to be removed?

gpiozero/boards.py
```diff
@@ -2102,6 +2105,9 @@ def value(self):
         """
         Returns :data:`True` if the socket is on and :data:`False` if the
         socket is off.  Setting this property changes the state of the socket.
+        Returns :data:`None` only when constructedwith :data:`initial_value`
```
lurch, line 2108: Missing space in "constructedwith" :wink:

gpiozero/boards.py
```diff
@@ -2102,6 +2105,9 @@ def value(self):
...
+        set to :data:`None` and neither :data:`on()` nor :data:`off()` have
+        been called since construction
```
lurch, line 2110: Missing full-stop at end of sentence.

gpiozero/boards.py
```diff
@@ -2063,13 +2066,13 @@ def __init__(self, socket=None, initial_value=False, pin_factory=None):
...
             raise EnergenieBadSocket('socket number must be between 1 and 4')
-        if initial_value is None:
-            raise EnergenieBadInitialValue("initial value can't be None")
         self._value = None
         super(Energenie, self).__init__(pin_factory=pin_factory)
         self._socket = socket
         self._master = _EnergenieMaster(pin_factory=pin_factory)
-        if initial_value:
+        if initial_value is None:
+            pass
```
lurch, line 2074: Ugh :slightly_frowning_face: 
Could you please make this:
```python3
if initial_value is not None:
    if initial_value:
        self.on()
    else:
        self.off()
```

tests/test_boards.py
```diff
@@ -1295,6 +1297,15 @@ def test_energenie(mock_factory):
...
         pins[5].assert_states_and_times([(0.0, False), (0.1, True), (0.25, False)])
+        [pin.clear_states() for pin in pins]
+        device3.on()
+        assert device3.value
+        pins[0].assert_states_and_times([(0.0, False), (0.0, True)])
+        pins[1].assert_states_and_times([(0.0, True), (0.0, False)])
+        pins[2].assert_states_and_times([(0.0, True), (0.0, False)])
+        pins[3].assert_states_and_times([(0.0, True), (0.0, False)])
+        pins[4].assert_states_and_times([(0.0, False)])
+        pins[5].assert_states_and_times([(0.0, False), (0.1, True), (0.25, False)])
```
lurch, line 1308: Would you mind adding tests that `device3.value = 0` and `device3.value = 1` work as expected, and `device3.value = None` raises an exception? Thanks!

gpiozero/boards.py
```diff
@@ -2063,13 +2066,13 @@ def __init__(self, socket=None, initial_value=False, pin_factory=None):
             raise EnergenieSocketMissing('socket number must be provided')
         if not (1 <= socket <= 4):
             raise EnergenieBadSocket('socket number must be between 1 and 4')
-        if initial_value is None:
-            raise EnergenieBadInitialValue("initial value can't be None")
```
lurch, line 2067: Just for the sake of tidiness, perhaps it's worth removing this exception from the library? :shrug:

lurch: ...and also from `exc.py` ? (sorry for not being more explicit)

NotBobTheBuilder: Should be removed :) (maybe you are seeing a partial diff?)

gpiozero/boards.py
```diff
@@ -2102,11 +2102,16 @@ def value(self):
...
+        Returns :data:`None` only when constructed with :data:`initial_value`
+        set to :data:`None` and neither :data:`on()` nor :data:`off()` have
+        been called since construction.
         """
         return self._value
 
     @value.setter
     def value(self, value):
+        if value is None:
+            raise TypeError('value cannot be None')
```
lurch, line 2114: A more specific exception would be better - hmm perhaps `EnergenieBadInitialValue` in `exc.py` could be renamed to `EnergenieBadValue` (with the detail-string updated too) and then used here? :slightly_smiling_face:

tests/test_boards.py
```diff
@@ -1295,6 +1297,25 @@ def test_energenie(mock_factory):
...
+
+        device3.value = False
+        device3.value = True
+        try:
+            device3.value = None
+        except:
```
lurch, line 1314: Should probably use the `with pytest.raises(ValueError)` syntax rather than using try / except?

tests/test_boards.py
```diff
@@ -1295,6 +1297,25 @@ def test_energenie(mock_factory):
...
+        assert device3.value
+        pins[0].assert_states_and_times([(0.0, False), (0.0, True)])
+        pins[1].assert_states_and_times([(0.0, True), (0.0, False)])
+        pins[2].assert_states_and_times([(0.0, True), (0.0, False)])
+        pins[3].assert_states_and_times([(0.0, True), (0.0, False)])
+        pins[4].assert_states_and_times([(0.0, False)])
+        pins[5].assert_states_and_times([(0.0, False), (0.1, True), (0.25, False)])
+
+        device3.value = False
+        device3.value = True
```
lurch, line 1311: I'd probably make this:
```python3
        device3.value = False
        assert not device3.value
        device3.value = True
        assert device3.value
```
(i.e. check that the value you set is also the value you read back!)

gpiozero/boards.py
```diff
@@ -2063,15 +2065,13 @@ def __init__(self, socket=None, initial_value=False, pin_factory=None):
...
-        if initial_value is None:
-            raise EnergenieBadInitialValue("initial value can't be None")
         self._value = None
         super(Energenie, self).__init__(pin_factory=pin_factory)
         self._socket = socket
         self._master = _EnergenieMaster(pin_factory=pin_factory)
         if initial_value:
             self.on()
-        else:
+        elif initial_value is not None:
```
lurch, line 2074: Sorry to be such a nitpicker, but although what you're proposing here is "more compact", I still prefer the "more explicit" version that I suggested:
```python3
if initial_value is not None:
    if initial_value:
        self.on()
    else:
        self.off()
```

I guess it's up to @bennuttall really though!

bennuttall:
Dave approves!

PR #868, opened 2020-06-13
Add missing Button import to led_button_remote_*.py examples

linusg:
`docs/examples/led_button_remote_1.py` and `docs/examples/led_button_remote_2.py` use `gpiozero.Button` but don't import it.


bennuttall:
Thanks!

PR #875, opened 2020-07-21
Make sure threads are brought down in all cases

jsiverskog:
If shutting down all threads in a Python program by doing something like:
```
  idx = 0
  for thread in threading.enumerate():
      if idx != 0:
          thread.join(timeout=0.25)
      idx += 1
```
gpiozero will always throw the "Thread failed to die within X seconds" error.

Set the stop flag in join (if unset) to avoid this from happening.

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -264,7 +264,7 @@ class DigitalOutputDevice(OutputDevice):
         )
         self._blink_thread.start()
         if not background:
-            self._blink_thread.join()
+            self._blink_thread.wait()
             self._blink_thread = None
 
     def _stop_blink(self):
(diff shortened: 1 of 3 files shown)


jsiverskog:
what do i have to do to get this merged?

waveform80:
Hmm, I'm not really convinced about this: the semantics of join don't usually include termination, just waiting for termination. But mostly my unease is that attempting to join all threads in an application indiscriminately (rather than just those threads that your code explicitly created), just seems like a bad idea. There's no guarantee that shutting down gpiozero's threads at that point is safe, not unless you've explicitly shut down all the pin factories first (and, if you do that, why not add a branch in the code to handle calling "stop" on any GPIOThread).

I *could* see an argument that gpiozero ought to have a publicly available termination function that could be called to perform all the cleanup it usually performs via `atexit` (so that users could call it at any time instead of having to wait for interpreter shutdown)?

PR #892, opened 2020-10-01
Memory pinout GB

chrisruk:
Uses GB for memory values >= 1GB and MB for values < 1GB

### DIFF
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -1305,11 +1305,15 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
             return self.board.format(style=style, **kw)
         elif content == 'specs':
             specs = self._asdict()
-            specs['memory'] /= 1024
+            if specs['memory'] < 1024:
+                specs['memory_unit'] = "MB"
+            else:
+                specs['memory'] /= 1024
+                specs['memory_unit'] = "GB"
             return dedent("""\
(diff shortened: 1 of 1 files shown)


bennuttall:
Looks good! ping @waveform80

gpiozero/pins/data.py
```diff
@@ -1304,18 +1304,24 @@ def __format__(self, format_spec):
                 })
             return self.board.format(style=style, **kw)
         elif content == 'specs':
+            specs = self._asdict()
+            if specs['memory'] < 1024:
+                specs['memory_unit'] = "MB"
+            else:
+                specs['memory'] /= 1024
```
lurch, line 1311: A bit late now, but I wonder if it'd be worth coercing this back to an integer, rather than leaving it as floating-point?

waveform80: I'm fairly relaxed about leaving it as a floating point (on the off-chance there's ever something with 1.5Gb of RAM), though ideally the output format template should limit the resulting value to 1 decimal place. Ideally, I'd throw my usual [format_size](https://github.com/piwheels/piwheels/blob/master/piwheels/format.py#L42) routine at this but I'm told the fact it uses logs makes it difficult to understand, and it's more general-purpose than we really need here :).

There's a pile of work I need to do on this module anyway to "properly" support the Pi4 (ASCII-art and what-not!) so I'll try and remember to have another look at this before the next release.

Issue #723, opened 2019-01-29
pinout: rotate option


bennuttall:
Let's add an option to rotate `pinout` by 180 degrees.

AthithyaIthayan:
@bennuttall I would Like to work over this issue can u just specify the pre existing ```reverse pinout``` command so that I can just change it. Thank You!

PR #894, opened 2020-10-02
Rotate option - to rotate pinout 180 degrees

chrisruk:
I would be interested in improving this, by not using a global variable, the difficulty I found being that you can't add additional parameters to __format__ calls as far as I understand.

Example pinout:

```
J8:
   3V3  (1) (2)  5V    
 GPIO2  (3) (4)  5V    
 GPIO3  (5) (6)  GND   
 GPIO4  (7) (8)  GPIO14
   GND  (9) (10) GPIO15
GPIO17 (11) (12) GPIO18
GPIO27 (13) (14) GND   
GPIO22 (15) (16) GPIO23
   3V3 (17) (18) GPIO24
GPIO10 (19) (20) GND   
 GPIO9 (21) (22) GPIO25
GPIO11 (23) (24) GPIO8 
   GND (25) (26) GPIO7 
 GPIO0 (27) (28) GPIO1 
 GPIO5 (29) (30) GND   
 GPIO6 (31) (32) GPIO12
GPIO13 (33) (34) GND   
GPIO19 (35) (36) GPIO16
GPIO26 (37) (38) GPIO20
   GND (39) (40) GPIO21
```

Rotated pinout:
```
J8:
(40) GPIO21    GND (39)
(38) GPIO20 GPIO26 (37)
(36) GPIO16 GPIO19 (35)
(34) GND    GPIO13 (33)
(32) GPIO12  GPIO6 (31)
(30) GND     GPIO5 (29)
(28) GPIO1   GPIO0 (27)
(26) GPIO7     GND (25)
(24) GPIO8  GPIO11 (23)
(22) GPIO25  GPIO9 (21)
(20) GND    GPIO10 (19)
(18) GPIO24    3V3 (17)
(16) GPIO23 GPIO22 (15)
(14) GND    GPIO27 (13)
(12) GPIO18 GPIO17 (11)
(10) GPIO15    GND  (9)
(8)  GPIO14  GPIO4  (7)
(6)  GND     GPIO3  (5)
(4)  5V      GPIO2  (3)
(2)  5V        3V3  (1)
```

### DIFF
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -664,10 +664,13 @@ class HeaderInfo(namedtuple('HeaderInfo', (
     def _format_full(self, style):
         Cell = namedtuple('Cell', ('content', 'align', 'style'))
         lines = []
-        for row in sorted(range(self.rows), reverse=rotate):
+        for row in range(self.rows):
             line = []
-            for col in sorted(range(self.columns), reverse=rotate):
-                pin = (row * self.columns) + col + 1
+            for col in range(self.columns):
+                if rotate:
+                    pin = (((self.rows - 1) - row) * self.columns) + ((self.columns - 1) - col) + 1
+                else:
+                    pin = (row * self.columns) + col + 1
                 try:
                     pin = self.pins[pin]
                     cells = [
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -47,6 +47,9 @@ from collections import namedtuple
 from ..exc import PinUnknownPi, PinMultiplePins, PinNoPins, PinInvalidPin
 from ..devices import Device
 
+# rotate pinout
+
+rotate = False
 
 # Some useful constants for describing pins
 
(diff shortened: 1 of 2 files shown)


bennuttall:
Sounds good. It was suggested in #723 and someone started working on it at a pycon sprint - it was basically finished but they never submitted a PR :(

Looking forward to it

lurch:
Wouldn't the rotated pinout look better with the pin-numbers down the middle rather than down the outside?

lurch:
Looking at the code, you might be able to avoid using a global variable by using the Style / format_spec stuff? But unfortunately that's all semi-magic @waveform80 python code... :confused:  (i.e. hard to decipher :wink: )

chrisruk:
Yep good point about the numbering being in the middle, I've just fixed that. 

Will have a look more at the style / format_spec

waveform80:
This is a nice idea, but the wrong  way of going about it; it is possible to customize format-specs -- in fact that's exactly what the current code does -- and that's probably what we ought to do here (add some parameters to the format spec to optionally specify rotation of the output).

However, I'm not convinced my current custom format-spec code is terribly good. The custom format-spec isn't well thought out and, while I've been playing with it to add some more utilities I've come to the conclusion it's one of those areas that needs a serious re-write in 2.x. It's also an opportunity to expand PinInfo to include details of the alternative functions of pins (so users can actually query "which are the hardware SPI pins" for example).

Anyway, I'm going to close this PR, but only because it's the wrong approach: the idea is good and is something I'll certainly incorporate into the re-write.

PR #896, opened 2020-10-12
Drop Python 2.7, 3.2, 3.3, 3.4 #799

### LINKED ISSUE #799: Python version deprecation policy
I think we should drop support for Python versions when they become EOL.

We currently test on Python 2.7, 3.2, 3.3, 3.4, 3.5, 3.6 and 3.7. Two of these (3.2 and 3.3) are EOL, and another (2.7) is approaching EOL:

Python version | Debian version | EOL | Debian version EOL
--- | --- | --- | ---
2.7 | Buster | 2020-01-01 | 2020-06-30
3.2 | Wheezy | 2016-02-20 | 2018-05-31
3.3 | None | 2017-09-29 | N/A
3.4 | Jessie | 2019-03-18 | 2020-06-30
3.5 | Stretch | 2020-09-13 | 2022-06-??
3.6 | None | 2021-12-23 | N/A
3.7 | Buster | 2023-06-27 | ?

## 3.2 / 3.3

I propose we drop 3.2 and 3.3 from tox config now. They're EOL, must be hardly used, and occasionally cause incompatibility issues we have to spend more time fixing.

## 2.7

I propose that we drop 2.7 from tox config now, and start allowing Python 3-only syntax in the library. Any future releases will support Python 3 only, even if they take place in 2019.

A more conservative approach could be to wait for 1 January to drop it. I just don't see the point creating new Python 2 packages at this late stage. Debian has stopped accepting new Python 2 packages, and is currently removing Python 2 packages from Bullseye (testing).

When we do drop it, there's no need to do a massive refactoring to take advantage of the opportunity for Python only 3 code, but any cleanup can be done as we go along, and any new code can benefit from not having to support both 2 and 3.

## 3.4

Python 3.4 is EOL but its Debian version is not, until June 2020.

I propose we drop 3.4 in time for the Debian Jessie EOL in June.

## 3.5

Python 3.5's EOL is approaching in 2020, but it's in Debian Stretch which is EOL in 2022, so I propose we keep supporting it until then.

## Voting

I propose that the three core maintainers (@waveform80, @lurch and I) should vote on each proposal, and we'll go with the majority for each one.

1. Drop Python 3.2 now [y/n]
2. Drop Python 3.3 now [y/n]
3. Drop Python 2.7 now [y/n]
4. Drop Python 2.7 in January (if 3 fails) [y/n]
5. Drop Python 3.4 in June [y/n]

There's no need to vote on newer Python versions for now, but the decisions taken here will provide precedent for future decisions.

Note that if a future gpiozero release drops support for 2.7, it may become not-installable for Python 2, but older gpiozero versions will still be available. If a future gpiozero release drops support for a Python 3.x version it may simply be that we say we don't support that version any more, but it may still be installable, may even work, but we would not guarantee it. If possible, we would specify this in the meta tags on PyPi to make it not-installable for unsupported 3.x versions.

fangchenli:
closes #799


waveform80:
Brilliant! This is a backwards incompatible change so it'll need to wait for 2.x, but this is something we need to start planning. It'll need rebasing given some of the recent changes, but this is an excellent start on that. I'll stick a 2.0 milestone on it so we don't forget it.

waveform80:
This is really great - thanks so much for keeping the branch up to date. I've had a check over this and everything looks good (thanks for tidying up the redundant imports too!); don't worry about that last test failure - just a trivial missing import for a recent change. I'll tidy that and one or two other bits up and merge this. Thanks again!

Issue #915, opened 2021-01-23
`MockFactory` instances not independent?


pmlk:
Operating system: Raspbian Buster / MacOS 11
Python version: 3.9
Pi model: Raspberry Pi Model B Plus Rev 1.2
GPIO Zero version: 1.5.1
Pin factory used: MockFactory / PiGPIOFactory

## Issue

It seems to be impossible to create two independent `MockFactory` instances. 

## Expectations

* be able to create multiple devices using the same pins with different instances of a `MockFactory`
* each `MockFactory` instance mocks a single, independent RasPi

## Notes

Interestingly, this works just fine with two instances of `PiGPIOFactory` even when they control the same pins on the same physical RasPi.

## Details

I would expect both these tests to pass (by not throwing any exceptions):

```python
import time

from gpiozero import DigitalOutputDevice
from gpiozero.pins.mock import MockFactory
from gpiozero.pins.pigpio import PiGPIOFactory


pin_number = 23


def test_multiple_factories_same_pin():
    pi_dns = "raspberrypi.local"
    factory_1 = PiGPIOFactory(host=pi_dns)
    factory_2 = PiGPIOFactory(host=pi_dns)

    led_1 = DigitalOutputDevice(pin_number, pin_factory=factory_1)
    led_2 = DigitalOutputDevice(pin_number, pin_factory=factory_2)

    led_1.on()
    time.sleep(1)
    led_2.off()


def test_multiple_mock_factories_same_pin():
    factory_1 = MockFactory()
    factory_2 = MockFactory()

    led_1 = DigitalOutputDevice(pin_number, pin_factory=factory_1)
    led_2 = DigitalOutputDevice(pin_number, pin_factory=factory_2)

    led_1.on()
    led_2.off()


if __name__ == "__main__":
    test_multiple_factories_same_pin()
    test_multiple_mock_factories_same_pin()
```

`test_multiple_mock_factories_same_pin` fails with `gpiozero.exc.GPIOPinInUse: pin 23 is already in use by <gpiozero.DigitalOutputDevice object on pin GPIO23, active_high=True, is_active=False>`.

I had a quick look at the code and added some print statements after looking at the abstract `Factory` in `gpiozero/pins/__init__.py`:

```python
    print(factory_1._reservations)
    print(factory_2._reservations)

    led_1 = DigitalOutputDevice(pin_number, pin_factory=factory_1)

    print(factory_1._reservations)
    print(factory_2._reservations)

[...]

PR #919, opened 2021-01-31
Independent mock factory instances

pmlk:
Addresses #915 

This allows to create multiple completely independent `MockFactory` instances, creating independent pins as shown by the test.

I honestly feel like this is pretty hacky. Digging through the code I found that `LocalPiFactory` defines and uses class-wide variables: 

```python
class LocalPiFactory(PiFactory):
    # ...
    pins = {}
    _reservations = defaultdict(list)
    _res_lock = Lock()

    def __init__(self):
        # ...
        self.pins = LocalPiFactory.pins
        self._reservations = LocalPiFactory._reservations
        self._res_lock = LocalPiFactory._res_lock
```

So the newly introduced `independent` argument in the `MockFactory.__init__` pretty much just reverses the above back to the behavior of `PiFactory`.

Please, let me know if there's a better way!

### DIFF
--- a/gpiozero/pins/mock.py
+++ b/gpiozero/pins/mock.py
@@ -40,7 +40,7 @@ str = type('')
 import os
 from collections import defaultdict, namedtuple
 from time import time, sleep
-from threading import Thread, Event
+from threading import Lock, Thread, Event
 try:
     from math import isclose
 except ImportError:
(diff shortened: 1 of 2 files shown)


waveform80:
Closing in favour of #946

PR #920, opened 2021-02-05
Fix bugs related to PiGPIOSoftwareSPI cleanup

knmorgan:
I was initially encountering the following stack trace when using the pigpio pin factory.

```
<...>
File "/xxx/gpiozero/pins/pigpio.py", line 499, in close
    self.factory.release_all(self)
AttributeError: 'PiGPIOSoftwareSPI' object has no attribute 'factory'
```

I discovered this was due to a misnamed variable, which I addressed in the first commit. Afterward, I noticed the following stack trace.

```
<...>
Error in atexit._run_exitfuncs:
  File "/xxx/gpiozero/devices.py", line 595, in _shutdown
    _devices_shutdown()
  File "/xxx/gpiozero/devices.py", line 588, in _devices_shutdown
    dev.close()
  File "/xxx/gpiozero/pins/pigpio.py", line 500, in close
    self._factory.connection.bb_spi_close(self._select_pin)
AttributeError: 'NoneType' object has no attribute 'bb_spi_close'
```

This was due to a combination of another misnamed variable and what seems to be a race condition where PiGPIOFactory object is closed before the software SPI object can close it. Both of these are addressed in the second commit.

### DIFF
--- a/gpiozero/pins/pigpio.py
+++ b/gpiozero/pins/pigpio.py
@@ -495,7 +495,7 @@ class PiGPIOSoftwareSPI(SPI, Device):
             # If the factory has died already or we're not present in its
             # internal list, ignore the error
             pass
-        if not self.closed:
+        if not self._closed and self._factory.connection:
             self._closed = True
             self._factory.connection.bb_spi_close(self._select_pin)
         self._factory.release_all(self)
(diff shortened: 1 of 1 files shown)


waveform80:
You're absolutely right, and this PR did indeed fix an issue in SPI. Unfortunately, SPI was even more of a mess than I'd remembered, so I wound up patching up lurch's earlier SPI cleaning PR and doing a bit of work on top of that to properly clean things up which has made this PR obsolete - but many thanks for submitting it, and sorry it took a while to look at!

knmorgan:
@waveform80 Are you referring to #929? Because from what I can tell it does not address the misnamed variables addressed in this PR.

PR #927, opened 2021-02-28
Add lgpio pin implementation

waveform80:
This is basically complete at this point and tested successfully under Ubuntu with the experimental lgpio packages available from the following PPA:

https://launchpad.net/~waveform/+archive/ubuntu/lg

Might be worth waiting for that package to be refined and added to RaspiOS (which I'm working on next week), but this could be approved before then anyway; won't harm anything as lgpio is still listed as secondary to RPi.GPIO in the default preference order. Bear in mind that this will *not* be testable on RaspiOS at the moment as the current kernel version lacks the constants for setting pulls on the GPIOs (but that should be remedied with the move to 5.10).

### DIFF
--- a/docs/api_pins.rst
+++ b/docs/api_pins.rst
@@ -347,6 +347,16 @@ RPi.GPIO
 .. autoclass:: gpiozero.pins.rpigpio.RPiGPIOPin
 
 
+lgpio
+=====
+
+.. module:: gpiozero.pins.lgpio
+
+.. autoclass:: gpiozero.pins.lgpio.LGPIOFactory
+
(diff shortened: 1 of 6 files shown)


bennuttall:
:+1: I'm going to test it out

It's missing inclusion in the docs e.g. on the `api_pins` page.

waveform80:
>  I'm going to test it out
> 
> It's missing inclusion in the docs e.g. on the `api_pins` page.

Good point - just a mo

bennuttall:
LGTM - I have tested on Ubuntu 20.04 after changing gpio permissions and GPIO and PWM work fine.

I haven't managed to get SPI working.

PR #482, opened 2016-10-05
Add Rotary Encoder component support

SrMouraSilva:
Fix RotaryEncoder support based in @lurch suggestions (https://github.com/PedalPi/Physical/issues/1)

---
This change is [Reviewable](https://reviewable.io/reviews/rpi-distro/python-gpiozero/482)

PR #928, opened 2021-02-28
Add RotaryEncoder support

waveform80:
Updated PR for #482 - adds RotaryEncoder input device (based on CompositeDevice), with when_rotated_* events, wait_for_rotate_* gate methods, comprehensive tests, and a basic recipe in the docs. Though I've just realized I've probably forgotten to include the component in the Input Devices API docs ... I'll fix that in a mo!


yurikleb:
That would be amazing to have Rotary Encoders included in this library.
Please let me know if there is any way to help/test?
I'm using this encoder for my project: https://www.4project.co.il/documents/doc_3241_2836.pdf

bennuttall:
Can't test this with real hardware as I don't have one but the API and docs LGTM.

My only suggestion would be (and we could probably do this in other known composite devices):

Override the `__repr__` for this class to return:

```
<gpiozero.RotaryEncoder object on pins GPIO4 and GPIO5>
```

rather than:

````
<gpiozero.RotaryEncoder object containing 2 devices: a, b>
```

bennuttall (approved):

bennuttall:
I've added a commit - if it's not ok for whatever reason just remove it and merge.

waveform80:
> That would be amazing to have Rotary Encoders included in this library.
> Please let me know if there is any way to help/test?
> I'm using this encoder for my project: https://www.4project.co.il/documents/doc_3241_2836.pdf

I'd certainly be interested in getting as many rotary encoders tested with this as possible. I've tested two different models ([this one](https://shop.pimoroni.com/products/rotary-encoder-illuminated-rgb) and [this one](https://shop.pimoroni.com/products/ec11-rotary-encoder-module)) which have both worked perfectly, and frankly I expect this class will work pretty well with any rotary encoder that has two output pins (usually labelled A and B it appears).

What I'm most interested in finding out is whether all/most encoders share the habit of treating the A pin as the first edge when rotating clockwise, and the B pin as the first when rotating anti-clockwise. That's the case in the two I've got, but it'd be nice to confirm that. If it's not the case, it's no big deal: we can simply add a note to the docs saying something like "encoder appearing to run backwards? Flip the pin specs around and you're good to go".

As to how you can test this, it should be reasonably simple: follow the instructions at the top of the [Development chapter](https://gpiozero.readthedocs.io/en/stable/development.html) in the docs to set yourself up with a clone of this repo in a venv. Then, simply run `git checkout rotary` when you're in the clone's directory and you can try out the new class. I've just set this branch to build in the docs so once that's complete you should find [a recipe](https://gpiozero.readthedocs.io/en/rotary/recipes.html#rotary-encoder) and [API docs](https://gpiozero.readthedocs.io/en/rotary/api_input.html#rotaryencoder).

waveform80:
> Can't test this with real hardware as I don't have one but the API and docs LGTM.
> 
> My only suggestion would be (and we could probably do this in other known composite devices):
> 
> Override the `__repr__` for this class to return:
> 
> ```
> <gpiozero.RotaryEncoder object on pins GPIO4 and GPIO5>
> ```
> 
> rather than:
> 
> ```
> <gpiozero.RotaryEncoder object containing 2 devices: a, b>
> ```

Oh, good point - that is nicer. *Might* be worth doing something like A=GPIO4 and B=GPIO5 but, given my earlier comment (that sometimes it may be necessary to flip the definitions if some encoders disagree on which edge goes first in each direction), it may well be preferable to leave that out.

bennuttall:
I've created a beta release from this branch if anyone wants to test it out:

```
pip3 install https://bennuttall.com/files/gpiozero-1.6.0b0-py3-none-any.whl
```

The docs for rotary encoder can be found here: https://gpiozero.readthedocs.io/en/rotary/api_input.html#rotaryencoder with an example recipe here: https://gpiozero.readthedocs.io/en/rotary/recipes.html#rotary-encoder

waveform80:
Bah, did not proof-read my docs properly: left an AbsoluteEncoder reference in there (subtle hint for another branch I've got lying around), and "the B pint to GPIO 20" ... hmmm, no clue what I was thinking there! Ahem.

yurikleb:
Just tested with [this encoder](https://www.sparkfun.com/products/11102) and it works really well.

I will test the clockwise/counterclockwise A/B pins behaviour a bit later and will update.

One thing to note is the encoder reading losing some of the step counts if the PI CPU is under a heavy load.
I noted this when running my code from Visual Studio Code vs Directly from the Terminal.
As Visual Studio takes quite a bit of the CPU it seems to affect the speed of the encoder readings, and as a result, some steps are lost. It might not be an issue for lower resolution encoders, but something to take an account if your project requires precision.

waveform80:
> One thing to note is the encoder reading losing some of the step counts if the PI CPU is under a heavy load.
> I noted this when running my code from Visual Studio Code vs Directly from the Terminal.
> As Visual Studio takes quite a bit of the CPU it seems to affect the speed of the encoder readings, and as a result, some steps are lost. It might not be an issue for lower resolution encoders, but something to take an account if your project requires precision.

Yup, that'll partially depend on the underlying pin library. If you're using RPi.GPIO (the default), then the pin handlers are running in the same thread as edge detection and so under load it's possible to "lose" edges. If you try the pigpio layer instead, it should be more reliable as that buffers the pin states before feeding them to the handlers. In fact, even the native layer *might* be better (although it's terribly slow in all other aspects, being pure Python) as it employs one thread to watch for edges and queues up events to a separate thread that dispatches them.

I tried it out today on a rotary encoder attached to an old Pi 1, and deliberately ran it under the native pin library for "maximum slowness!". Even with the CPU idle, I managed to lose a few steps on that, but that's not too surprising. In fact, given that was under the pure Python driver I was pretty happy with it overall :)

PR #900, opened 2020-10-26
Add complete Pi4 support to NativePin. Fixes #895 and #777

lurch:
Also add more comprehensive pull-state tests.

**MUCH** overdue, but I finally got round to dusting off, polishing and testing my old patch for NativePin support for Pi4 :tada: 

Tested on: Pi 1A, Pi 1A+, Pi Zero WH, Pi 2B, Pi 3A+, Pi 4B. (the testing felt like it took longer than the coding!)

PR #929, opened 2021-03-02
Fix NativePin for Pi 4

waveform80:
This is a re-factor and re-base of @lurch's #900 PR. The initial commit is from the original PR but re-based onto the current master branch, and I've done a bit of re-factoring work in the second commit which makes the d-t querying a bit more generic and splits the pull-handling in two refinements of the NativePin class.

### DIFF
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -205,6 +205,17 @@ class GPIOMemory(object):
         else:
             offset = 0
         self.mem = mmap.mmap(self.fd, 4096, offset=offset)
+        # Register reads and writes must be in native format (otherwise
+        # struct resorts to individual byte reads/writes and you can't hit
+        # half a register :). For arm64 compat we have to figure out what the
+        # native unsigned 32-bit type is...
+        try:
+            self.reg_fmt = {
+                struct.calcsize(fmt): fmt
+                for fmt in (nstr('@I'), nstr('@L'))
+            }[4]
+        except KeyError:
+            raise RuntimeError('unable to find native unsigned 32-bit type')
 
     def close(self):
         self.mem.close()
@@ -222,10 +233,10 @@ class GPIOMemory(object):
         raise IOError('unable to determine gpio base')
 
     def __getitem__(self, index):
-        return struct.unpack_from(nstr('<L'), self.mem, index * 4)[0]
+        return struct.unpack_from(self.reg_fmt, self.mem, index * 4)[0]
 
     def __setitem__(self, index, value):
-        struct.pack_into(nstr('<L'), self.mem, index * 4, value)
+        struct.pack_into(self.reg_fmt, self.mem, index * 4, value)
 
 
 class GPIOFS(object):
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -620,8 +620,8 @@ class Native2711Pin(NativePin):
 
     def _reg_init(self, factory, number):
         super(Native2711Pin, self)._reg_init(factory, number)
-        self._pull_offset = self.factory.mem.GPPUPPDN_OFFSET + ((number * 2) // 32)
-        self._pull_shift = (number * 2) % 32
+        self._pull_offset = self.factory.mem.GPPUPPDN_OFFSET + (number // 16)
+        self._pull_shift = (number % 16) * 2
 
     def _get_pull(self):
         pull = (self.factory.mem[self._pull_offset] >> self._pull_shift) & 3
@@ -636,7 +636,8 @@ class Native2711Pin(NativePin):
             value = self.GPIO_PULL_UPS[value]
         except KeyError:
             raise PinInvalidPull('invalid pull direction "%s" for pin %r' % (value, self))
-        regvalue = self.factory.mem[self._pull_offset]
-        regvalue &= ~(0x3 << self._pull_shift)
-        regvalue |= (value & 0x3) << self._pull_shift
-        self.factory.mem[self._pull_offset] = regvalue
+        self.factory.mem[self._pull_offset] = (
+            self.factory.mem[self._pull_offset]
+            & ~(3 << self._pull_shift)
+            | (value << self._pull_shift)
+            )
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -311,7 +311,7 @@ class GPIOFS(object):
                 pass
             else:
                 try:
-                    with io.open(self.path('unexport'), 'w+b') as f:
+                    with io.open(self.path('unexport'), 'wb') as f:
                         f.write(str(pin).encode('ascii'))
                 except IOError as e:
                     if e.errno == errno.EINVAL:
--- a/gpiozero/pins/native.py
+++ b/gpiozero/pins/native.py
@@ -278,7 +278,7 @@ class GPIOFS(object):
                                          os.O_RDONLY | os.O_NONBLOCK)
                     except IOError as e:
                         if e.errno == errno.ENOENT:
-                            with io.open(self.path('export'), 'w+b') as f:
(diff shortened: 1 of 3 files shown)


waveform80:
Tested under Python 2.7 as well, locally on a Pi 4. I've also tested the implementation on an old Pi 1, Pi 2, Pi 3B+, and a Pi 4.

lurch:
I'm happy to give this the thumbs-up if you've verified it as working properly on various different models of RPi. :+1: 

Will be good to see this finally merged, so long after the release of the Pi4 :grinning:

waveform80:
> I'm happy to give this the thumbs-up if you've verified it as working properly on various different models of RPi. +1
> 
> Will be good to see this finally merged, so long after the release of the Pi4 grinning

Absolutely, and yes please do - I've verified it under current RaspiOS on a Pi1 (2835), Pi2 (2836), Pi3+ (2837), and Pi4 (2711). Haven't given it a whirl yet on a CM4/400 but I did verify device-tree parsing code was being correctly used in all those cases rather than the dict-fallback so I'd be reasonably confident that even if the CM4/400 *did* have a difference there (which I'm sure they don't) it shouldn't matter provided the d-t reflected it accurately.

If you've got any spare time, you could cast your expert eye over the RotaryEncoder PR? (I've been noticing I'd missed stupid things like the bounce-time initializers!) Not to worry if not though - it's had a bit of third-party testing already so I'm reasonably confident in it. Just having a crack at getting something for 7-seg displays in too (singletons only). Basically I'm hoping we can cut a release this week, as I need to get a version of gpiozero which includes lgpio into Ubuntu 21.04 and we're already into the freeze period.

lurch:
> even if the CM4/400 _did_ have a difference there (which I'm sure they don't)

The source-code for the DeviceTree stuff is all at https://github.com/raspberrypi/linux/tree/rpi-5.10.y/arch/arm/boot/dts if you wanted to investigate the differences between different models.

> If you've got any spare time, you could cast your expert eye over the RotaryEncoder PR?

Normally I would, but I'm afraid I've been quite busy with https://github.com/raspberrypi/pico-sdk/ recently....

It's been a _long_ time since I've looked at GpioZero, so I suspect it'd take me quite a while to get up to speed again.

gpiozero/pins/native.py
```diff
@@ -324,7 +425,10 @@ def __init__(self):
         self.mem = GPIOMemory(self.pi_info.soc)
         self.fs = GPIOFS(self, queue)
         self.dispatch = NativeDispatchThread(self, queue)
-        self.pin_class = NativePin
+        if self.pi_info.soc == 'BCM2711':
+            self.pin_class = Native2711Pin
+        else:
+            self.pin_class = Native2835Pin
```
lurch, line 431: So I guess this means that `NativePin` itself will never be directly used? Would it be worth putting in an exception or something to enforce that?

lurch (approved):
I've not tested this, but it looks similar enough to my previous PR, which I did test.

gpiozero/pins/native.py
waveform80: Yes, NativePin is effectively an abstract class at this point. I've never been hugely fussed about banning the instantiation of abstract classes; it's one of those issues that's come up sufficiently rarely and (provided abstract methods are marked with some appropriate error like NotImplementedError) is pretty obvious if/when it does that it's not a headache to track down. In this case I think I'll leave it - I don't particularly want to complicate the hierarchy at the pin level with Python's abstractclass machinery.

waveform80:
> > even if the CM4/400 _did_ have a difference there (which I'm sure they don't)
> 
> The source-code for the DeviceTree stuff is all at https://github.com/raspberrypi/linux/tree/rpi-5.10.y/arch/arm/boot/dts if you wanted to investigate the differences between different models.

Thanks for the link! I've dug through diffs of the d-t's between Ubuntu and RaspiOS before (usually when debugging u-boot stuff) but haven't had cause to dig into the GPIO related nodes before. I'll take a look a bit later! 

> > If you've got any spare time, you could cast your expert eye over the RotaryEncoder PR?
> 
> Normally I would, but I'm afraid I've been quite busy with https://github.com/raspberrypi/pico-sdk/ recently....
> 
> It's been a _long_ time since I've looked at GpioZero, so I suspect it'd take me quite a while to get up to speed again.

Fair enough :) The Pico SDK's docs are looking very professional (been skimming the C PDF ones, though I've yet to find time to actually *use* it yet)! I should take a look at asciidoc at some point...

waveform80:
> I've not tested this, but it looks similar enough to my previous PR, which I did test.

Yup, same idea basically: query the d-t for the GPIO registers, map the address to the CPU's view, fallback to the model lookup in a dict if that fails for I/O related reasons. I'll get @bennuttall to test it on some hardware later (so it's more than just me that's run this :) then we can merge it.

gpiozero/pins/native.py
lurch: Should `_get_pull` and `_set_pull` "base methods" in `NativePin` raise `NotImplementedError` ?

lurch:
> The Pico SDK's docs are looking very professional (been skimming the C PDF ones, though I've yet to find time to actually use it yet)!

Thanks :grin: Getting the Doxygen included into the AsciiDoc was "challenging" :wink: Found three different projects that claimed to be able to do that, but none of them turned out to be suitable for our purposes, so I had to do it via XSLT, which was a bit of a steep learning curve (99% of XSLT tutorials are geared at doing simple XML->HTML translations).

gpiozero/pins/native.py
waveform80: `_set_pull` effectively does by inheritance (admittedly not NotImplementedError, I think it's PinFixedPull, but still that'd be a pretty big hint to anyone running into it that an override for `_set_pull` was missing. But yes, `_get_pull` is missing one (it'll just return 'floating' by default). Yup, you've convinced me: I should chuck a `NotImplementedError` into those (won't cost anything at runtime as the super-class should never get called for those). Will do that in a mo.

waveform80:
> > The Pico SDK's docs are looking very professional (been skimming the C PDF ones, though I've yet to find time to actually use it yet)! I should take a look at asciidoc at some point...
> 
> Thanks grin Getting the Doxygen included into the AsciiDoc was "challenging" wink Found three different projects that claimed to be able to do that, but none of them turned out to be suitable for our purposes, so I had to do it via XSLT, which was a bit of a steep learning curve... (99% of XSLT tutorials are geared at doing simple XML->HTML translations).

Ooohh, I'm actually a bit of a fan of XSLT - the new version of structa I've been working on uses it in possibly the stupidest way I've tried yet (as an interface rendering layer for the Linux terminal :). I really must get that finished at some point... And yes, I remember it being quite horrific to learn. I think I wound up just reading the spec and experimenting until a light-bulb went on it my head!

lurch (approved):

waveform80:
Well, getting this working on Ubuntu (and thus arm64 too) turned out to be more interesting than I'd expected, but it's now working and passing the test suite. @bennuttall this is ready for merge :)

gpiozero/pins/native.py
```diff
@@ -205,6 +205,17 @@ def __init__(self, soc):
...
         self.mem = mmap.mmap(self.fd, 4096, offset=offset)
+        # Register reads and writes must be in native format (otherwise
+        # struct resorts to individual byte reads/writes and you can't hit
+        # half a register :). For arm64 compat we have to figure out what the
+        # native unsigned 32-bit type is...
+        try:
+            self.reg_fmt = {
+                struct.calcsize(fmt): fmt
+                for fmt in (nstr('@I'), nstr('@L'))
+            }[4]
```
lurch, line 216: Oh boy... more impenetrable @waveform80 code :wink:

lurch:
Whoop! :tada:

PR #930, opened 2021-03-10
Seven segment character displays (single and multiple)

waveform80:
The continuation of #488.

I re-worked the original PR to provide the character string assigned via the "display" method as the string value of the component, and added some logic to handle the reverse mapping (deriving the string state from the LED state).

This did raise a few complications; it's entirely possible, even likely, to have a n:1 mapping of strings to LED states (consider 5 and S, or 0 and O) in which case which reverse mapping should be selected, should it be stable, etc. etc. Then the question arose of how to deal with the decimal point LED given that it essentially applies to all states, and doubling the size of the mapping isn't particularly pleasant. Hence, I've dealt with that as an effectively distinct state (and thus character) in the value. Hence one can assign "A" to the display, or "A." and in the latter case it'll happily accept the trailing "." as an indication of the decimal-point's state, then using the remaining prefix to determine the state of the rest of the LEDs.

At that point I decided it would be quite nice if it could be made generic (for n-segment displays) so I added a mapping parser from a simple text format and moved the default mapping into a resource'd text file (see `gpiozero/fonts/7seg.txt` in the new PR, along with `gpiozero/fonts/14seg.txt`).

Initially, I'd decided that there wasn't time to deal with the myriad questions that a multi-character display generates (threading etc. etc.) so I'd started by excising that chunk of the PR, but then I wound up hacking on it over the weekend and getting further than expected so that's wound up back in this PR too. Given that the single character display now has a string type, it made obvious sense for the multi-character one to do likewise. This also generated a few issues regarding the multiple decimal points often available in these displays but the simple solution is to allow any *sequence* of strings, whether that's a multi-character string (with no decimal points) or a *tuple* of strings which can thus explicitly split at decimal points.

I've tweaked the threading stuff from the original quite extensively as well: the new code pre-calculates the LED transitions before starting the thread and knows how to activate multiple characters simultaneously if their LED states are equivalent.

Full coverage tests are included for all the new classes.


waveform80:
Fonts split from tools into separate module (and docs chapter) as requested.

bennuttall (approved):
Just fixed a reference to `tools.load_segment_font` but otherwise LGTM

PR #941, opened 2021-03-13
Implement polling for internal devices

### LINKED ISSUE #393: Internal Device events don't fire
QUESTION Can the "TimeOfDay" be used with LED or Buzzer as I have tried using the LED but on hour change nothing changes and the  LED carries on flashing due to the activated and deactivated giving it a 5v or 0 never just one of them being actioned. TerryR

waveform80:
And clean up event handling generally. All the logic to do with wrapping callbacks, and ensuring watch threads is set up should be handled by the EventsMixin or other helper classes (in this case a property analogue called "event").

Fixes #393

### DIFF
--- a/docs/api_internal.rst
+++ b/docs/api_internal.rst
@@ -88,8 +88,8 @@ attributes, so alternative methods using the other paradigms would also work.
 .. note::
     Note that although the constructor parameter ``pin_factory`` is available
     for internal devices, and is required to be valid, the pin factory chosen
-    will not make any difference. Reading a remote Pi's CPU temperature, for
-    example, is not currently possible.
+    will not make any practical difference. Reading a remote Pi's CPU
+    temperature, for example, is not currently possible.
 
 
 Regular Classes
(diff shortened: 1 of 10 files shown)


gpiozero/mixins.py
```diff
@@ -197,11 +233,27 @@ def __init__(self, *args, **kwargs):
...
         self._active_event = Event()
         self._inactive_event = Event()
-        self._when_activated = None
-        self._when_deactivated = None
         self._last_active = None
         self._last_changed = self.pin_factory.ticks()
 
+    def _all_events(self):
+        """
+        Generator functionw which yields all :class:`event` instances defined
```
bennuttall, line 241: `functionw`

bennuttall:
The (inherited) docstring for `when_activated` / `when_deactivated` says:

```
If the function accepts a single mandatory parameter, the device that
activated will be passed as that parameter.
```

but as far as I can tell, it doesn't get passed.

Probably not as useful as it would be for GPIO devices, but I can still see a scenario where multiple events could trigger a callback and you'd want to know which one. But there are ways around it obviously.

I'm just adding class-specific docstrings for `.when_activated` and `when_deactivated` so I'll remove that bit for now.

bennuttall:
Correction: it *does* pass it. I was mistaken.

bennuttall:
I've added an intro to the internal devices docs page, added `when_activated` and `when_deactivated` to each class to precisely document what causes them to fire, and exposed the new `PolledInternalDevice` class at module top-level (partly because the docs couldn't find it when I included it, but still it could be useful there).

I've also switched some source/values examples to when_activated/deactivated.

A couple of thoughts:

- We could add aliases like `when_available` (or something) as `google.when_activated` doesn't quite seem right
- Should we expose `event_delay` in subclasses, so people can overwrite it? Or should we at least document what it's set to for each one.

bennuttall:
Is `wait_for_active` hooked up?

waveform80:
> Is `wait_for_active` hooked up?

It should be implicitly by `EventsMixin._fire_events` - if it doesn't work, that's a bug.

docs/api_internal.rst
```diff
@@ -40,11 +40,56 @@ GPIO Zero also provides several "internal" devices which represent facilities
...
+:class:`TimeOfDay` object has :attr:`~TimeOfDay.value`,
+:attr:`~TimeOfDay.values`, :attr:`~TimeOfDay.is_active`,
+:attr:`~TimeOfDay.when_activated` and :attr:`~TimeOfDay.when_deactivated`
+attributes, so alternative methods using the other paradigms would also work.
+
+.. note::
+    Note that although the constructor parameter ``pin_factory`` is available
+    for internal devices, and is required to be valid, the pin factory chosen
+    will not make any difference. Reading a remote Pi's CPU temperature, for
+    example, is not currently possible.
```
waveform80, line 92: Not *precisely* true, but probably true enough for the purposes of the documentation. The pedant in me, however, wants to point out that it *will* make a difference to whatever calls the pin-factory's `ticks` method (which `EventsMixin` does for the purpose of figuring out the timing of events to report to handlers, like `Button.hold_time`). However, I don't *think* anything in internal devices actually uses the ticks value beyond passing it around because it has to. So, yeah, don't bother correcting this my inner pedant!

docs/faq.rst
```diff
@@ -393,6 +393,61 @@ curve for beginners by making it easy to get started and easy to build up to
 more advanced projects.
 
 
+What's the difference between when_pressed, is_pressed and wait_for_press?
+==========================================================================
```
waveform80, line 397: Niiiiiiice!

bennuttall: Definitely needed - but hard to capture the "question" - it's usually people saying "my code doesn't work" and they've just done something like reassigned `when_pressed` in a loop - at least we'll have something to link to. Also hard to cover the general case but it should help.

docs/api_internal.rst
bennuttall: "will not make any *practical* difference" is probably more accurate - and if at any point this changes we can update the note.

bennuttall (approved):

PR #943, opened 2021-03-18
Add details about v1.6.2

bennuttall:
...and correct docs referring to 1.6.0 as the last version supporting Python 2

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,8 +128,8 @@ Python 2 support
 
 .. warning::
 
-    Version 1.6.0 of GPIO Zero is the last to support Python 2. The next release
-    will be Version 2.0.0 and will support Python 3 only.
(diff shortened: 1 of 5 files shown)


bennuttall:
Release from [branch v162](https://github.com/gpiozero/gpiozero/tree/v162) but merge these changes into master

waveform80:
It might be better to simply state 1.6.x will be the final Python 2.x supporting release just in case there's anything else we need to fix up in this branch (critical bugs or what-not). Just implies that there won't be a 1.7.x with new features that still supports Python 2.x - basically critical bug fixes only for 1.6.x (any new features will be 2.x, etc.)

bennuttall:
I did consider that... may as well - bear with me

bennuttall:
Done

bennuttall:
Just updated the wording on both branches

PR #944, opened 2021-03-19
Robots done right

waveform80:
A probably controversial change for 2.x, hence @bennuttall ought to decide whether or not this should go in. `PhaseEnableRobot` is basically a redundant duplicate of `Robot`. However, `Robot` has grown several features which haven't been copied over to `PhaseEnableRobot` (notably curved cornering). Moreover, when I implemented "proper" keyword-args after we dropped py2.x support several tests failed because we were assuming that the "enable" parameter for Motor, which we'd documented (though not enforced) as keyword-only, was positional and could be represented as a tuple.

A cleaner solution all round is to ditch `PhaseEnableRobot` (removing the duplication), and just have Robot constructed from two `Motor`s *or* `PhaseEnableMotor`s (or, weirdly a mix-and-match, although that's probably not very practical). After all, Robot doesn't really care *what* left and right motors it has as long as they can go "forwards" or "backwards" (in other words as long as they're objects with those methods).

However, as noted this *is* a breaking change - it'll break the code of anyone currently using `Robot`. Potentially, we *could* alleviate some of that by, say, detecting tuples being passed in and implicitly converting them to a `Motor` instance but of course that doesn't solve the breakage for anyone using `PhaseEnableMotor`.

The PR's also incomplete: I haven't gotten around to updating the recipes/examples code - but there is an entry in the deprecation chapter. Let me know if you think it's worth finishing it off or if it would be acceptable but needs the mitigations mentioned above added.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [3.5, 3.6, 3.7, 3.8]
 
     steps:
     - uses: actions/checkout@v2
@@ -25,7 +25,6 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
         make develop
     - name: Run tests
       run: |
--- a/docs/compat.rst
+++ b/docs/compat.rst
@@ -138,6 +138,10 @@ enable robot boards now inherit all the functionality of :class:`Robot` because
 that's all they use. Theoretically you could also mix and match regular motors
 and phase-enable motors although there's little sense in doing so.
 
+The former functionality (passing tuples to the :class:`Robot` constructor)
+will remain as deprecated functionality for gpiozero 2.0, but will be removed
+in a future 2.x release.
+
 
 .. _Python documentation: https://docs.python.org/3/
 .. _porting guide: https://docs.python.org/3/howto/pyporting.html
--- a/docs/compat.rst
+++ b/docs/compat.rst
@@ -3,6 +3,8 @@ Backwards Compatibility
 =======================
 
 
+.. currentmodule:: gpiozero
+
 GPIO Zero 2.x is a new major version and thus backwards incompatible changes
 can be expected. We have attempted to keep these as minimal as reasonably
 possible while taking advantage of the opportunity to clean up things. This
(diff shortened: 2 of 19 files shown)


waveform80:
Ha! lurch was waaaaay ahead of me ... #219 (in fact this PR still isn't exactly what that RFC suggested but it's a step closer). Hmmm....

bennuttall:
I'm totally ok with this approach.

The only weird thing about creating devices within devices (which is not new) is you can do stuff like this:

```
In [13]: lb1 = LEDBoard(2,3,4)

In [14]: lb2 = LEDBoard(5,6,7)

In [15]: lb = LEDBoard(lb1, lb2)

In [16]: lb1.on()

In [18]: tuple(lb.value)
Out[18]: 
(LEDBoardValue(device_0=1, device_1=1, device_2=1),
 LEDBoardValue(device_0=0, device_1=0, device_2=0))

In [19]: lb1.close()

In [20]: lb
Out[20]: <gpiozero.LEDBoard object containing 2 unnamed devices>

In [21]: lb.on()

In [22]: tuple(lb.value)
TypeError: <lambda>() missing 3 required positional arguments: 'device_0', 'device_1', and 'device_2'
```

waveform80:
> I'm totally ok with this approach.
> 
> The only weird thing about creating devices within devices (which is not new) is you can do stuff like this:
> 
> [example demonstrating abuse of object lifetimes]

Yes - the blithe answer to that is "well, don't do that then!" :) The slightly more nuanced answer is that `CompositeDevice` doesn't currently "own" the things it's composed of. Or rather, it assumes it does but there's no mechanism enforcing this.

What would you *expect* to happen if you attempt to close an owned device? There's a few potential avenues I can see here:

* If you close a subordinate device, the owning device is now "broken" and the next interaction with it will raise an exception (that's more or less what happens at the moment but we could clean up the error message a bit)

* If you close a subordinate device it closes the owning device cleanly. I don't particularly like this approach because it encourages "bad behaviour" without punishing it with an exception :)

* If you attempt to close a subordinate device, it simply raises an exception and refuses to close. This would be cleaner than the current situation though it's going to make our shutdown code rather more complex; at the moment we just let everything call `__del__` on the way out and stuff closes in whatever order it wants to. With this mechanism, in order to avoid a flood of (ignored) exceptions on script shutdown we'd have to order the `__del__` calls, or have some mechanism of knowing when `__del__` was a result of script shutdown or not.

waveform80:
Added deprecation facilities so this is much less controversial now, and assuming all the tests pass I'll go ahead and merge this as I've got a ton of stuff I need to rebase on top of master now.

PR #945, opened 2021-03-19
Fix Button imports in the migrating_from_rpigpio

rmisev:
The gpiozero contains Button class, but not Buttons.


bennuttall:
:blush: thanks

PR #984, opened 2021-07-22
closes #983 - removed print statement left from debugging

### LINKED ISSUE #983: Unncessary/unintended print() statement on LoadAverage.load_average ?
Hi there,

I noticed that on `internal_devices.py` the `load_average` property of the `LoadAverage` class seems to unnecessarily print the representation of the opened file as well as the stored value:

```python
Python 3.7.3 (default, Jan 22 2021, 20:04:44) 
[GCC 8.3.0] on linux
>>> from gpiozero import CPUTemperature, DiskUsage, LoadAverage
>>> loadCPU = LoadAverage(min_load_average=0, max_load_average=1, minutes=1)    
>>> print(loadCPU.load_average)
<_io.TextIOWrapper name='/proc/loadavg' mode='r' encoding='UTF-8'>
0.76
>>> 
```

That doesn't seem to be the intended behavior, since similar properties for other classes do not behave this way:
```python
>>> diskUsagePercent = DiskUsage()
>>> print(diskUsagePercent.usage)
14.494867817330109
>>> 
```

This is the current implementation of `LoadAverage.load_average`, starting on line 434 of `internal_devices.py`, with the referred print statement after opening the `load_average_file`:

```python

    @property
    def load_average(self):
        """
        Returns the current load average.
        """
        with io.open(self.load_average_file, 'r') as f:
            print(repr(f))
            file_columns = f.read().strip().split()
            return float(file_columns[self._load_average_file_column])
```

Thank you for looking into it.

thiagosfsilva:
Removes the unintended print statement as per opened issue.

### DIFF
--- a/gpiozero/internal_devices.py
+++ b/gpiozero/internal_devices.py
@@ -437,7 +437,6 @@ class LoadAverage(PolledInternalDevice):
         Returns the current load average.
         """
         with io.open(self.load_average_file, 'r') as f:
-            print(repr(f))
             file_columns = f.read().strip().split()
             return float(file_columns[self._load_average_file_column])
 


waveform80:
/me checks the blame history ... doh!

bennuttall:
Ha! I would have bet on it being me :)

PR #1004, opened 2021-10-31
Add info for Raspberry Pi Zero 2 W

lurch:
I squished the name down to 'Zero2W' so that it'd still fit in the 6-character space for the model-name in the ASCII-art.
I don't have a Zero 2 W myself yet - this is using the info from https://github.com/raspberrypi/documentation/commit/58f878dd869e8155d61e4d582861e5d4752bfb70

### DIFF
--- a/gpiozero/pins/data.py
+++ b/gpiozero/pins/data.py
@@ -1015,6 +1015,7 @@ class PiBoardInfo(namedtuple('PiBoardInfo', (
                 0xe:  '3A+',
                 0x10: 'CM3+',
                 0x11: '4B',
+                0x12: 'Zero2W',
                 0x13: '400',
                 0x14: 'CM4',
                 }.get(revcode_type, '???')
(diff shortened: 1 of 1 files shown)


waveform80:
Hang on ... pretty sure I've got a PR for this in the private repo with full ASCII art and all! Let me check I got things right there and, if so I'll merge this and then rebase that on top -- thanks!

bennuttall:
The text should be "Zero 2 W" not "Zero2W". We already have "Zero" and "Zero W".

waveform80:
Yes, but unfortunately that won't fit in the pretty ASCII art (and rejigging it to fit is ... surprisingly hard; yes, I've tried!)

lurch:
...unless @waveform80 wants to tweak things so that the model-name-displayed-in-ASCII-art is actually different to the model name we use elsewhere in code? :shrug: (i.e. yet another dictionary-mapping)

PR #1006, opened 2021-11-25
RGBLED color list cycling

AnnoNuem:
Make the RGBLED fade from a color of a list to the next color of the list. Fades over all colors and back to the first color of the list. Uses colorzero to compute the gradient between two colors of the list. 

A new `_cycle_device` function handles the repeated cycling and is also called by the old `_blink_device` function.

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -1105,6 +1105,67 @@ class RGBLED(SourceMixin, Device):
             self._blink_thread.stop()
             self._blink_thread = None
 
+
+    def cycle(self, colors, steps=125, fps=25, n=None, background=True):
+        """ 
+        Make the device fade from a color of a list to the next color of the list.
+
+        :param list colors
+            List of colorzero colors
+
+        :type n: int or None
+        :param steps:
+            Number of steps to fade from one color to the next
+
+        :type n: int or None
+        :param fps:
+            Number of times the color is changed per second
+
+        :type n: int or None
+        :param n:
+            Number of times to fade through complete list; :data:`None` (the default) means forever.
+
+        :param bool background:
+            If :data:`True` (the default), start a background thread to
+            continue blinking and return immediately. If :data:`False`, only
+            return when the blink is finished (warning: the default value of
+            *n* will result in this method never returning).
+        """
+        
+        sequence = []
+        for i in range(len(colors)-1):
+            g = colors[i].gradient(colors[i+1], steps)
+            for c in g:
+                r,g,b = c.rgb
+                sequence.append(((r,g,b),1/fps))
+
+        self._stop_blink()
+        self._blink_thread = GPIOThread(
+            self._cycle_device,
+            (
+                sequence, n
+            )
+        )
+        self._blink_thread.start()
+        if not background:
+            self._blink_thread.join()
+            self._blink_thread = None
+
+
+    def _cycle_device(self, values, n):
+        sequence = (
+                cycle(values) if n is None else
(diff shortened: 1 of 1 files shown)


lurch:
Kind of a duplicate of #1003 ?

AnnoNuem:
> 

Similar yes. But the idea here is to cycle over several colors and not only fade between two colors. E.g. we can cycle over all colors by calling the function with red, green and blue. 

This could also be extented to the blink method to blink in different colors.

lurch:
https://github.com/gpiozero/gpiozero/pull/1003/files also adds a `cylcle_color` method.
Perhaps yourself and @tinitaet would like to collaborate on a common solution, rather than creating conflicting PRs? :slightly_smiling_face:

tinitaet:
@AnnoNuem I am open for a collaboration. We could share some approaches and probably find some common ground.

AnnoNuem:
@tinitaet 

I think I will not have any time for more coding for this project. However, I will take time for questions and discussion. 

My idea was a more flexible approach: A parser which can interpret an arbitrary amount of colors, durations and transitions similar to a sheet of music. The existing functions would then become `sheets of color` objects which can be given to the parser.  Each `sheet` object would contain `color` objects, and speed and repetition information. Each `color` object would contain color, duration and transition information. I also dreamed of midi compatibility and being able to parse midi files or instrument as color.

I hope I was at least able to give you some inspiration.

tinitaet:
@AnnoNuem 
This actually sounds like a very cool idea. The basic question one hast to ask is, of this should be a part of the `gpiozero` module of if this would rather profit from being its own module, that uses gpiozero. By doing this, one could write their own wrapper, data format and stuff for the mentioned midi-compatibility and other things, without bloating gpiozero. If you are still interested, you or I could open an new project here on github to get this started!
To be honest this immediately sparked some ideas, since a friend of mine composed a couple of songs for my role-playing-group in midi and it would be sick to see them light up the room!

AnnoNuem:
Personally I would prefer to keep it all in `gpiozero`, but if additional libraries for midi support are required, then possible another project makes sense. But that depends also on the other contributors of gpiozero. If you want to start right away creating a new project is probably the faster way.
As I said before I will not have time for coding, hence it would be better if you create a new project.
Also: I did not thoroughly check whether similar projects exist already...

lurch:
Just IMHO, but a "sheets of color" functionality sounds perhaps like something a bit too complex for a basic "RGBLED" class?
Of course @bennuttall and @waveform80 might disagree, and their opinion is more important than mine.

Perhaps the https://gpiozero.readthedocs.io/en/stable/source_values.html would be a good way for a 3rd-party project to connect a midi-parser into the rest of the GPIO Zero framework? :shrug:

lurch:
> a friend of mine composed a couple of songs for my role-playing-group in midi and it would be sick to see them light up the room!

Reminds me of https://www.raspberrypi.com/news/this-piano-controls-an-led-rainbow-with-raspberry-pi-zero/ :grinning:

AnnoNuem:
@tinitaet Just wondering: Did you made a decision already? Did you start a new project?

tinitaet:
@AnnoNuem not yet. Unfortunately I am also a bit busy at the moment.

NBoumakis:
I was thinking of developing something similar, a library that can parse some set of instructions and display patterns on various types of LEDs, e.g. a simple LED or a WS2812 matrix. I agree that it would be too complex for this library. Perhaps we can agree on some type of protocol that allows intercommunication.

Is anyone interested in a collaboration?

PR #1010, opened 2022-01-05
Ensure mock pin factory at device init

jacevan:


lurch:
Huh?
https://gpiozero.readthedocs.io/en/stable/api_pins.html#mock-pins shows how to use mock pins.

jacevan:
Sorry, I did this incorrectly.

PR #1040, opened 2022-04-30
add thread safety to queue access in GPIOQueue - fix #975

### LINKED ISSUE #975: motion_detected fails with exception - "RuntimeError: deque mutated during iteration"
*Originally reported in #973 by @dbrandesky and moved to discussions, but turned out to be a bug*

I have a fairly "simple" loop on a Raspberry Pi Zero W that runs two applications: the first basically generates random "arrivals" of a simulated train and activates GPIO pins when those events occur; the second is a function that monitors the state of an external motion detector, and passes that state into a function which will return True if motion is detected.

The exception is as follows:

```python
  File "TrainGenerator.py", line 107, in <module>
    detector_state = garage_motion(mqttc, topic_state, motion_sensor
)
  File "TrainGenerator.py", line 54, in garage_motion
    if sensor.motion_detected:
  File "/home/pi/TrainGenerator/venv/lib/python3.7/site-packages/gpi
ozero/input_devices.py", line 350, in is_active
    return self.value > self.threshold   
  File "/home/pi/TrainGenerator/venv/lib/python3.7/site-packages/gpi
ozero/input_devices.py", line 604, in value
    return super(MotionSensor, self).value
  File "/home/pi/TrainGenerator/venv/lib/python3.7/site-packages/gpi
ozero/input_devices.py", line 325, in value
    return self._queue.value
  File "/home/pi/TrainGenerator/venv/lib/python3.7/site-packages/gpi
ozero/mixins.py", line 587, in value
    return self.average(self.queue)
  File "/usr/lib/python3.7/statistics.py", line 311, in mean
    T, total, count = _sum(data)
  File "/usr/lib/python3.7/statistics.py", line 147, in _sum
    for n,d in map(_exact_ratio, values):
RuntimeError: deque mutated during iteration
```

The offending function is this:

```python
def garage_motion(mqtt_client, topic, sensor):
    if sensor.motion_detected:
        mqtt_publish(mqtt_client, topic, 'ON')
        return True
    else:
        mqtt_publish(mqtt_client, topic, 'OFF')
        return False
```

This function is run every 0.1 seconds in the main loop.

Full code can be found [here](https://pastebin.com/6zXvvWnX).

chaddcw:


bennuttall:
Thanks! @waveform80 will you take a look?

jogi-k:
Suddenly - after about 10 month working without any problems 24/7 - this issue occurred reproducible in my application, it didn't work any longer. I applied the pull-request on my installation. It works perfectly. 

:+1:

gpiozero/mixins.py
```diff
@@ -577,9 +579,11 @@ def fill(self):
             while not self.stopping.wait(self.sample_wait):
                 value = self.parent._read()
                 if value not in self.ignore:
```
lurch, line 581: Hmmm, if the `value` _is_ in `self.ignore`, then I guess `self.queue` won't get appended to, and therefore `len(self.queue)` won't change, and therefore there's no chance of `self.full` getting set either?
I.e. perhaps this function could change to:
```python
    def fill(self):
        try:
            while not self.stopping.wait(self.sample_wait):
                value = self.parent._read()
                if value not in self.ignore:
                    with self.lock:
                        self.queue.append(value)
                        if not self.full.is_set() and len(self.queue) >= self.queue.maxlen:
                            self.full.set()
                    if (self.partial or self.full.is_set()) and isinstance(self.parent, EventsMixin):
                        self.parent._fire_events(self.parent.pin_factory.ticks(), self.parent.is_active)
        except ReferenceError:
            # Parent is dead; time to die!
            pass
```
?

chaddcw: Moving the middle if in the function inside the first lock (and eliminating the second `with self.lock:`) looks like a good change.

I'm not sure about that last if in the function being protected by `if value not in self.ignore:`

The existing code allows `self.parent._fire_events()` to be activated when `value` is in `self.ignore` and `self.partial` is set.  The suggested change won't allow that to happen.

Thoughts?

bennuttall: I've added a commit to address these points. Does it look ok to you both?

Issue #1042, opened 2022-05-11
Motor class remote


gnicki2000:
I´m using Raspi zero 2 w. The Motor class doesen´t work via remote factory. LED and Servo work fine so it can´t be hardware or connection. But still Motor says it can´t connect. See: https://stackoverflow.com/questions/72163157/raspi-remote-gpio-pizero-problems

lurch:
Fixed by #792 ?

gnicki2000:
Thank you verry much. The 
super(Motor, self).__init__(_order=devices.keys(), **pin_factory=pin_factory,** **devices)
in line 1223 ist sill missing when installing gpiozero via pip install.

PR #792, opened 2019-08-18
fixing motors to work with pin_factory

yisraeldov:
This should fix #791 

It fixes my issue at least

PR #1043, opened 2022-05-11
Fix for remote Motor device

gnicki2000:
See Issue #1042 / #792

I hope I´m doing this the right way this is my first time using github....


lurch (approved):
LGTM :+1:

Issue #422, opened 2016-08-31
WeatherHAT support


waveform80:
Cyntech very kindly sent a [WeatherHAT](https://github.com/CyntechUK/WeatherHAT) over ages ago; would be nice to add support for it.

Note to self: push on proper I2C support in 1.4 and this (and the Touch4Pi they also very kindly sent) should become possible.

lurch:
And I guess with proper I2C support, it might be possible to support a bunch of different Pimoroni HATs too.

waveform80:
Yeah - it opens up a lot of possibilities as there's plenty of HATs out there with SPI and I2C

bennuttall:
Is there any general I2C support in progress anywhere? Does it rely on an i2c library or can we do it natively?

lurch:
There are I2C libraries out there https://duckduckgo.com/?q=python+i2c+raspberry+pi&t=canonical&ia=web but I've never tried using them so I don't know how good they are.
Maybe @Gadgetoid can give us a few pointers on what he uses in the Pimoroni software he writes? :-)

Issue #808, opened 2019-10-20
Add support for ADS1015 and ADS1115 analog-to-digital converters


recantha:
Device name: ADS1015 and ADS1115 (Texas Instruments)
Protocol: e.g. I2C
URL of example product: https://shop.pimoroni.com/products/adafruit-12-bit-adc
Further information e.g. data sheet: Datasheets as follows:
https://www.adafruit.com/datasheets/ads1015.pdf
https://www.adafruit.com/datasheets/ads1115.pdf

This would be really useful as the little breakout boards you can get from China are a lot cheaper than just buying an MCP3008 chip, for example. Thanks for your consideration :-)

Please provide example code for using this device with Python, using another library:
Using the Adafruit Circuit Python libraries:
https://github.com/adafruit/Adafruit_CircuitPython_ADS1x15

import time
import board
import busio
import adafruit_ads1x15.ads1015 as ADS
from adafruit_ads1x15.analog_in import AnalogIn

# Create the I2C bus
i2c = busio.I2C(board.SCL, board.SDA)

# Create the ADC object using the I2C bus
ads = ADS.ADS1015(i2c)

# Create single-ended input on channel 0
chan = AnalogIn(ads, ADS.P0)

# Create differential input between channel 0 and 1
#chan = AnalogIn(ads, ADS.P0, ADS.P1)

print("{:>5}\t{:>5}".format('raw', 'v'))

while True:
    print("{:>5}\t{:>5.3f}".format(chan.value, chan.voltage))
    time.sleep(0.5)

bennuttall:
Thanks for all the info, Mike. Once we've got around to basic I2C support, we'll take a look at this.

Issue #802, opened 2019-09-20
TonalBuzzer for i2c buzzer


pangoro24:
Operating system: e.g. Raspbian Buster
Python version: e.g. 3.5.3
Pi model: e.g. Pi 3 Model B
GPIO Zero version: e.g. 1.5.1

I want to try the code below, however, I am using an i2c buzzer. Is possible to play notes using this kind of buzzers? I am using smbus2 to write data to the i2c device.

```python
from gpiozero import TonalBuzzer
from gpiozero.tones import Tone
from time import sleep
t = TonalBuzzer(21) # change to whatever pin the buzzer is connected

v1 = ["G4", "G4", "G4", "D4", "E4", "E4", "D4"]
v2 = ["B4", "B4", "A4", "A4", "G4"]
v3 = ["D4", "G4", "G4", "G4", "D4", "E4", "E4", "D4"]

song = [v1,v2,v3,v2]

for verse in song:
    for note in verse:
        t.play(note)
        sleep(0.4)
        t.stop()
        sleep(0.1)
    sleep(0.2)
```

bennuttall:
We don't support I2C yet. If you can share some example code that would be helpful.

Issue #573, opened 2017-07-12
Colour Sensor


bsimmo:
I just had a look to see if you had a supported colour sensor and notice there isn't one.
Are there any plans to ?

The only one I've used is the TCS34725 as seen on the Enviro pHAT.

dglaude:
As far as I know, there is no I2C support in gpiozero.
So many sensor, that one included seems to be outside of the current scope.

The good thing about TCS34725 on Enviro pHAT or standalone (https://shop.pimoroni.com/products/adafruit-rgb-color-sensor-with-ir-filter) is that @pimoroni made a Python library.

I am stuck with this DF Robot sensor (https://shop.pimoroni.com/products/gesture-sensor) for witch I did not find python code, only Arduino library.

bennuttall:
As @dglaude said, we don't currently support I2C but we intend to add support in the next release (after the one that's due this month). I'm not sure what the support will be but it will at the very least provide fundamental support for the protocol, so hopefully you will be able to access the colour sensor in some way. Watch out for updates and you can help us test it!

ukBaz:
Can I gently nod towards https://github.com/intel-iot-devkit/mraa again?
It has support for I2C for Raspberry Pi and could be used as a pin factory.

bsimmo:
Excellent, it wasn't actually for me as such, I was replying to a question on the RPI forum and thought I would have a look it gpiozero supported it. 
A colour sensor wasn't there, thought best to ask in case something is there, not thought of, or just not in the current stable documents :-)


Aside/
I only have and know about the Enviro pHAT using it (and yes they have a simple library and I see Adafruit does too that can be used.) because of our local (uk) Raspi. JAM and connecting it the The Internet of Things LoRA gateway, other than reading about it in the Mag.
No idea if there are any other interfaces used by other colour sensors that may have been supported.

Quite happily play and test things out :-D

alexsavio:
I have been working with the Enviro+ and the particle matter sensor kit. I was a bit sad I could not use gpiozero for the setup, and also a bit sad about the state of the code quality of the tools I had to use instead. I wonder what lower level system support is missing to make that happen in gpiozero, is it I2C support? What are the blockers for that to happen? Maybe I can do something about it.

bennuttall:
Yeah, we don't have any I2C support at the moment. If you want to have a go at taking some base I2C code and making it more gpiozero-like, that would be handy.

PR #1070, opened 2023-03-08
Add I2C support and support for INA219 device (from pi-ina219)

TiborVoelcker:
Hi, all!

I needed I2C support for Remote GPIO so I added some bits and pieces to get it working. I tried to keep it as close to the SPI implementation as possible by pretty much copying everything.

This only includes a PiGPIOHardwareI2C implementation for now. I also did not add any additional documentation (except the copied docstrings).
For now I only added the `i2c_read_byte_data` , `i2c_read_word_data` , `i2c_write_byte_data` and `i2c_write_word_data` to read/write single bytes/words. This might not be the best solution, but worked for me.

I can work on this further for a bit, if this implementation is wanted. But I wanted to check first before I put more hours in.

As I was adding I2C to communicate with a INA219, I also added this as a new I2C device. This implementation is entirely stolen from [chrisb2's pi_ina219](https://github.com/chrisb2/pi_ina219). It has MIT license, but I'm not sure how to attribute him correctly or if I should add his MIT license somewhere?

### DIFF
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -72,6 +72,7 @@ from .spi_devices import (
 )
 from .i2c_devices import (
     I2CDevice,
+    INA219
 )
 from .output_devices import (
     OutputDevice,
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -70,6 +70,9 @@ from .spi_devices import (
     MCP3302,
     MCP3304,
 )
+from .i2c_devices import (
+    I2CDevice,
+)
 from .output_devices import (
     OutputDevice,
     DigitalOutputDevice,
--- a/gpiozero/i2c_devices.py
+++ b/gpiozero/i2c_devices.py
@@ -9,6 +9,8 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+from math import trunc
+import time
 from .devices import Device
 from .exc import DeviceClosed, GPIOZeroError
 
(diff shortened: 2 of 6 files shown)


TiborVoelcker:
I can also add a example file for using the INA219, which I used to read out my sensor over Remote GPIO.

Also, to add some links: If this implementation is wanted, this could help resolve #422, #573, #802, #808 and probably some others too.

Please let me know what kind of interface is wanted for reading the I2C. Is the read/write for byte/word enough? Maybe it should be also made clear which endianness is used.
Of course, I can also add documentation at the end.

lurch:
I've no idea if this is suitable (that's a @waveform80 decision), and I have no time to test any of it (I've never done any I2C stuff), but I'll leave some review-comments anyway in case they're useful :slightly_smiling_face:

gpiozero/__init__.py
```diff
@@ -70,6 +70,10 @@
     MCP3302,
     MCP3304,
 )
+from .i2c_devices import (
+    I2CDevice,
+    INA219
```
lurch, line 75: Should this mention of INA219 only be in the second commit and not the first one?

gpiozero/pins/pi.py
```diff
@@ -55,6 +57,34 @@ def spi_port_device(clock_pin, mosi_pin, miso_pin, select_pin):
     raise SPIBadArgs('invalid pin selection for hardware SPI')
 
 
+I2C_HARDWARE_PINS = {
```
lurch, line 60: Could maybe also include the additional I2C ports available on the Pi4? https://elinux.org/RPi_BCM2711_GPIOs
But I dunno how you'd want to deal with the way that some I2C ports are available on multiple pins :man_shrugging:

gpiozero/pins/pi.py
```diff
@@ -217,6 +247,98 @@ def _get_spi_class(self, shared, hardware):
...
+        defaults and converted into the pin format (from the port format) if
+        necessary.
+
+        Returns a tuple of ``(i2c_args, other_args)``.
+        """
+        try:
+            device = kwargs.pop("device")
+        except KeyError:
+            raise I2CBadArgs(
+                'device address is not specified as the *device* keyword'
```
lurch, line 288: Looks like this `device` kwarg isn't documented? Is it the hardware-ID of the particular device on the I2C bus?

How does your I2C implementation cope with multiple different devices (with different hardware addresses) being connected to the same I2C bus?

gpiozero/pins/pi.py
```diff
@@ -217,6 +247,98 @@ def _get_spi_class(self, shared, hardware):
...
+                    'port {i2c_args[port]} is not a valid I2C port'.format(
+                        i2c_args=i2c_args))
+            i2c_args = {
+                "data_pin": selected_hw["data"],
+                "clock_pin": selected_hw["clock"]
+                }
+        else:
+            raise I2CBadArgs(
+                'you must either specify port and device, or clock_pin, '
+                'mosi_pin, miso_pin, and select_pin; combinations of the two '
```
lurch, line 328: I guess the part here talking about mosi_pin and miso_pin etc. needs to be changed :wink:

gpiozero/i2c_devices.py
```diff
@@ -46,4 +46,346 @@ def __repr__(self):
...
+        self._min_device_current_lsb = self._calculate_min_current_lsb()
+        self._gain = None
+        self._auto_gain_enabled = False
+        self._voltage_range = None
+        self._current_lsb = None
+        self._power_lsb = None
+
+    def configure(self, voltage_range=RANGE_32V, gain=GAIN_AUTO,
+                  bus_adc=ADC_12BIT, shunt_adc=ADC_12BIT):
+        """Configure and calibrate how the INA219 will take measurements.
```
lurch, line 146: Feels like this should be part of the `__init__` method, rather than being a separate `configure` method? :thinking:

gpiozero/pins/pi.py
```diff
@@ -55,6 +57,34 @@ def spi_port_device(clock_pin, mosi_pin, miso_pin, select_pin):
...
+    :exc:`~gpiozero.I2CBadArgs` if the pins do not represent a valid hardware
+    I2C device.
+    """
+    for port, pins in I2C_HARDWARE_PINS.items():
+        if all((
+                data_pin  == pins['data'],
+                clock_pin  == pins['clock'],
+                )):
+            return port
+    raise I2CBadArgs('invalid pin selection for hardware SPI')
```
TiborVoelcker, line 85: Wrong error message here

gpiozero/__init__.py
TiborVoelcker: First of all, thanks for the review @lurch! Sorry I did not see it earler, I somehow missed the notification.
You are right, this was just a sloppy split of the commits. I will fix everything at once in a bit.

gpiozero/pins/pi.py
TiborVoelcker: I was only looking at https://de.pinout.xyz/pinout/i2c.
I will gladly add them, but I am a bit confused: In your source, there are SDA/SCL 1-6 defined, but 2 is missing?
Also, what are the different banks? SDA/SCL 0 is defined for every bank, do I just add all of them?
And lastly: e.g. SDA/SCL 6 is defined as the same as SCA/SCL 0. I can add this, no problem. But what is going on there?

gpiozero/pins/pi.py
TiborVoelcker: The `device` kwarg is the device address on the I2C bus. Completely forgot to add this in the docstring.

gpiozero/i2c_devices.py
TiborVoelcker: Like I said, this was completley stolen from [chrisb2's pi_ina219](https://github.com/chrisb2/pi_ina219).
I can add a call to `configure` to `__init__`, but it might be left out on purpose. If you want to use the device with the same settings as it already has (without knowing or caring what they are), you can use it without calling the `configure` method.

gpiozero/pins/pi.py
TiborVoelcker: Yup, sloppy copying 🙈

TiborVoelcker:
I also need help figuring out the Copyright notices and License issue. Regarding chris and possibly myself.

gpiozero/i2c_devices.py
lurch: Again this is a call for @bennuttall or @waveform80 to make, but IIRC GPIO Zero tries to be much "higher level" / "easier to use" than other typically lower-level libraries.

gpiozero/pins/pi.py
lurch: > I was only looking at https://de.pinout.xyz/pinout/i2c.

See https://github.com/pinout-xyz/Pinout.xyz/issues/322 :wink: 

You probably want to have a look at https://datasheets.raspberrypi.com/bcm2711/bcm2711-peripherals.pdf for more in-depth information. But to answer your specific questions...

> there are SDA/SCL 1-6 defined, but 2 is missing?

I2C-2 and I2C-7 are used by the HDMI ports, and aren't available on the GPIO pins.

> Also, what are the different banks?

On the Raspberry Pi 4 only the GPIOs in bank 0 are available on the GPIO header (see the column labelled "RPi4 connection" in that table). I _think_ some of the bank 1 GPIOs are available on the CM4, but I can't remember off the top of my head. Easiest option is to just ignore bank 1 for now.

> And lastly: e.g. SDA/SCL 6 is defined as the same as SCA/SCL 0. I can add this, no problem. But what is going on there?

GPIO0 when set to ALT-mode 0 gives you access to the SDA0 signal. GPIO0 when set to ALT-mode 5 gives you access to the SDA6 signal. Similarly, GPIO22 when set to ALT-mode 5 would also give you access to the SDA6 signal.
The different ALT-modes for each GPIO pin allow you to choose which signals you want mapped to each GPIO pin. See also some of the discussion in https://github.com/pinout-xyz/Pinout.xyz/issues/457

gpiozero/i2c_devices.py
```diff
@@ -0,0 +1,49 @@
...
+    specified with the constructor.
+    """
+
+    def __init__(self, **i2c_args):
+        self._i2c = None
+        super().__init__(pin_factory=i2c_args.pop("pin_factory", None))
+        self._i2c = self.pin_factory.i2c(**i2c_args)
+
+    def close(self):
+        if getattr(self, "_spi", None):
```
lurch, line 33: Another typo?

gpiozero/i2c_devices.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# Copyright (c) 2016-2021 Dave Jones <dave@waveform.org.uk>
+# Copyright (c) 2020 Grzegorz Szymaszek <gszymaszek@short.pl>
+# Copyright (c) 2016-2019 Andrew Scheller <github@loowis.durge.org>
+# Copyright (c) 2016-2018 Ben Nuttall <ben@bennuttall.com>
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+from math import trunc
+import time
```
lurch, line 13: Looks like these two imports aren't used?

gpiozero/i2c_devices.py
TiborVoelcker: Yup.

gpiozero/i2c_devices.py
TiborVoelcker: They should be used. `trunc` is used on line 303. `time` is used on line 235 and line 281.

TiborVoelcker: Ah. I know now what you meant. These imports are for the second commit. It's fixed now.

Issue #921, opened 2021-02-12
Why force active_state = False when pull_up = True?


pmlk:
My question is regarding the following couple of lines:

https://github.com/gpiozero/gpiozero/blob/a6f7efeb749045b4da7519e6abbc03309ee4a0f9/gpiozero/input_devices.py#L112-L124

I'd like to understand the reasoning behind this seemingly very deliberate design choice. From a few examples I remember seeing, it seems to force the following assumption (when `pull_up=True`): A button is connected to a pin and GND; when the button is pressed, the pin is electrically LOW (connected to GND) but logically the pin's state should be HIGH (or "active"), when the button is released the pin is pulled electrically HIGH (to avoid a floating pin), but should be logically LOW.

This makes sense. But why assume and force this scenario when `pull_up=True`? Why not allow `pull_up=True` and `active_state=True`?

Let me offer my perspective and my current use case:

I have a water reservoir with two sensors, one near the top to signal a "full" state and one near the bottom to signal an "empty" state. Basically, when there is water the sensors close a connection, when there is not enough water the connection is open.

I would like to connect the "Empty Sensor" to a pin and have the pin be logically HIGH when there is no more water. To me the most obvious way is to simply enable a pull up resistor for the pin and have the logical state be analogous to the electrical state. As long as there is water in the reservoir the sensor closes the connection between the pin and GND (electric == logic == LOW), once the reservoir is empty, the connection between the pin and GND is open, but the pin is pulled up to an electrical as well as logical HIGH.

Of course, there are a number of ways to get around this. I would just like to understand the reasoning behind this. 

Also, I'd be happy to open a PR to allow `pull_up=True and active_state=True`.

pmlk:
Also just noticed that passing `pull_up=True` and `active_state=False` throws a `PinInvalidState` Exception (`'Pin %d is not floating, but "active_state" is not None'`). To me this is even more astonishing since I am just explicitly passing what is otherwise automatically assumed anyways.

I would propose the following: 

1. Only when `active_state is None` should `active_state` be set automatically based on `pull_up`.
2. When both `active_state is None` and `pull_up is None` ("floating"), set `active_state=True` as a sensible default, rather than throwing an exception.

OR even more radical

3. decouple these two options entirely:
    * set sensible defaults (e.g. `pull_up=None`<sup>[1](#footnote1)</sup>, `active_state=True`)
    * never make an assumption about one option based on the other, assume the user knows what their doing

[...]

Issue #911, opened 2020-12-30
InputDevices Problem (Button on GPIO27)


Koloblix:
Operating system: Ubuntu Mate 20.10
Python version: e.g. 3.8.6
Pi model: Pi 4 Model B 4gb
GPIO Zero version: 1.5.1
Pin factory used: RPiGPIO

Using simple Python Script playing sounds on button pressed. Just works fine on all button, but on GPIO27.

at first my Script:

```python
from gpiozero import Button
import pygame.mixer
from pygame.mixer import Sound
from signal import pause

pygame.mixer.init()

button_sounds = {
    Button(4): Sound("/home/user/bla1.wav"),
    Button(17): Sound("/home/user/bla2.wav"),
    Button(22): Sound ("/home/user/bla3.wav"),
    Button(27): Sound ("/home/user/bla4.wav"),
}
for button, sound in button_sounds.items():
    button.when_pressed = sound.play
    print(button, "pressed")  

pause()
```
@print(button, "pressed")  it prints in shell:
```python
pygame 1.9.6
Hello from the pygame community. https://www.pygame.org/contribute.html
<gpiozero.Button object on pin GPIO4, pull_up=True, is_active=True> pressed
<gpiozero.Button object on pin GPIO17, pull_up=True, is_active=True> pressed
<gpiozero.Button object on pin GPIO22, pull_up=True, is_active=True> pressed
<gpiozero.Button object on pin GPIO27, pull_up=True, is_active=False> pressed
```
and on GPIO27 is_active=False 
Using "Button(27, active_state=True)

makes this error:
```python
gpiozero.exc.PinInvalidState: Pin 27 is not floating, but "active_state" is not None
```

bennuttall:
Did you set the pin state outside of Python?

I don't think you want to use `active_state`. I think you want `pull_up=bool` e.g `Button(27, pull_up=False)`

Koloblix:
No, havent set the pin state outside of python. Buttons are pulled down with 10k to GND, i dont need internal pull up/down.

Just wondering why GPIO27 is the only button, which dosnt work, and print says "is_active=FALSE"

lurch:
If you're using external pull-down resistors, you probably want e.g. `Button(27, pull_up=None, active_state=True)`

PR #1099, opened 2023-10-28
Support setting custom active_state for InputDevice

Makman2:
I hit the same limitation as described in #921. I use normally-closed switches for my experimental setup and needed to change the `active_state` parameter of `Button` to something else while using an internal pull-up or pull-down resistor to avoid soldering something and keep the experimental setup simple.

As [I commented in the issue](https://github.com/gpiozero/gpiozero/issues/921#issuecomment-1783804969), I went with following approach from the possibilities proposed by the original author of the issue @pmlk:

> Only when active_state is None should active_state be set automatically based on pull_up.

I left the other suggestion (which I think would be useful too), because it can be discussed separately and done in another PR:

> When both active_state is None and pull_up is None ("floating"), set active_state=True as a sensible default, rather than throwing an exception.

Please refer to my other reasonings in my [above mentioned comment](https://github.com/gpiozero/gpiozero/issues/921#issuecomment-1783804969). I am now lazy to copy them here :stuck_out_tongue: 

- [x] I read the [contributing guide](https://gpiozero.readthedocs.io/en/stable/contributing.html)
- [x] Unit tests added
- [x] Docs adjusted

Unit tests execute successfully skipping the real-pin tests.

I hope I have read everything correctly under the [contributing page](https://gpiozero.readthedocs.io/en/stable/contributing.html).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -204,6 +204,7 @@ See the `contributors page`_ on GitHub for more info.
 .. _Martchus: https://github.com/gpiozero/gpiozero/commits?author=Martchus
 .. _Martin O'Hanlon: https://github.com/martinohanlon/commits?author=martinohanlon
 .. _Mike Kazantsev: https://github.com/gpiozero/gpiozero/commits?author=mk-fg
+.. _Mischa Krüger: https://github.com/gpiozero/gpiozero/commits?author=Makman2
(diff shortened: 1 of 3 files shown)


fuslwusl:
Is there any progress with this topic? I have the same problem and need a way to handle normally-closed switches.

MauroPfister:
I also stumbled across #911 and would love to see this PR getting merged :smiley:

Issue #1111, opened 2024-01-13
Date deprectation warnings


MusicalNinjaRandInt:
14 Warnings
during test_internal_devices
from internal devices:595
only in 3.12

MusicalNinjaRandInt:
I intended to open this in my open fork as a reminder to fix it with a PR I'm working on. Sorry for using the wrong tab - need more coffee ...

MusicalNinjaRandInt:
OK - this challenge wouldn't leave me in peace so I've made an attempt to suggest how it could be managed - see #1116

PR #1112, opened 2024-01-13
Python 3.12 compatibility

MusicalNinjaRandInt:
### Dict interface to `importlib.metadata.entry_points()` fully deprecated

 - Created a small module `ep.py` which gets the entry point groups in a fashion compatible with both 3.9 and 3.12
- Added test coverage for `pin_class=[str]` when instatiating `MockFactory`
(NOTE: This was also broken in 3.9, have fixed that bug as well)

### Deprecation of `datetime.utcnow()`

- Filtered out deprecation warning for now
- Raised #1111 for "proper" handing in the future
(NOTE: `warnings.catch_warnings()` is not threadsafe but already in use in a few places, so I felt OK adding it here as well)

### Various

- Added python3.12 to github action
- updated the docs to reflect that fact that 3.9 is the lowest version currently supports (based on the `test.yml` and `setup.cfg`)
- provided a `python_version` to `conftest.py` which provides major.minor as a `float` for use in e.g. `xfail`, assuming this wil have repeated usefulness in the future

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,9 @@ jobs:
           - os: ubuntu-22.04
             python: "3.11"
             experimental: false
+          - os: ubuntu-22.04
+            python: "3.12"
+            experimental: false
 
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.experimental }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,11 +3,7 @@
 
 name: Python package
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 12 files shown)


MusicalNinjaRandInt:
I've deliberately not squashed this yet to assist with resolving review comments and ideas. I will squash when you're ready and also remove the github-codecov integration at that time.

bennuttall:
Thanks! We'll take a look when we get chance

MusicalNinjaRandInt:
Thanks. I don't have a physical test rig at the moment (gave my son my Pi3, which is how I found this awesome library) and these chances really need test_real_pins to be run on them. Not expecting any issues but that's when they usually occur ;)

MusicalNinjaRandInt:
I'm going to close this as it is effectively superceded by #1118, which resolves the `importlib.meradata` related changes and #1116 which does a better job of dealing with the `utcnow` deprecation

MusicalNinjaRandInt:
I'm going to close this as it is effectively superseded by https://github.com/gpiozero/gpiozero/pull/1118, which resolves the importlib.metadata related changes and https://github.com/gpiozero/gpiozero/pull/1116 which does a better job of dealing with the utcnow deprecation

PR #1116, opened 2024-01-30
Make TimeOfDay timezone aware - closes #1111

### LINKED ISSUE #1111: Date deprectation warnings

MusicalNinjaDad:
## Background

`utcnow()` is deprecated in python3.12 to reduce the danger of naive utc times being incorrectly interpreted as local times. [see the warning in the docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow)

The current implementation uses a flag, `utc`, in the instance to track whether the start and end times are naive utc or local time. This flag defaults to `True`.

This leads to deprecation warnings being raised which can only be dealt with in a non-thread-safe manner and collides with the future assumptions that users will make based upon the above deprecation.

## Semantics of the change

I have tried to preserve the following semantics to ensure not only tehcnical but also logical backwards compatibility:
- default handling is for times to be treated as UTC if nothing else is specified
- `utc=False` is used for specifying that times are "local time"
- specifying `utc=True` will force a naive utc, raise a custom deprecation warning to better explain the situation and allow the official warning to propogate (providing the user with full context and avoiding threading issues)
- where a timezone is given which is not a fixed offset (e.g.: a location which may observe daylight saving time) the current time in the location today, right now is used for comparison

## Additional points for discussion

1. It could be valuable to implement a new flag `local` (default: `False`) while still initially preserving the behaviour of `utc`. Given that `utc` is in future primarily used to flag local time `TimeOfDay(time(7), time(8), local=True)` would be more obvious in its intentions than `TimeOfDay(time(7), time(8), utc=False)`. This would also allow for the utc flag to be moved to a less prominent place in the documentation, the arguments order and possibly fully deprecated in v3.0.
1. When providing a `datetime` instance as a start or end time the date is ignored and only the time element stored. This could be confusing, now that more use is made of the specifics of the start and end time objects (by leveraging their `tzinfo`). Any changes to this behaviour would significantly change the semantics of the function. Perhaps there is a way to allow for the user to decide whether they really mean "at this one specific point in time", or "at this time every day"? It may be worth formally documenting the possibility to provide a datetime object and highlighting this behaviour. (The ability to pass a datetime object is currently undocumented but tested).

## Additional changes included
- adjustments to the validation process should now allow for substitution of datetime.datetime and datetime.time objects with others which provide the required interfaces but may not officially subclass the standard objects.
- adjustments the development environment to provide for virtual environments and running the tests on changes to every branch should contributors use these. (the commits from #1122)

## Changes originally included but now removed from this PR based on comments
- fixes to the implementation of entry_points to ensure python3.12 compatibility. This effectively removes the need for #1112. Given that the deprecation of `utcnow`occurs in 3.12 it makes sense to include these changes here - otherwise testing in 3.12 will cause multiple tests to fail.

Closes #1111

### DIFF
--- a/gpiozero/internal_devices.py
+++ b/gpiozero/internal_devices.py
@@ -13,7 +13,7 @@ import os
 import io
 import warnings
 import subprocess
-from datetime import datetime, time
+from datetime import datetime, time, timezone
 
 from .devices import Device
 from .mixins import EventsMixin, event
@@ -489,8 +489,11 @@ class TimeOfDay(PolledInternalDevice):
     *start_time* and *end_time* (inclusive) which are :class:`~datetime.time`
     instances.
 
+    Note that *start_time* may be greater than *end_time*, indicating a time
+    period which crosses midnight.
+
     The following example turns on a lamp attached to an :class:`Energenie`
-    plug between 07:00AM and 08:00AM::
+    plug between 07:00AM and 08:00AM UTC::
 
         from gpiozero import TimeOfDay, Energenie
         from datetime import time
@@ -504,8 +507,35 @@ class TimeOfDay(PolledInternalDevice):
 
         pause()
 
-    Note that *start_time* may be greater than *end_time*, indicating a time
-    period which crosses midnight.
+    By default start and end times are timezone-aware UTC times. If you wish to
+    specify the time-zone for the start and/or end time you can do so when 
+    contructing the time, for example to switch on when it is office hours in 
+    both London and Los Angeles::
+
+        from gpiozero import TimeOfDay,
+        from datetime import time,
+        from zoneinfo import ZoneInfo
+
+        tz_LA = ZoneInfo('America/Los_Angeles')
+        tz_London = ZoneInfo('Europe/London')
+
+        officehours = TimeOfDay(time(8,30,tzinfo=tz_LA), time(18,00,tzinfo=tz_London))
+
+    If you would like to ignore timezones and use "local time" (whatever time
+    your Pi's internal clock says) then set `utc` to `False`. To
+    switch on during whatever your Pi thinks are local office hours::
+        
+        from gpiozero import TimeOfDay,
+        from datetime import time
+
+        officehours = TimeOfDay(time(8,30), time(18,00), utc=False)
+
+    .. note::
+        For backwards compatibility you can also select to use naive UTC times by
+        setting `utc` to `True` - this is no longer recommended,
+        instead you should leave `utc` set to `None` or explicity
+        specify `tzinfo=UTC` (`from datetime import UTC`) - both will give the
+        same result. 
 
     :param ~datetime.time start_time:
         The time from which the device will be considered active.
@@ -514,51 +544,97 @@ class TimeOfDay(PolledInternalDevice):
         The time after which the device will be considered inactive.
 
     :param bool utc:
-        If :data:`True` (the default), a naive UTC time will be used for the
-        comparison rather than a local time-zone reading.
+        If `None` (the default), UTC time will be used for the comparison. 
+        If `False` the local clock time will be use, ignoring the timezone. 
+        (If `True` a naive UTC time will be used - this is not recommended, 
+        see the note above)
 
     :type event_delay: float
     :param event_delay:
-        The number of seconds between file reads (defaults to 10 seconds).
+        The number of seconds between file reads (defaults to 5 seconds).
 
     :type pin_factory: Factory or None
     :param pin_factory:
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
     """
-    def __init__(self, start_time, end_time, *, utc=True, event_delay=5.0,
+
+    def __init__(self, start_time, end_time, *, utc=None, event_delay=5.0,
                  pin_factory=None):
-        self._start_time = None
-        self._end_time = None
-        self._utc = True
+    
+        utc_deprecation_warning = (
+        'Using utc=True is deprecated and scheduled for removal in a future version.'
+        ' Please use tz="UTC" or provide a Timezone-Aware start and end time'
+        )
+        if utc:
+            warnings.warn(utc_deprecation_warning, DeprecationWarning)
+
+        self._aware = utc is None
+        self._utc = utc
+
         super().__init__(event_delay=event_delay, pin_factory=pin_factory)
         try:
             self._start_time = self._validate_time(start_time)
             self._end_time = self._validate_time(end_time)
             if self.start_time == self.end_time:
                 raise ValueError('end_time cannot equal start_time')
-            self._utc = utc
             self._fire_events(self.pin_factory.ticks(), self.is_active)
         except:
             self.close()
             raise
 
     def __repr__(self):
+        reprname = f'gpiozero.{self.__class__.__name__} object'
+        if self.aware:
+            reprstart = f'{self.start_time.replace(tzinfo=None)} [{self.start_time.tzinfo}]'
+            reprend = f'{self.end_time.replace(tzinfo=None)} [{self.end_time.tzinfo}]'
+            reprtz = ''
+        else:
+            reprstart = f'{self.start_time}'
+            reprend = f'{self.end_time}'
+            reprtz = f'{(" local", " UTC")[self.utc]}'
         try:
             self._check_open()
-            return (
-                f'<gpiozero.{self.__class__.__name__} object active between '
-                f'{self.start_time} and {self.end_time} '
-                f'{("local", "UTC")[self.utc]}>')
+            return f'<{reprname} active between {reprstart} and {reprend}{reprtz}>'
         except DeviceClosed:
             return super().__repr__()
 
     def _validate_time(self, value):
-        if isinstance(value, datetime):
-            value = value.time()
-        if not isinstance(value, time):
-            raise ValueError(
+        # If we have a datetime or similar we only want the time.
+        # hasattr is faster than try-except if we usually expect try to fail - and
+        # we'll probably be getting a time more often than a datetime
+        if hasattr(value, 'timetz'): 
+            value = value.timetz()
+        
+        if not self.aware:
+            try:
+                assert value.tzinfo == None
+            except AttributeError:
+                True # Want to include this branch in coverage report
+                pass # pass is excluded from coverage in setup.cfg 
+            except AssertionError:
+                raise ValueError(
+                'utc must be None if start_time or end_time contain tzinfo')
+
+        # Using try-except to cope with cases where someone has used an object
+        # that offers comparison with time but is not a subclass of time.
+        # Not relying on time's current implementation that checks for timetuple()
+        # as this may change in the future
+        if self.aware:            
+            try: # We need to be able to replace tzinfo and compare to aware time
+                value.replace(tzinfo=timezone.utc) < time(1, tzinfo=timezone.utc)
+            except (AttributeError, TypeError):
+                raise ValueError(
                 'start_time and end_time must be a datetime, or time instance')
+        else:
+            try: # We need to be able to compare to naive time
+                value < time(1)
+            except TypeError:
+                raise ValueError(
+                'start_time and end_time must be a datetime, or time instance')
+            
+        if self.aware and value.tzinfo == None: # Default to UTC
+            value = value.replace(tzinfo=timezone.utc)
         return value
 
     @property
@@ -578,10 +654,19 @@ class TimeOfDay(PolledInternalDevice):
     @property
     def utc(self):
         """
-        If :data:`True`, use a naive UTC time reading for comparison instead of
-        a local timezone reading.
+        If `None` (the default), UTC time will be used for the comparison. 
+        If `False` the local clock time will be use, ignoring the timezone. 
+        (If `True` a naive UTC time will be used - this is not recommended, 
+        see the note above)
         """
         return self._utc
+    
+    @property
+    def aware(self):
+        """
+        Whether the comparison will be performed in a timezone-aware manner
+        """
+        return self._aware
 
     @property
     def value(self):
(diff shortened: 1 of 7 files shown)


gpiozero/internal_devices.py
```diff
@@ -504,8 +507,35 @@ class TimeOfDay(PolledInternalDevice):
...
+    both London and Los Angeles::
+
+        from gpiozero import TimeOfDay,
+        from datetime import time,
+        from zoneinfo import ZoneInfo
+
+        tz_LA = ZoneInfo('America/Los_Angeles')
+        tz_London = ZoneInfo('Europe/London')
+
+        officehours = TimeOfDay(time(8,30,tzinfo=tz_LA), time(18,00,tzinfo=tz_London))
```
lurch, line 522: I think this example of using different timezones for the start-time and end-time is a bit confusing?

gpiozero/internal_devices.py
```diff
@@ -514,51 +544,97 @@ class TimeOfDay(PolledInternalDevice):
...
+        The number of seconds between file reads (defaults to 5 seconds).
 
     :type pin_factory: Factory or None
     :param pin_factory:
         See :doc:`api_pins` for more information (this is an advanced feature
         which most users can ignore).
     """
-    def __init__(self, start_time, end_time, *, utc=True, event_delay=5.0,
+
+    def __init__(self, start_time, end_time, *, utc=None, event_delay=5.0,
```
lurch, line 562: What backwards-compatibility implications does changing the default-value for `utc` have?

gpiozero/internal_devices.py
```diff
@@ -514,51 +544,97 @@ class TimeOfDay(PolledInternalDevice):
...
                  pin_factory=None):
-        self._start_time = None
-        self._end_time = None
-        self._utc = True
+    
+        utc_deprecation_warning = (
+        'Using utc=True is deprecated and scheduled for removal in a future version.'
+        ' Please use tz="UTC" or provide a Timezone-Aware start and end time'
```
lurch, line 567: Lowercase 'a' on 'Aware'? :wink:

gpiozero/internal_devices.py
```diff
@@ -578,10 +654,19 @@ def end_time(self):
...
-        a local timezone reading.
+        If `None` (the default), UTC time will be used for the comparison. 
+        If `False` the local clock time will be use, ignoring the timezone. 
+        (If `True` a naive UTC time will be used - this is not recommended, 
+        see the note above)
         """
         return self._utc
+    
+    @property
+    def aware(self):
```
lurch, line 665: I think `aware` by itself is too ambiguous, perhaps `timezone_aware` would be better?

gpiozero/internal_devices.py
```diff
@@ -514,51 +544,97 @@ class TimeOfDay(PolledInternalDevice):
...
-        comparison rather than a local time-zone reading.
+        If `None` (the default), UTC time will be used for the comparison. 
+        If `False` the local clock time will be use, ignoring the timezone. 
+        (If `True` a naive UTC time will be used - this is not recommended, 
+        see the note above)
 
     :type event_delay: float
     :param event_delay:
-        The number of seconds between file reads (defaults to 10 seconds).
+        The number of seconds between file reads (defaults to 5 seconds).
```
lurch, line 554: I guess the "file reads" here doesn't make sense? (looks like a copy-paste typo?)

gpiozero/internal_devices.py
```diff
@@ -504,8 +507,35 @@ class TimeOfDay(PolledInternalDevice):
...
+        from datetime import time,
+        from zoneinfo import ZoneInfo
+
+        tz_LA = ZoneInfo('America/Los_Angeles')
+        tz_London = ZoneInfo('Europe/London')
+
+        officehours = TimeOfDay(time(8,30,tzinfo=tz_LA), time(18,00,tzinfo=tz_London))
+
+    If you would like to ignore timezones and use "local time" (whatever time
+    your Pi's internal clock says) then set `utc` to `False`. To
```
lurch, line 525: Strictly speaking I believe the "internal clock" is always UTC, so perhaps change it to just "Pi's clock" ?

lurch:
There seem to be a bunch of code-change overlaps with your entrypoint PR - perhaps those changes ought to be removed from _this_ PR?

gpiozero/internal_devices.py
MusicalNinjaRandInt: The general concept or the way the code is written? I felt the example was quite powerful, but maybe the initial paragraph above needs some work?
Would it be more readable as:

```python
tz_LA = ZoneInfo('America/Los_Angeles')
tz_London = ZoneInfo('Europe/London')

LA_opens = time(8,30,tzinfo=tz_LA)
London_closes = time(18,00,tzinfo=tz_London)

officehours = TimeOfDay(LA_opens, London_closes)
```

gpiozero/internal_devices.py
MusicalNinjaRandInt: Sematically, the behaviour is still the same: default behaviour is for the times to be treated as UTC. Technically, those will now be timezone-aware UTC times rather than naive ones.
TimeOfDay will still act exactly as before if the default is used, `True` is passed or `False` is passed.

The only risk I see is if someone has made some kind of use of the fact that previously the times were naive *and* had relied on implicity specifying `utc` rather than explicitly stating `utc=True`. I can't thin of any cases where someone would be doing something with the times which needed them to be naive, other than working in "local time".

Would it be worthwhile rephrasing `self._aware = utc is None` to something which takes a little more space and makes the behaviour more apparent. At the moment most of the relevant logic is later in `_validate_time()`

gpiozero/internal_devices.py
MusicalNinjaRandInt: We'd need to ask @bennuttall - it looks like he wrote that in #941

MusicalNinjaRandInt:
> There seem to be a bunch of code-change overlaps with your entrypoint PR - perhaps those changes ought to be removed from _this_ PR?

Yes - That makes sense. I had built this on top of the python3.12 relevant bits, before beginning working on 3.7 compatibility.

I'll take them out of here

MusicalNinjaRandInt:
As a general question @lurch - would you prefer individual commits to resolve each comment and then a re-squash once you're happy with the whole thing, or some other approach?

gpiozero/internal_devices.py
lurch: I think the general concept of using different timezones for the start-time and end-time is a bit confusing.

I think your example assumes that the reader knows that the LA timezone is "negative" compared to the London timezone? And does it assume that LA doesn't close before London opens?

gpiozero/internal_devices.py
lurch: I think I'd need to do more reading up to understand the difference between "timezone-aware UTC times" and "naive UTC times" to understand what you're saying here; which I'm afraid I don't have time to do right now :wink:

lurch:
> would you prefer individual commits to resolve each comment and then a re-squash once you're happy with the whole thing, or some other approach?

That's a question for Ben and Dave, the maintainers of the repo. I just try to help out by occasionally leaving (hopefully) helpful comments here and there :laughing:

gpiozero/internal_devices.py
MusicalNinjaRandInt: @bennuttall , @waveform80 - the more I think about this; the more I think it would make sense to take the deprecation a step further and introduce a new boolean keyword argument `local` in addition to the changes to `utc`
I would move `utc` to the end of the signature and documentation, placing `local` more prominently.

This would make the normal usage:
1. use UTC times (in a timezone-aware fashion) = default
2. specify `tzinfo` in the construction of the start & end times to use a specific timezone
3. specify `local=True` to use local time

While still maintaining backwards compatibility, by keeping `utc` semantically identical to now.

@lurch  - I'd also be interested in your views on this if you have any; your comments have been really useful in cleaning up rough edges.

gpiozero/internal_devices.py
```diff
@@ -514,51 +544,98 @@ class TimeOfDay(PolledInternalDevice):
...
+    
+        utc_deprecation_warning = (
+        'Using utc=True is deprecated and scheduled for removal in a future version.'
+        ' Start and end times will default to tzinfo=datetime.UTC unless you specify'
+        ' different tzinfo when defining them.'
+        )
+        if utc:
+            warnings.warn(utc_deprecation_warning, DeprecationWarning)
+
+        self._aware = utc is None
```
lurch, line 573: _Perhaps_ we don't need a separate `self._aware` attribute, and the `timezone_aware` property could just `return self._utc is None` ?

lurch:
Perhaps it's also worth mentioning in the docstrings that if this is given a full `datetime` object, rather than just a `time` object, the "date" part of the datetime will be silently discarded, and only the "time" part will be used?

gpiozero/internal_devices.py
MusicalNinjaRandInt: I had thought about that, and chosen to leave this in __init__ as I feel it is a lot more explicit than having the logic later in the code as part of the property. As a bonus (but possible YAGNI) this is also where any logic around a `local` argument would be placed, and used to define self._aware.
I'll rename this to _tzaware to be a bit more explicit.

MusicalNinjaRandInt:
> Perhaps it's also worth mentioning in the docstrings that if this is given a full `datetime` object, rather than just a `time` object, the "date" part of the datetime will be silently discarded, and only the "time" part will be used?

Yes - that's one for Ben & Dave as this is original functionality (see "Additional points for discussion - point 2) in the PR description for my view...

bennuttall:
## Review

### Overview
Reworks `TimeOfDay` to be timezone-aware, replacing the deprecated `datetime.utcnow()` with `datetime.now(tz=...)`. Preserves the old `utc=True`/`utc=False` API for backward compatibility (with a deprecation warning on `utc=True`), and adds a new default mode (`utc=None`) where start/end times carry their own `tzinfo` (defaulting to UTC if unspecified). Also bundles unrelated dev-environment changes: a devcontainer, CI trigger/coverage tweaks, and two shebang fixes. +352/−64 across 8 files.

I checked this out locally and ran the full test suite — **386 passed**, including a genuinely thorough set of new DST-crossing test cases (LA/London, DST vs non-DST). The core datetime logic is sound: exploiting Python's own aware-`time` comparison semantics (which normalize by UTC offset) to keep the "start > end wraps midnight" logic uniform between naive and aware paths is a neat piece of design, and it's well covered by tests.

### Issues found

**1. Blocking: a broken git submodule reference committed by accident**
The diff adds `.devcontainer` as a gitlink (`new file mode 160000`, `Subproject commit 771535c1...`), but there's no accompanying `.gitmodules` file. I checked out the branch locally and confirmed: `.devcontainer/` ends up as an **empty directory** with no way to resolve the reference. This is almost certainly an artifact of the contributor having a nested `.git` inside their local `.devcontainer` (e.g. from a devcontainer template) that got picked up by `git add` as a submodule instead of plain files. This needs to be fixed before merge — remove the nested `.git` and re-add `.devcontainer`'s actual contents as regular tracked files, or drop it from this PR.

**2. Documentation contains invalid Python syntax**
Both new code examples in the `TimeOfDay` docstring have trailing commas on `from` imports with nothing following, e.g.:
```python
from gpiozero import TimeOfDay,
from datetime import time,
```
I verified with `ast.parse()` — both raise `SyntaxError: trailing comma not allowed without surrounding parentheses`. These are copy-pasteable examples in rendered docs; as written they'd fail for any user who tries them.

**3. Docstring recommends `datetime.UTC`, which doesn't exist on the package's minimum supported Python**
The note says: "specify `tzinfo=UTC` (`from datetime import UTC`)". `datetime.UTC` was only added in **Python 3.11**; `setup.cfg` declares `python_requires = >=3.9`. A user on 3.9/3.10 following this advice gets an `ImportError`. The code itself correctly uses `timezone.utc` everywhere (3.2+ compatible) — it's only the doc suggestion that's version-inconsistent. Should recommend `from datetime import timezone` / `timezone.utc` instead, or gate the `UTC` mention as 3.11+-only.

**4. Odd validation code / stray debug artifact**
In `_validate_time`:
```python
try:
    assert value.tzinfo == None
except AttributeError:
    True # Want to include this branch in coverage report
    pass # pass is excluded from coverage in setup.cfg
except AssertionError:
    raise ValueError(...)
```
The bare `True` expression statement is a workaround for `pass` being in the project's `coverage:report` `exclude_lines`, added purely so this branch registers as covered. It works, but reads like leftover debugging cruft and will likely puzzle the next person to touch this code. Worth a comment explaining *why* it's there (which it partially has) or restructuring to avoid needing the trick — e.g. `if not hasattr(value, 'tzinfo'): pass` instead of assert/except-for-control-flow.

Similarly, the aware-path interface check (`value.replace(tzinfo=timezone.utc) < time(1, tzinfo=timezone.utc)`) is a fairly indirect way to validate "does this look like a time/datetime" — functionally fine and tested, but non-obvious; a short comment on *why* this particular expression was chosen would help maintainability.

### Minor / style
- Docstring for `utc` property/param still says `:param bool utc:` — the real type is now `Optional[bool]` (`None`/`True`/`False` are all meaningful).
- The deprecation warning text ("Start and end times will default to tzinfo=datetime.UTC unless you specify...") reads like it's describing current behavior but is actually advice about the *recommended alternative* — a little confusing on first read; could be reworded to something like "Instead, leave `utc` unset — start and end times will then default to UTC unless you give them their own `tzinfo`."
- Several new docstring lines have trailing whitespace, and "contructing" → "constructing" typo.
- No changelog entry, though that may just follow this project's per-release-batch convention.
- `.github/workflows/test.yml` changes CI to run on all branches/PRs (previously `master`-only) and adds `--cov` + Codecov upload (`codecov/codecov-action@v3`, which is a couple of major versions behind current). Reasonable rationale per the PR description, but broadens CI minutes usage — worth a conscious yes/no from a maintainer rather than being bundled silently into a timezone PR.
- `setup.cfg` comments out `addopts = -rsx --cov --tb=short` for local runs (coverage interferes with `pdb`) while compensating with `--cov` in CI — reasonable, but changes default local test behavior for all contributors; worth calling out explicitly in the PR description.

### Test coverage
Excellent — parametrized tests cover naive-local, naive-UTC (deprecated), tz-aware with explicit `tzinfo`, mismatched `utc=`/`tzinfo=` combinations (correctly raising), fixed offset vs `ZoneInfo`, and DST transitions across two different real-world zones. This is noticeably more thorough than most PRs in this repo.

### Security
None relevant.

### Recommendation
The date/time logic itself is solid and well-tested — this is good, careful work. But it isn't mergeable as-is: the broken `.devcontainer` submodule reference must be removed, and the docstring's invalid syntax / wrong `datetime.UTC` availability claim should be fixed since they'll actively mislead users. The stray `True` coverage-trick line is a nice-to-fix but not blocking.

---

bennuttall:
Thanks for the contribution. Continuing this at #1228

MusicalNinjaDad:
> Thanks for the contribution. Continuing this at #1228

Thanks @bennuttall - sorry for the extra mess in this one, I was really struggling to get a sandboxed setup which worked for gpiozero and didn't pollute any PRs.

PR #1118, opened 2024-02-02
ensure python3.7 compatibility - closes #1113

### LINKED ISSUE #1113: 3rd party import-metadata not supported?
Operating system: Raspbian Buster (Volumio OS)
Python version: 3.7
Pi model:  Pi 5
GPIO Zero version: 2.0
Pin factory used: RPiGPIO

In a python3 venv, I used pip3 to install gpiozero and import-metadata. However when trying to run a very simple button press detection script, I get the following error:

```
  File "button-check.py", line 1, in <module>
    from gpiozero import Button
  File "/home/volumio/deadman-env/lib/python3.7/site-packages/gpiozero/__init__.py", line 19, in <module>
    from .pins import (
  File "/home/volumio/deadman-env/lib/python3.7/site-packages/gpiozero/pins/__init__.py", line 21, in <module>
    from ..devices import Device
  File "/home/volumio/deadman-env/lib/python3.7/site-packages/gpiozero/devices.py", line 19, in <module>
    from importlib.metadata import entry_points
ModuleNotFoundError: No module named 'importlib.metadata'
```
My understanding based on the documentation is that Python 3.5 and newer is supported by gpiozero 2.0, but this error seems to imply that importing libraries is being accomplished with Python 3.8's built-in import-metadata rather than the 3rd part external one. And since I'm using Python 3.7, that wouldn't work. So I guess either something wrong is happening or the documentation is misleading?

MusicalNinjaRandInt:
- compatibility restored by using `importlib` backports as needed
- aim to ensure compatibility is maintained by running tests against 3.7 in pipeline
- documentation updated to reflect the policy of supporting oldoldstable

Note: supporting 3.6 would require additional backports to be used; supporting 3.5 is not feasible (pip doesn't support it anymore and f-strings would break)

I've moved the handling of  `entry_points` to a dedicated module as the specifics are different for each version. The approach implemented also provides forward compatibility with 3.12

Sphinx 4.0 is not available for python3.7 - so I had to change the pipeline to only install test dependencies, not `make develop` - this is also faster anyway

I put dev containers in place so I could code and verify specifically on buster, bullseye and bookworm. I've left them (with a note in the Developing chapter docs) as they could be useful to others and don't harm.

This closes #1113

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,86 @@
+ARG DEBIAN=debian
+FROM debian:${DEBIAN}
+ARG PYVERSION=3
+
+# ---
+# Create the default user - vscode mounts workspace directory chowned to 1000:1000
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+RUN groupadd --gid $USER_GID $USERNAME \
+&& useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
+&& apt-get update \
+&& apt-get install -y sudo \
+&& echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
+&& chmod 0440 /etc/sudoers.d/$USERNAME
+
+# ---
+# Install basic dev tools
+RUN apt update \
+&& apt install -y --no-install-recommends \
+    ca-certificates \
+    git \
+    locales \
+    locales-all \
+    make \
+&& rm -rf /var/lib/apt/lists/*
+
+# ---
+# Install python
+RUN if [ $PYVERSION -eq 3 ] \
+    ; then \
+        apt update \
+        && apt install -y --no-install-recommends \
+            python3 \
+            python3-distutils \
+            python3-pip \
+            python3-venv \
+        && rm -rf /var/lib/apt/lists/* \
+        && ln -s /usr/bin/python3 /usr/bin/py \
+    ; else \
+        echo "Install from source" \
+        && apt update \
+        && apt install -y --no-install-recommends \
+            build-essential \
+            gdb \
+            lcov \
+            libbz2-dev \
+            libffi-dev \
+            libgdbm-dev \
+            libgdbm-compat-dev \
+            liblzma-dev \
+            libncurses5-dev \
+            libreadline6-dev \
+            libsqlite3-dev \
+            libssl-dev \
+            lzma lzma-dev \
+            pkg-config \
+            tk-dev \
+            uuid-dev \
+            zlib1g-dev \
+        && rm -rf /var/lib/apt/lists/* \
+        && cd /opt \
+        && git clone --branch $PYVERSION --single-branch https://github.com/python/cpython \
+        && cd cpython \
+        && ./configure --prefix=/usr\
+        && make -j \
+        && make altinstall \
+        && ln -s /usr/bin/python$PYVERSION /usr/bin/py \
+    ; fi
+
+
+# ---
+# Install tools needed for building docs
+RUN apt update \
+&& apt install -y --no-install-recommends \
+    graphviz \
+    inkscape \
+    latexmk \
+    texlive-fonts-recommended \
+    texlive-latex-recommended \
+    texlive-latex-extra \
+    texlive-xetex \
+    xindy \
+&& rm -rf /var/lib/apt/lists/*
+
+USER $USERNAME
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -4,13 +4,13 @@ ARG PYVERSION=3
 
 # ---
 # Create the default user - vscode mounts workspace directory chowned to 1000:1000
-ARG USERNAME=vscode
+ARG USERNAME=gpiozero
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 RUN groupadd --gid $USER_GID $USERNAME \
 && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
-&& apt-get update \
-&& apt-get install -y sudo \
+&& apt update \
+&& apt install -y sudo \
 && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
 && chmod 0440 /etc/sudoers.d/$USERNAME
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,12 @@ jobs:
       matrix:
         include:
           - os: ubuntu-20.04
+            python: "3.7"
+            experimental: false
+          - os: ubuntu-20.04
+            python: "3.8"
+            experimental: false
+          - os: ubuntu-20.04
             python: "3.9"
             experimental: false
           - os: ubuntu-20.04
@@ -20,6 +26,10 @@ jobs:
           - os: ubuntu-22.04
             python: "3.11"
             experimental: false
+          - os: ubuntu-22.04
+            python: "3.12"
+            experimental: false
+
 
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.experimental }}
@@ -35,7 +45,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        make develop
+        pip install -e .[test]
 
     - name: Run tests
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,11 +3,7 @@
 
 name: Python package
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
 
 jobs:
   test:
@@ -43,4 +39,7 @@ jobs:
 
     - name: Run tests
       run: |
-        pytest -v -k 'not test_real_pins.py'
+        pytest -v -k 'not test_real_pins.py' --cov
+
+    - name: Upload coverage reports to Codecov
+      uses: codecov/codecov-action@v3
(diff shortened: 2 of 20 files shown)


docs/compat.rst
```diff
@@ -11,6 +11,13 @@ Backwards Compatibility
 
 .. currentmodule:: gpiozero
 
+We aim to ensure that gpiozero will run on all versions of python included in releases
```
lurch, line 14: I wonder if it might make sense to split this Docker / devcontainer stuff out into a separate PR, so that it can be reviewed and merged independently of this one?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,86 @@
+ARG DEBIAN=debian
```
lurch, line 1: I wonder if it might make sense to split this Docker / devcontainer stuff out into a separate PR, so that it can be reviewed and merged independently of this one?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+ARG PYVERSION=3
+
+# ---
+# Create the default user - vscode mounts workspace directory chowned to 1000:1000
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+RUN groupadd --gid $USER_GID $USERNAME \
+&& useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
+&& apt-get update \
```
lurch, line 12: Why the mixture of `apt-get` here and `apt` below?

.github/workflows/test.yml
```diff
@@ -3,18 +3,20 @@
 
 name: Python package
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
```
lurch, line 6: I dunno much about Github workflows, but does this change mean that any **untrusted code** that anybody pushes in a PR will get run automatically by GitHub CI? (and is that potentially problematic?)

gpiozero/boards.py
```diff
@@ -949,10 +949,20 @@ def __init__(self, *pins, dp=None, font=None, pwm=False, active_high=True,
 
         if font is None:
             if len(pins) == 7:
-                with resources.files('gpiozero.fonts').joinpath('7seg.txt').open() as f:
+                try:
+                    fontpath = resources.files('gpiozero.fonts').joinpath('7seg.txt')
```
lurch, line 953: Seems to be a lot of repetition here (between 7seg and 14seg), so I wonder if it might be worth abstracting it out into a separate helper function?

gpiozero/ep.py
```diff
@@ -0,0 +1,25 @@
+"""
+Provides access to the gpiozero entry points:
+
+.. code-block:: python
+    from gpiozero.ep import PinFactory_entry_points
```
lurch, line 5: Do we want a more meaningful name than `gpiozero.ep` ? :shrug:

.devcontainer/Dockerfile
MusicalNinjaRandInt: Sure - that would also give a chance to consider prebuilding the docker images, providing them from ghcr.io and adding some more info in the docs. It's only valuable to those who want to develop or extend gpiozero ...
I was using it to check compatibility with each specific install but none of the below comments look like there is much risk. And I can remove it as a last step to finishing this PR anyway...

.devcontainer/Dockerfile
MusicalNinjaRandInt: No good reason (the bad one is "done at different times"). I'll change for consistency

.github/workflows/test.yml
MusicalNinjaRandInt: The only time this would make a difference in your project is if one of the contributors is developing on a branch in gpiozero/gpiozero - then the workflow would trigger on each push - probably a good thing.

It already triggers on PRs to master in the original version.

The advantage is that other developers working on a fork have the tests run in pipeline on each push to each branch - which is good practice. Leaving this out means anyone else who wants to make use of the workflow on a fork needs to make a similar change manually and then think to remove it before raising the PR.

gpiozero/boards.py
MusicalNinjaRandInt: Yes, good idea.
I'd suggest I move it up to the imports at the top, to avoid having `import` hidden in the code and use:

```python
from importlib import resources
try:
    files = resources.files
except AttributeError:
    import importlib_resources
    files = importlib_resources.files
```

That would probably be the best compromise between not importing in the middle of code and not running code during the imports.(?)

gpiozero/ep.py
MusicalNinjaRandInt: Any ideas what? I leant on the approach used for `exc` and tried to avoid anything that would collide with `entry_points`

Would it make sense to expose `PinFactory_entry_points` and `MockPinClass_entry_points` via `__init__.py`? In that case I would also add them to the docs; the question is "where?": a separate "Entry Points API" would be very short ...

lurch: Again, a question for Ben & Dave...

gpiozero/boards.py
lurch: The only caveat I'd add is that `files` is perhaps too generic of a name to exist at global scope - maybe `resource_files` ?

.github/workflows/test.yml
lurch: Apologies if I was being too cautious.

gpiozero/boards.py
MusicalNinjaRandInt: I think I found the answer: positioning just under the imports, named `fontslocation`and pointing to the path where the fonts reside...

lurch: Looks a lot neater :+1:

.devcontainer/Dockerfile
MusicalNinjaRandInt: @bennuttall , @waveform80: What is your view on this?
IMO the ability to code, debug and test in the specific versions is part of supporting them - and I can't think of a better way of enabling that than with devcontainers for oldoldstable through to stable. Having the github workflow run on all supported python versions is good practice to help ensure that no future changes break backward compatibility, but doesn't necessarily help enough when fixing any issues which it throws up...

If you leave this in would you like more docs and prebuilt docker images as well?

gpiozero/ep.py
MusicalNinjaRandInt: @bennuttall , @waveform80 : Would you like:

- anything exposing via __init__.py to make it easier for others to use as well
- some documentation on this topic (in which case where?)
- improved naming for the module?

waveform80:
Thank you for the PR -- I'm going to echo several of @lurch's comments because I'm a bit pressed for time tonight, and I'm afraid ultimately I'm going to go in a rather different direction than this is proposing, but I do appreciate the effort to try and support as wide a range of Python versions as possible.

Ultimately I think it's fairly reasonable to drop 3.7 (and 3.8) support at this point in time. The current Debian+RaspiOS release (bookworm) is on 3.11, Debian oldstable (bullseye) is 3.9, it's only Debian oldoldstable (buster) that's on 3.7 and I don't think there's any support for that on the RaspiOS side at this point. On the Ubuntu side, the current LTS (jammy) is 3.10, the prior LTS (focal) is 3.8 but [deadsnakes PPA](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) permits those users to have any up-to-date Python installation they want.

This was the reasoning behind my (ahem ... attempted but misspelled) limit on 3.9 as a [base version](https://github.com/gpiozero/gpiozero/blob/5dc786c2128f84524d5b574583e06602e8f6b387/setup.cfg#L31) (and incidentally, thanks for bringing that to my attention!).

On this PR specifically:

* As @lurch suggests, the Docker stuff should be split into its own PR. I've barely touched Docker, so I can't judge whether any of this should go in. I'm not against it, but I'm not going to blindly include it as part of another PR (and unfortunately it's mixed into various commits here so I can't easily cherry-pick it out).
* Rather than include 3.7/3.8 as supported versions, I'd rather ditch those versions and firmly bump everything (`setup.cfg`, `tox.ini`, etc.) to 3.9 as a base. This removes the need for basically all the changes in `boards.py`.
* On fixing 3.9's use of `entry_points`, I'd rather rely on the `importlib_metadata` backport; i.e. beyond an initial `try..except ImportError` to pull in the relevant `entry_points` function from either the backport (on 3.9) or the stdlib (on later versions), the body of the code in the module should just call `entry_points` as it would be in later versions (3.10 onwards).

Normally, I'd wait to see if the author wanted to pursue those changes but unfortunately I'm under some time pressure to get GPIO Zero compatible with 3.12 (it's holding up proposed migration in Ubuntu, which is my day job) so in this instance I'm afraid I'm just going to make those changes myself, sorry!

I'll leave this open in case you want to convert it to, say, a Docker-support PR -- but honestly that may be better as a separate PR of its own, and just closing this one.

lurch:
> it's only Debian oldoldstable (buster) that's on 3.7 and I don't think there's any support for that on the RaspiOS side at this point.

Correct - Raspberry Pi Ltd only support Raspberry Pi OS (based on bookworm) and Raspberry Pi OS Legacy (based on bullseye), with the latter getting security-updates only. https://www.raspberrypi.com/software/operating-systems/  https://www.raspberrypi.com/news/new-old-functionality-with-raspberry-pi-os-legacy/

MusicalNinjaRandInt:
Hi @waveform80 , no worries.
On 3.12 compatibility: I did that a few weeks ago for my own personal reasons - and ended up killing the PR as everything for 3.12 is included in this approach to the deprecation of the Dict interface and in #1116

(It was @bennuttall specific request to be compatible for 3.7 that led to bringing in the importlib backports - you can drop those completely if 3.9 is the target)

I'm afraid I won't be able to do much until Monday - but I can offer a 3.9-3.12 compatibility PR with about 90 minutes investment (and easily lose the docker changes)

MusicalNinjaRandInt:
Ah - see you merged already today using the backport. I'll kill this PR

PR #1122, opened 2024-02-19
Support venv and use of github actions for contributors

### LINKED ISSUE #1109: First contribution hurdles in .gitignore, workflow, scripts
Operating system: Windows11 / WLS2 (Ubuntu 20.04)
Python version: 3.12
Pi model: N/A
GPIO Zero version: N/A
Pin factory used: MockFactory

As a first time contributor I ran into minor issues which required simple adjustments to make developing feasible.
I develop in a simple virtual environment (via `python -m venv .venv`) for each project leaving my system and user level python untouched, use vscode as an editor and like to have github actions run regression tests on a matrix of versions for each push on every branch

1. .gitignore doesn't have relevant entries for venv or vscode
2. .github/workflows/test.yml only triggers on `master`
3. the doc preview script uses the system-wide default python and not the virtual environment

While making these minor changes is not a big deal after forking, they need to be excluded from any pull requests - leading to more complicated rebasing before raising a PR and limited testing possible after rebasing. Without the .gitignore changes, git identified 4987 changed files in the branch I created for a PR - which makes it easy to make a mistake.

MusicalNinjaDad:
Resolves #1109

The current .gitignore, setup.py and test.yml are focussed on a specific development set up.
These changes support a few additional development setup preferences without breaking anything. (Do no harm):
- Using pythons venv virtual environments,
- using editors other than vim
- debugging (via) tests,
- leveraging github actions for regression testing,
- visualising code coverage results on codecov.io

Each of these adjustments do no harm, and are individually committed to allow for cherry-picking if there is something you do not like.

If these changes are not integrated into the main codebase then any PRs raised by someone who has made a similar adjustment to support their environment will either silently include them or need rebasing before raising and then be prone to error due to a large number of changes suddenly being present in working tree, regression testing not executing etc. on the final code included in the PR.

A note on codecov.io - the action will do nothing if the repo is not previously linked, so there is no data uploaded etc. unless the repo owner has specifically set this up in advance. I.e. no harm in the action being present at the end of the pipeline.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,7 +62,9 @@ gpiozero_mock_pin_classes =
     mocktriggerpin   = gpiozero.pins.mock:MockTriggerPin
 
 [tool:pytest]
-addopts = -rsx --cov --tb=short
+# addopts = -rsx --cov --tb=short
+# Uncomment the above line to provide basic coverage information in local runs
+# Comment out the above line to allow for debugging of tests (broken by --cov)
 testpaths = tests
 
 [coverage:run]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,11 +3,7 @@
 
 name: Python package
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
 
 jobs:
   test:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,3 +43,8 @@ jobs:
     - name: Run tests
       run: |
         pytest -v -k 'not test_real_pins.py' --cov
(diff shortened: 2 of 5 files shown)


setup.cfg
```diff
@@ -62,7 +62,9 @@ gpiozero_mock_pin_classes =
     mocktriggerpin   = gpiozero.pins.mock:MockTriggerPin
 
 [tool:pytest]
-addopts = -rsx --cov --tb=short
+# addopts = -rsx --cov --tb=short
+# Uncomment the above line to provide basic coverage information in local runs
```
lurch, line 66: I assume that this was enabled for a reason, so perhaps it should be left enabled by default? :shrug:

PR #1124, opened 2024-02-19
Fix class graph generation

### LINKED ISSUE #1123: Generating class diagrams for docs fails (silently) and breaks diagrams
Operating system: Bookworm, Bullseye
Python version: 3.9, 3.11 & 3.12
Pi model: N/A
GPIO Zero version: 2.0.1
Pin factory used: N/A

Regenerating the class diagrams for the docs leads to blank diagrams (dot, png & svg files are affected). This occurs via both `make doc` and `make preview`

Initial debugging suggests that the issue is (at least partly) due to the interaction of:
https://github.com/gpiozero/gpiozero/blob/6bfab8aa52c4098503589c3cd0bff7371a598ff4/scripts/class_graph#L121
and
https://github.com/gpiozero/gpiozero/blob/6bfab8aa52c4098503589c3cd0bff7371a598ff4/setup.cfg#L119

which equates to `for py_file in Path(PROJECT_ROOT)/Path('gpiozero/**').glob('*.py')`

I think that what is actually wanted is `for py_file in Path(PROJECT_ROOT)/Path('gpiozero').glob('**/*.py')`

MusicalNinjaDad:
Regenerating the class diagrams for the docs leads to blank diagrams (dot, png & svg files are affected). This occurs via both `make doc` and `make preview`

The issue appears to have been that the search path `gpiozero/**/*.py` was split into `gpiozero/**`, as the base search path and `*.py` as the glob pattern.

I have simply adjusted the approach to take `PROJECT_ROOT` as the base path and the entire `search_path` as the glob.

Additionally I have committed the regenerated images which appear, to my eye, to contain all the correct elements but have a new layout, probably due to improvements in the generation algorithm.

Feel free to cherry pick the last 2 commits from this set, if you don't want the extra changes as described in #1122 

This closes #1123


.gitignore
```diff
@@ -1,10 +1,15 @@
...
+.venv
 
 # Editor detritus
 *.vim
 *.swp
 tags
+# If you wish to contribute and use a different editor, please don't 
+# extend this list, or it will quickly get very long.
+# Instead create a '.gitignore-global' file in your home directory and 
+# run 'git config --global core.excludesfile ~/.gitignore-global'
```
lurch, line 12: Personally, I don't think that gpiozero has so many contributors that this would become a problem :wink:

MusicalNinjaRandInt: 😞 it's such an awesome project with a great purpose and really nice code.
There must be more dads out there who, like me, stumble across it with their kids and want to learn/hone skills and help out...

lurch: I agree that it's a great project (which, as I'm sure you've discovered by now is a lot more complex under-the-hood than it first appears), but IMHO telling people that they need to set up a `.gitignore-global` is a bit of a needless barrier to any potential contributors.

MusicalNinjaRandInt: @lurch - that's interesting feedback. Personally, I found the changes I needed to make to .gitignore etc. were the biggest barriers, as I felt very uncomfortable doing that for someone else's project. Each to their own, I guess.
How does this comment look. I've updated it here (because of this conversation) and also in #1122 where it belongs.

MusicalNinjaDad: @lurch  - I've found your opinions to be very valuable guidance.

and yes, life would be a lot easier for me too, if I were able to keep them separate. Doing that in this case gives me a few bad options: 1) manage 4992 untracked files in my repo and run no tests in my github pipeline (too error prone for my liking); 2) do some history-rewriting, rebasing magic just before creating the PR (again error prone); 3) wait a few months for the first PR to be reviewed and approved before doing anything else (not really very attractive either)

Again, thanks for your guidance and ideas. I'm going to move on for now and leave this PR open in case a fix for this bug is wanted at some point in the future ...

bennuttall:
Sorry I didn't see this at the time. I have just hit the issue myself and resolved it in much the same way. Thanks anyway!

PR #1134, opened 2024-03-26
really drop python<=3.7 support

kloczek:
Filter all code over `pyupgrade --py38-plus`.

### DIFF
--- a/tests/test_fonts.py
+++ b/tests/test_fonts.py
@@ -4,13 +4,7 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
-from __future__ import (
-    unicode_literals,
-    absolute_import,
-    print_function,
-    division,
-)
-str = type('')
+str = str
 
 import io
 
--- a/tests/test_fonts.py
+++ b/tests/test_fonts.py
@@ -4,8 +4,6 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
-str = str
-
 import io
 
 import pytest
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# vim: set fileencoding=utf-8:
 #
 # GPIO Zero: a library for controlling the Raspberry Pi's GPIO pins
 #
--- a/gpiozero/__init__.py
+++ b/gpiozero/__init__.py
@@ -1,4 +1,3 @@
-# vim: set fileencoding=utf-8:
 #
 # GPIO Zero: a library for controlling the Raspberry Pi's GPIO pins
 #
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1,4 +1,3 @@
-# vim: set fileencoding=utf-8:
 #
 # GPIO Zero: a library for controlling the Raspberry Pi's GPIO pins
 #
--- a/gpiozero/compat.py
+++ b/gpiozero/compat.py
@@ -1,4 +1,3 @@
-# vim: set fileencoding=utf-8:
 #
 # GPIO Zero: a library for controlling the Raspberry Pi's GPIO pins
 #
--- a/gpiozero/devices.py
+++ b/gpiozero/devices.py
@@ -1,4 +1,3 @@
-# vim: set fileencoding=utf-8:
 #
 # GPIO Zero: a library for controlling the Raspberry Pi's GPIO pins
 #
--- a/gpiozero/exc.py
+++ b/gpiozero/exc.py
@@ -1,4 +1,3 @@
-# vim: set fileencoding=utf-8:
 #
 # GPIO Zero: a library for controlling the Raspberry Pi's GPIO pins
 #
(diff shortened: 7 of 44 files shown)


lurch:
Are all the vim-comment removals necessary? Whilst I don't use Vim myself, I assume whoever added those headers (@bennuttall or @waveform80 ?) finds them useful; and removing those changes from this PR would make it a bit smaller.

kloczek:
> Are all the vim-comment removals necessary?

All python files since python 3.x are UTF-8 and vim already has hardcoded thar in its settings which was not the case with python 2.x.
Nevertheless that patch is 100% result of `pyupgrade`.

tests/test_fonts.py
```diff
@@ -4,13 +4,7 @@
...
 # SPDX-License-Identifier: BSD-3-Clause
 
-from __future__ import (
-    unicode_literals,
-    absolute_import,
-    print_function,
-    division,
-)
-str = type('')
+str = str
```
Siecje, line 7: Remove?

bennuttall:
## Review

### Overview
A mechanical cleanup pass (`pyupgrade --py38-plus`) across 44 files, +32/−85. Three categories of change:
1. Removes the `# vim: set fileencoding=utf-8:` modeline from ~40 files.
2. Replaces `io.open()` with builtin `open()`, and drops redundant `'r'` mode args (4 files: `fonts/__init__.py`, `internal_devices.py`, `pins/local.py`, `pins/native.py`).
3. Replaces `IOError` with `OSError` (they've been aliases since Python 3.3) across `pins/lgpio.py`, `pins/native.py`, `pins/pigpio.py`, `pins/style.py`, `gpiozerocli/pinout.py`, `gpiozerocli/pintest.py`.
4. Drops a dead `from __future__ import (unicode_literals, absolute_import, print_function, division)` block in two test files.

The base repo already declares `python_requires = >=3.9` in `setup.cfg`, so this PR isn't changing supported versions — it's just finishing the removal of Python 2-era source cruft that packaging already excludes.

### Correctness
No behavioral changes intended, and I don't see any — `io.open` ≡ `open`, `IOError` ≡ `OSError`, dropping an explicit `'r'` mode is a no-op, and removing dead `__future__` imports on Python 3.9+ is a no-op. This should be functionally safe.

### Issues found

1. **Unused `import io` left behind** — in `gpiozero/internal_devices.py`, `gpiozero/pins/local.py`, `gpiozero/pins/native.py`, and `gpiozero/fonts/__init__.py`, every `io.open()` call in the file was converted to `open()`, but `import io` at the top of each file was not removed, and there's no other use of the `io` module in any of them. `make lint` (pylint) would flag these as unused imports.

2. **Inconsistent coverage — one file missed entirely**: `tests/test_threads.py` still has `# vim: set fileencoding=utf-8:` at the top, even though the PR touched 43 other files to remove exactly this line.

3. **Inconsistent coverage — `IOError` left in test files**: `tests/test_devices.py`, `tests/test_internal_devices.py`, and `tests/test_pins_data.py` were opened by this PR (to strip the modeline) but still use `IOError` in mocks/`pytest.raises(IOError)`, while the production code they test was switched to `OSError`. This won't break tests (same class, same object identity), but it's an inconsistent half-migration within files the PR already touched.

4. **Leftover no-op statement**: in `tests/test_fonts.py` and `tests/test_tones.py`, `str = type('')` was rewritten to `str = str` rather than being deleted. It's now a pointless self-assignment of a builtin. Should just be removed along with the trailing blank line.

### Style / convention
No changelog entry added — likely fine for an internal cleanup PR, but worth confirming against the project's usual practice for this kind of change.

### Test coverage / CI
No test changes needed since behavior is unchanged. Worth confirming CI actually ran and passed on GitHub before merging (the fork's branch is also named `master`, which can confuse some tooling).

### Security
None — no security-relevant code paths touched.

### Recommendation
Safe to merge functionally, but worth a quick follow-up to: remove the now-unused `io` imports, delete the `str = str` lines, and pick up the missed `tests/test_threads.py` modeline / `IOError` occurrences for full consistency with the PR's stated goal.

---

bennuttall:
Correction to my earlier review: I flagged `tests/test_threads.py` as a file this PR "missed" removing the vim modeline from. That was wrong — this PR is quite a few commits behind current `master`, and `test_threads.py` didn't exist yet at this PR's base commit, so it's not something the PR omitted. Apologies for the noise.

Separately, I pushed a follow-up commit to this branch fixing something I *should* have caught the first time (I hadn't actually run the test suite): the `io.open` → `open()` conversion silently broke test isolation. `tests/test_internal_devices.py` and `tests/test_pins_data.py` do `mock.patch('io.open', ...)`, but since production code now calls the builtin `open()` directly, those mocks stopped intercepting — the affected tests were falling through to real filesystem reads (e.g. actual `/proc/cpuinfo`) instead of the fixture data. Retargeted the mocks to the module-qualified name (e.g. `gpiozero.internal_devices.open`) so they actually exercise the mocked content again. Also cleaned up the unused `import io` left in 4 files and the dead `str = str` line in 2 test files. Full suite passes (369 passed) after the fix.

---
Posted by Claude Sonnet 5 (Anthropic) via Claude Code.

PR #1147, opened 2024-05-28
Add speed/direction motor

pigrammer3:
See #300 and #330. Add support for a motor driven by three pins: one that's speed and the other two like the two forward and backward pins on the regular `Motor` class. I couldn't think of a better name than `SpeedDirectionMotor`.

### DIFF
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -1207,6 +1207,20 @@ class Motor(SourceMixin, CompositeDevice):
         PWM_WITH_DIGITAL_ENABLE = auto()
         DIGITAL_WITH_PWM_ENABLE = auto()
 
+        @property
+        def has_pwm(self):
+            """
+            Returns :data:`True` if the control type involves PWM.
+            """
+            return "PWM" in self.name
+
+        @property
+        def has_enable(self):
+            """
+            Returns :data:`True` if the control type involves an enable pin.
+            """
+            return "ENABLE" in self.name
+
     def __init__(self, forward, backward, *, enable=None, pwm=True,
                  enable_is_pwm=False, pin_factory=None):
         if pwm and enable_is_pwm:
@@ -1240,7 +1254,7 @@ class Motor(SourceMixin, CompositeDevice):
         (full speed backward) and 1 (full speed forward), with 0 representing
         stopped.
         """
-        if self.has_enable:
+        if self.control_type.has_enable:
             return (self.forward_device.value - self.backward_device.value) * self.enable_device.value
         return self.forward_device.value - self.backward_device.value
 
@@ -1262,22 +1276,6 @@ class Motor(SourceMixin, CompositeDevice):
             self.stop()
 
     @property
-    def has_enable(self):
-        """
-        Returns :data:`True` if the motor is controlled using an enable pin,
-        either digital or PWM, and :data:`False` otherwise.
-        """
-        return self.control_type in {self.ControlType.DIGITAL_WITH_PWM_ENABLE, self.ControlType.PWM_WITH_DIGITAL_ENABLE, self.ControlType.DIGITAL_WITH_DIGITAL_ENABLE}
-
-    @property
-    def has_pwm(self):
-        """
-        Returns :data:`True` if the motor is controlled using PWM, either for
-        direction or enable, and :data:`False` otherwise.
-        """
-        return self.control_type in {self.ControlType.PWM, self.ControlType.PWM_WITH_DIGITAL_ENABLE, self.ControlType.DIGITAL_WITH_PWM_ENABLE}
-
-    @property
     def is_active(self):
         """
         Returns :data:`True` if the motor is currently running and
@@ -1296,7 +1294,7 @@ class Motor(SourceMixin, CompositeDevice):
         """
         if not 0 <= speed <= 1:
             raise ValueError('forward speed must be between 0 and 1')
-        if not self.has_pwm:
+        if not self.control_type.has_pwm:
             if speed not in (0, 1):
                 raise ValueError(
                     'forward speed must be 0 or 1 with non-PWM Motors')
--- a/gpiozero/output_devices.py
+++ b/gpiozero/output_devices.py
@@ -1189,7 +1189,7 @@ class Motor(SourceMixin, CompositeDevice):
     :param bool enable_is_pwm:
         If :data:`True`, construct a :class:`PWMOutputDevice` instance for
         the enable pin, allowing variable speed control using the enable pin
-        and not the motor controller pins. Additionally, if this is
+        and not the forward and backward pins. Additionally, if this is
         specified, *enable* must not be :data:`None` and *pwm* must be
         :data:`False`. The default is :data:`False`.
 
@@ -1209,14 +1209,15 @@ class Motor(SourceMixin, CompositeDevice):
             ('forward_device', PinClass(forward, pin_factory=pin_factory)),
             ('backward_device', PinClass(backward, pin_factory=pin_factory)),
         ))
-        if enable is not None and not enable_is_pwm:
-            devices['enable_device'] = DigitalOutputDevice(
-                enable,
-                initial_value=True,
-                pin_factory=pin_factory
-            )
-        elif enable is not None:
-            devices['enable_device'] = PWMOutputDevice(enable, pin_factory=pin_factory)
+        if enable is not None:
+            if not enable_is_pwm:
+                devices['enable_device'] = DigitalOutputDevice(
+                    enable,
+                    initial_value=True,
+                    pin_factory=pin_factory
+                )
+            else:
+                devices['enable_device'] = PWMOutputDevice(enable, pin_factory=pin_factory)
         super().__init__(_order=devices.keys(), pin_factory=pin_factory, **devices)
 
     @property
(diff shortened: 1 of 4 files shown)


gpiozero/output_devices.py
```diff
@@ -1292,6 +1292,127 @@ def stop(self):
...
+
+        :param float speed:
+            The speed at which the motor should turn. Can be any value between
+            0 (stopped) and the default 1 (maximum speed).
+        """
+        if not 0 <= speed <= 1:
+            raise ValueError('forward speed must be between 0 and 1')
+        self.backward_device.off()
+        self.forward_device.on()
+        self.speed_device.value = speed
```
lurch, line 1385: As a safety-measure, I wonder if this should _first_ set `self.speed_device.value` to `0` before turning the `forward_device` on? I.e. if you do
```python3
my_speed_motor.backward(1)
time.sleep(1)
my_speed_motor.forward(0.2)
```
you probably **don't** want a momentary blip where the motor is running forwards at full speed?

lurch:
Hrmmm, looking at my comments in #366 (specifically [this one](https://github.com/gpiozero/gpiozero/issues/366#issuecomment-224642935) ), I wonder if this could be shoe-horned into the existing `Motor()` class by e.g. adding a `pwm_the_enable_pin` parameter (which defaults to `False`) to `Motor.__init__` ?
And then if you use `Motor(forward=2, backward=3, enable=4, pwm=True)` you'd get the current behaviour of `forward_device` & `backward_device` being `PWMOutDevice`s and `enable_device` being a `DigitalOutputDevice`; but if you instead did  `Motor(forward=2, backward=3, enable=4, pwm_the_enable_pin=True)` then you'd get `forward_device` & `backward_device` being `DigitalOutputDevice`s and `enable_device` being a `PWMOutputDevice` (the behaviour in this PR) ? :thinking:
Would probably make sense to make passing both `pwmn=True` and `pwm_the_enable_pin=True` be treated as an error. Passing `enable=None` and `pwm_the_enable_pin=True` should also be treated as an error.

And I'm too tired right now to think of a better name than `pwm_the_enable_pin` :laughing:

bennuttall:
I thought this felt familiar. Does it not work the same if you just use the enable pin on the `Motor` class?

lurch:
> > And I'm too tired right now to think of a better name than `pwm_the_enable_pin` 😆
> 
> Yeah, that makes sense. How about `pwm_enable`?

I did consider that, but thought that people might confuse `pwm=True` and `pwm_enable=True` :confused: 
I guess `enable_is_pwm` might work? :thinking: 

> > I thought this felt familiar. Does it not work the same if you just use the enable pin on the `Motor` class?
> 
> That wouldn't work...  (Unless I misunderstood...)

I think @bennuttall was suggesting that rather than having `enable` be PWM and `forward` & `backward` be Digital (which is what you're doing in this PR), have you tried having `enable` be Digital and `forward` and `backward` be PWM? (which is something the `Motor()` class already supports)
Or does the particular motor-driver chip you're using not support that mode of operation?

bennuttall:
Yes I'm pretty sure that's where we ended up last time

gpiozero/output_devices.py
```diff
@@ -1185,24 +1186,37 @@ class Motor(SourceMixin, CompositeDevice):
         :class:`DigitalOutputDevice` instances, allowing only direction
         control.
 
+    :param bool enable_is_pwm:
+        If :data:`True`, construct a :class:`PWMOutputDevice` instance for
+        the enable pin, allowing variable speed control using the enable pin
+        and not the motor controller pins. Additionally, if this is
```
lurch, line 1192: Perhaps "not the motor controller pins" should be "rather than the forward and backward pins" ? (since enable is also one of the motor controller pins? :thinking: )

gpiozero/output_devices.py
```diff
@@ -1185,24 +1186,37 @@ class Motor(SourceMixin, CompositeDevice):
...
         ))
-        if enable is not None:
+        if enable is not None and not enable_is_pwm:
             devices['enable_device'] = DigitalOutputDevice(
                 enable,
                 initial_value=True,
                 pin_factory=pin_factory
             )
+        elif enable is not None:
+            devices['enable_device'] = PWMOutputDevice(enable, pin_factory=pin_factory)
```
lurch, line 1219: Only a minor tweak, but IMHO this might look better without the duplicate `enable is not None` tests? i.e. something like:
```python3
        if enable is not None:
            if not enable_is_pwm:
                devices['enable_device'] = DigitalOutputDevice(
                    enable,
                    initial_value=True,
                    pin_factory=pin_factory
                )
            else:
                devices['enable_device'] = PWMOutputDevice(enable, pin_factory=pin_factory)
```

gpiozero/output_devices.py
```diff
@@ -1212,6 +1226,8 @@ def value(self):
         (full speed backward) and 1 (full speed forward), with 0 representing
         stopped.
         """
+        if hasattr(self, "enable_device"):
+            return (self.forward_device.value - self.backward_device.value) * self.enable_device.value
```
lurch, line 1230: Hah, clever :wink:

tests/test_outputs.py
```diff
@@ -975,6 +975,19 @@ def test_motor_pins_nonpwm(mock_factory):
         assert motor.backward_device.pin is b
         assert isinstance(motor.backward_device, DigitalOutputDevice)
 
+def test_motor_pins_enable_pwm(mock_factory, pwm):
+    s = mock_factory.pin(1)
+    f = mock_factory.pin(2)
+    b = mock_factory.pin(3)
+    with Motor(f, b, enable=s, pwm=False, enable_is_pwm=True) as motor:
```
lurch, line 982: Looking at the other Motor tests here, it seems that the Motor constructor is passed explicit numbers rather than pin objects? i.e. should this be
```python3
    with Motor(2, 3, enable=1, pwm=False, enable_is_pwm=True) as motor:
```

(and obviously the same for the other tests you've added too)

gpiozero/output_devices.py
```diff
@@ -1264,16 +1284,20 @@ def backward(self, speed=1):
...
             raise ValueError('backward speed must be between 0 and 1')
-        if isinstance(self.backward_device, DigitalOutputDevice):
+        if isinstance(self.backward_device, DigitalOutputDevice) and isinstance(self.enable_device, DigitalOutputDevice):
             if speed not in (0, 1):
                 raise ValueError(
                     'backward speed must be 0 or 1 with non-PWM Motors')
+        if isinstance(self.enable_device, PWMOutputDevice):
+            self.forward_device.off()
+            self.enable_device.value = speed
+            self.backward_device.off()
```
lurch, line 1299: Should this be `self.backward_device.on()` ?? (and if so, how come your unit-tests didn't pick this up?)

gpiozero/output_devices.py
```diff
@@ -1245,16 +1261,20 @@ def forward(self, speed=1):
...
             if speed not in (0, 1):
                 raise ValueError(
                     'forward speed must be 0 or 1 with non-PWM Motors')
+        if isinstance(self.enable_device, PWMOutputDevice):
+            self.backward_device.off()
+            self.enable_device.value = speed
+            self.forward_device.on()
+            return
         self.backward_device.off()
         self.forward_device.value = speed
```
lurch, line 1279: Again, minor nitpick but I wonder if
```python3
        if isinstance(self.enable_device, PWMOutputDevice):
            self.backward_device.off()
            self.enable_device.value = speed
            self.forward_device.on()
            return
        self.backward_device.off()
        self.forward_device.value = speed
```
would be any clearer if written as
```python3
        self.backward_device.off()
        if isinstance(self.enable_device, PWMOutputDevice):
            self.enable_device.value = speed
            self.forward_device.on()
        else:
            self.forward_device.value = speed
```
? :shrug:

lurch:
Should the `stop()` method also do
```python3
    if isinstance(self.enable_device, PWMOutputDevice):
        self.enable_device.value = 1
```
after turning off the forward and backward pins, in order to activate the 'Fast Stop' feature you mentioned?

Thanks for adding such comprehensive tests :+1:

gpiozero/output_devices.py
```diff
@@ -1289,12 +1291,14 @@ def backward(self, speed=1):
...
-        if isinstance(self.backward_device, DigitalOutputDevice) and isinstance(self.enable_device, DigitalOutputDevice):
+        if isinstance(self.backward_device, DigitalOutputDevice) and (
+            not hasattr(self, "enable_device") or isinstance(self.enable_device, DigitalOutputDevice)
+        ):
             if speed not in (0, 1):
                 raise ValueError(
                     'backward speed must be 0 or 1 with non-PWM Motors')
         self.forward_device.off()
-        if isinstance(self.enable_device, PWMOutputDevice):
+        if hasattr(self, "enable_device") and isinstance(self.enable_device, PWMOutputDevice):
```
lurch, line 1301: I wonder if it'd be worth having a `self.enable_is_pwm` boolean attribute and checking that, rather than all of these `hasattr(...) and isinstance(...)` tests? :man_shrugging:

lurch: Sounds good. But I wonder if `DIGITAL`, `PWM`, `DIGITAL_WITH_DIGITAL_ENABLE`, `PWM_WITH_DIGITAL_ENABLE`, `DIGITAL_WITH_PWM_ENABLE` is a slightly clearer set of names?

gpiozero/output_devices.py
```diff
@@ -1231,6 +1261,22 @@ def value(self, value):
         else:
             self.stop()
 
+    @property
+    def has_enable(self):
```
lurch, line 1265: I wonder if it might make sense to hoist these properties up into the `ControlType` class?
I've not used the `Enum` type in Python before, but after a quick bit of experimentation, the following _seems_ to work?

```python3
class ControlType(Enum):
    DIGITAL = auto()
    PWM = auto()
    DIGITAL_WITH_DIGITAL_ENABLE = auto()
    PWM_WITH_DIGITAL_ENABLE = auto()
    DIGITAL_WITH_PWM_ENABLE = auto()

    @property
    def has_pwm(self):
        return 'PWM' in self.name

    @property
    def has_enable(self):
        return 'ENABLE' in self.name
```

PR #1171, opened 2024-09-21
Sharing the gift of time saved.

mjm41:
My 8 year old wants me to help him build an RC boat. I had a Raspberry Pi 5 I just bought. There's a piBoat tutorial online and a few others. Nearly everything is built with older hardware and GPIO Zero has made all of that obsolete. So I have to do it from scratch. Then I find out that even the Pi5 is too new for some of the software like pigpiod. Please grant me the knowledge that we can share with others this quick bit of information so they can save some time. That will be great.


bennuttall:
I've updated the docs accordingly in the latest release

Issue #1166, opened 2024-08-21
lgpio pin factory is broken on RPi5 since kernel 6.6.45


HiassofT:
Operating system: e.g. RPiOS Bookworm
Python version: e.g. 3.11.2
Pi model: e.g. Pi 5
GPIO Zero version: 2.0.1
Pin factory used: lgpio

A recent change in the RPi kernel switched the gpiochip number on RPi5 to 0 (see https://github.com/raspberrypi/linux/pull/6144) but gpiozero still uses a hardcoded chip number of 4 on RPi5 (see https://github.com/gpiozero/gpiozero/blob/master/gpiozero/pins/lgpio.py#L66) and thus fails to find the GPIO pin

```
root@raspberrypi:~# python3
Python 3.11.2 (main, May  2 2024, 11:59:08) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gpiozero
>>> b = gpiozero.Button(4)
/usr/lib/python3/dist-packages/gpiozero/devices.py:295: PinFactoryFallback: Falling back from lgpio: 'can not open gpiochip'
  warnings.warn(
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/gpiozero/pins/pi.py", line 408, in pin
    pin = self.pins[info]
          ~~~~~~~~~^^^^^^
KeyError: PinInfo(number=7, name='GPIO4', names=frozenset({'J8:7', 'BCM4', 'GPIO4', 4, 'WPI7', '4', 'BOARD7'}), pull='', row=4, col=1, interfaces=frozenset({'', 'gpio', 'dpi', 'spi', 'uart', 'i2c'}))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/gpiozero/devices.py", line 103, in __call__
    self = super().__call__(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/gpiozero/input_devices.py", line 412, in __init__
    super().__init__(
  File "/usr/lib/python3/dist-packages/gpiozero/mixins.py", line 417, in __init__
    super().__init__(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/gpiozero/input_devices.py", line 162, in __init__
    super().__init__(
[...]

PR #1181, opened 2024-12-11
BUGFIX: Allow manual `chip` parameter in LGPIOFactory

Eliotdoesprogramming:
This came up for me when using gpiozero on a new RaspberryPi 5, using the lgpio library I was able to access the gpio pins on chip 0, but when using gpiozero I was unable to access the pins. Looking further, I found that the LGPIO factory was always overwriting the `chip` variable. This is a simple fix for allowing the user to choose a gpiochip

Current Implementation: 

```python
    def __init__(self, chip=None):
        super().__init__()
        chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 else 0
        self._handle = lgpio.gpiochip_open(chip)
        self._chip = chip
        self.pin_class = LGPIOPin
```

Proposed Implimentation:
```python
    def __init__(self, chip=None):
        super().__init__()
        if chip is None:
            chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 else 0
        self._handle = lgpio.gpiochip_open(chip)
        self._chip = chip
        self.pin_class = LGPIOPin
```

in the current implimentation, the `chip` parameter will always be overwritten on line 66. I propose a simple fix by adding a `if not chip` check which will allow users to manually determine their chip. If chip is left blank, it will fall back to the default implimentation of `chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 else 0`

### DIFF
--- a/gpiozero/pins/lgpio.py
+++ b/gpiozero/pins/lgpio.py
@@ -64,7 +64,7 @@ class LGPIOFactory(LocalPiFactory):
     def __init__(self, chip=None):
         super().__init__()
         if chip is None:
-            chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 else 0
+            chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 and os.path.exists('/dev/gpiochip4') else 0
         self._handle = lgpio.gpiochip_open(chip)
         self._chip = chip
         self.pin_class = LGPIOPin
--- a/gpiozero/pins/lgpio.py
+++ b/gpiozero/pins/lgpio.py
@@ -63,7 +63,8 @@ class LGPIOFactory(LocalPiFactory):
     """
     def __init__(self, chip=None):
         super().__init__()
-        chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 else 0
+        if chip is None:
+            chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 else 0
         self._handle = lgpio.gpiochip_open(chip)
         self._chip = chip
         self.pin_class = LGPIOPin


lurch (approved):
LGTM

lurch:
See also #1166

Eliotdoesprogramming:
Happy to add extra  behavior into the PR, and I should have "allow edits by maintainers" enabled on this one.

Would we want validation that the path exists for *any* gpiochip?

e.x.

```python
def __init__(self, chip=None):
    super().__init__()
    if chip is None:
        chip = 4 if (self._get_revision() & 0xff0) >> 4 == 0x17 else 0
    if not os.path.exists(f"/dev/gpiochip{chip}"):
        raise OSError(f"path /dev/gpiochip{chip}  does not exist")
    self._handle = lgpio.gpiochip_open(chip)
    self._chip = chip
    self.pin_class = LGPIOPin
```

Eliotdoesprogramming:
Happy to add this behavior into the patch following #1166

pelwell:
Thanks, that's great. Now it just needs somebody with the right privileges (i.e. not me) to hit merge.

bennuttall:
Thanks both

@waveform80 are you ok with this?

Eliotdoesprogramming:
any update here? @bennuttall @waveform80

HJLK:
On June 15, 2026 this patch still is not present in this current lib, **why**? I spent two weeks of learning in the crypts of GPIO management on Pi5 with Arch Linux, and finally I can blink a LED...

gplasky:
I've also proposed #1216 to fix this and the other related bug at once (as well as attempting to future-proof the detection) but I don't feel strongly which of these gets merged -- it just needs to happen. Are the maintainers of this repo MIA?

PR #1185, opened 2025-01-19
Add gpiod pin factory, again

numericOverflow:
Adding gpiod as a pin factory because it's project has active support.  A number of other pin factories have minimal, if any support at this point.  This lack of base support is causing issues on new systems/kernels/platforms.  This results in headaches for those users who depend on updated devices.

Note:  I am re-pushing code written by @warthog618 since they have declined to proceed with PR #1165

I still think it's a good add, so I've merged the proposed changes into a new branch and re-opened a PR so gpiozero can review and merge.

Again, full credit goes to warthog618 for the contribution, I'm just picking up the torch to get it merged since they changed their mind.  I provided clear attribution in every commit comment to make sure it does not look as if I am trying to take undue credit for this code.

### DIFF
--- a/docs/api_pins.rst
+++ b/docs/api_pins.rst
@@ -134,6 +134,8 @@ they are tried by default.
 +=========+===============================================+===========================================+
 | lgpio   | :class:`gpiozero.pins.lgpio.LGPIOFactory`     | :class:`gpiozero.pins.lgpio.LGPIOPin`     |
 +---------+-----------------------------------------------+-------------------------------------------+
+| gpiod   | :class:`gpiozero.pins.gpiod.GpiodFactory`     | :class:`gpiozero.pins.gpiod.GpiodPin`     |
++---------+-----------------------------------------------+-------------------------------------------+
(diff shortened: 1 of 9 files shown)


chronos1970:
With the deprecation of sysfs, pigpio interrupts no longer work (see [604](https://github.com/joan2937/pigpio/issues/604 ) and [542](https://github.com/joan2937/pigpio/issues/542)). And, as far as I can tell, lgpio spawns new polling threads, for each pin that needs to be monitored. This not only wastes performance, but also induces timing delays, rendering the PI useless for many time sensitive applications. This PR makes interrupt based monitoring possible again.

@bennuttall please take a look at this when you have time. Thank you @warthog618 for the original PR and for the [detailed write up](https://github.com/gpiozero/gpiozero/pull/1165#issuecomment-2381074190).

numericOverflow:
Also giving shout out to @waveform80 to see if he has time to review.
It'd be a great improvement to the dire state of gpio support in recent OS releases.

numericOverflow:
@bennuttall, @lurch  - Any chance this gets merged at some point?

PR #1211, opened 2025-12-04
Add ModMyPi Programmable Christmas Star class

mtouzot:
I've added the ModMyPiStar to control this [star](https://thepihut.com/products/raspberry-pi-christmas-tree-star). This class is based on their [github repo](https://github.com/modmypi/Programmable-Christmas-Star).

Also, I've added the doc and build the HTML part without issues.

If you see something is wrong or something is missing, don't hesitate

### DIFF
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1324,6 +1324,101 @@ class PiHutXmasTree(LEDBoard):
         )
 
 
+class ModMyPiStar(LEDBoard):
+    """
+    Extends :class:`LEDBoard` for the `ModMyPi Programmable Christmas Tree Star`_:
+    a Christmas tree star with 30 white LEDs.
+
+    The 30 white LEDs can be accessed through the attributes inner and outer.
+    The inner attribute controls the five inner LEDs, while the outer attribute
+    controls the 25 outer LEDs. Alternatively, as with all descendents of
+    :class:`LEDBoard`, you can treat the instance as a sequence of LEDs (the
+    first element is the inner LEDs).
+
+    The ModMyPi star board pins are fixed and therefore there's no need to
+    specify them when constructing this class. The following example turns on
+    all the inner LEDs before turning on the outer LEDs one at a time::
+
+        from gpiozero import ModMyPiStar
+        from time import sleep
+
+        tree = ModMyPiStar()
+
+        for light in tree:
+            light.on()
+            sleep(1)
+
+    The following example turns on the outer LEDs successively and increasingly
+    quickly before flashing the inner LEDs::
+
+        from time import sleep
+        from star import ModMyPiStar
+
+        step = 0.5
+        count = 0
+        star = ModMyPiStar(pwm=True)
+        leds = star.leds
+
+        while True:
+            if(count%26!=0):
+                leds[count%26].on()
+                sleep(step)
+                leds[count%26].off()
+
+            count += 1
+            step = step*0.99
+
+            if(step <= 0.0001):
+                star.inner.blink(on_time=0.5,off_time=0.5,n=5)
+                sleep(5)
+                count = 0
+                step = 2
+
+    :param bool pwm:
+        If :data:`True`, construct :class:`PWMLED` instances for each pin. If
+        :data:`False` (the default), construct regular :class:`LED` instances.
+
+    :type initial_value: bool or None
+    :param initial_value:
+        If :data:`False` (the default), all LEDs will be off initially. If
+        :data:`None`, each device will be left in whatever state the pin is
+        found in when configured for output (warning: this can be on). If
+        :data:`True`, the device will be switched on initially.
+
+    :type pin_factory: Factory or None
+    :param pin_factory:
+        See :doc:`api_pins` for more information (this is an advanced feature
+        which most users can ignore).
+
+    .. _ModMyPi Programmable Christmas Tree Star: https://thepihut.com/products/raspberry-pi-christmas-tree-star
+
+    .. attribute:: inner
+
+        Returns the :class:`LED` or :class:`PWMLED` representing the five inner
+        white LEDs in the center of the star.
+
+    .. attribute:: outer
+
+        Returns the :class:`LED` or :class:`PWMLED` representing the twenty-five
+        outer white LEDs.
+
+    """
+    def __init__(self, pwm=False, initial_value=False, pin_factory=None):
+        super(ModMyPiStar, self).__init__(
+            outer=LEDBoard(
+                A=8,B=7,C=12,D=21,E=20,F=16,G=26,H=19,I=13,J=6,K=5,L=11,M=9,
+                N=10,O=22,P=27,Q=17,R=4,S=3,T=14,U=23,V=18,W=15,X=24,Y=25,
+                pwm=pwm, initial_value=initial_value,
+                _order=('A','B','C','D','E','F','G','H','I','J','K','L',
+                        'M','N','O','P','Q','R','S','T','U','V','W','X','Y'),
+                pin_factory=pin_factory),
+            inner=2,
+            pwm=pwm, initial_value=initial_value,
+            _order=('inner','outer'),
+            pin_factory=pin_factory
+            )
+
+
 class LedBorg(RGBLED):
     """
     Extends :class:`RGBLED` for the `PiBorg LedBorg`_: an add-on board
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1336,23 +1336,24 @@ class ModMyPiStar(LEDBoard):
     first element is the inner LEDs).
 
     The ModMyPi star board pins are fixed and therefore there's no need to
-    specify them when constructing this class. The following example turns on
-    all the inner LEDs before turning on the outer LEDs one at a time::
+    specify them when constructing this class.
+
+    The following example turns on all the inner and outer LEDs::
 
         from gpiozero import ModMyPiStar
         from time import sleep
 
-        tree = ModMyPiStar()
+        star = ModMyPiStar()
 
-        for light in tree:
+        for light in star:
             light.on()
             sleep(1)
 
     The following example turns on the outer LEDs successively and increasingly
     quickly before flashing the inner LEDs::
 
+        from gpiozero import ModMyPiStar
         from time import sleep
-        from star import ModMyPiStar
 
         step = 0.5
         count = 0
@@ -1366,10 +1367,10 @@ class ModMyPiStar(LEDBoard):
                 leds[count%26].off()
 
             count += 1
-            step = step*0.99
+            step *= 0.99
 
-            if(step <= 0.0001):
-                star.inner.blink(on_time=0.5,off_time=0.5,n=5)
+            if step <= 0.0001:
+                star.inner.blink(on_time=0.5, off_time=0.5, n=5)
                 sleep(5)
                 count = 0
                 step = 2
--- a/gpiozero/boards.py
+++ b/gpiozero/boards.py
@@ -1330,10 +1330,12 @@ class ModMyPiStar(LEDBoard):
     a Christmas tree star with 30 white LEDs.
 
     The 30 white LEDs can be accessed through the attributes inner and outer.
-    The inner attribute controls the five inner LEDs, while the outer attribute
-    controls the 25 outer LEDs. Alternatively, as with all descendents of
-    :class:`LEDBoard`, you can treat the instance as a sequence of LEDs (the
-    first element is the inner LEDs).
+    The inner attribute controls the five inner LEDs as a single group, while
+    the outer attribute controls the 25 outer LEDs. Those 25 LEDs can be controlled
+    individually.
+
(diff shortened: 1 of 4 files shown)


mtouzot:
Also, you can see the results of Github Actions here : https://github.com/mtouzot/gpiozero/actions/runs/19955502884

gpiozero/boards.py
```diff
@@ -1324,6 +1324,101 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
 
 
+class ModMyPiStar(LEDBoard):
+    """
+    Extends :class:`LEDBoard` for the `ModMyPi Programmable Christmas Tree Star`_:
+    a Christmas tree star with 30 white LEDs.
+
+    The 30 white LEDs can be accessed through the attributes inner and outer.
+    The inner attribute controls the five inner LEDs, while the outer attribute
+    controls the 25 outer LEDs. Alternatively, as with all descendents of
```
lurch: Based on your code, it sounds like all 5 inner LEDs can only be turned on/off together, but the outer 25 LEDs can be turned on or off individually? If so, that's probably worth clarifying here, as currently it sounds like all 30 LEDs are individually controllable.

gpiozero/boards.py
```diff
@@ -1324,6 +1324,101 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+        leds = star.leds
+
+        while True:
+            if(count%26!=0):
+                leds[count%26].on()
+                sleep(step)
+                leds[count%26].off()
+
+            count += 1
+            step = step*0.99
```
lurch: You could also do `step *= 0.99` :wink:

gpiozero/boards.py
```diff
@@ -1324,6 +1324,101 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+
+            if(step <= 0.0001):
```
lurch:
```suggestion
            if step <= 0.0001:
```
Brackets aren't needed for simple `if` expressions.

gpiozero/boards.py
```diff
@@ -1324,6 +1324,101 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+                star.inner.blink(on_time=0.5,off_time=0.5,n=5)
```
lurch:
```suggestion
                star.inner.blink(on_time=0.5, off_time=0.5, n=5)
```
Readability tweak.

gpiozero/boards.py
```diff
@@ -1324,6 +1324,101 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+
+        for light in tree:
+            light.on()
+            sleep(1)
+
+    The following example turns on the outer LEDs successively and increasingly
+    quickly before flashing the inner LEDs::
+
+        from time import sleep
+        from star import ModMyPiStar
```
lurch: Hmmm, in the earlier example you used `from gpiozero import ModMyPiStar` ?

gpiozero/boards.py
mtouzot: Thanks, I've fixed it

gpiozero/boards.py
mtouzot: I made a copy / paste mistake

lurch:
I don't think the description has been fixed? It still says:
"The 30 white LEDs can be accessed through the attributes inner and outer.
    The inner attribute controls the five inner LEDs, while the outer attribute
    controls the 25 outer LEDs."
which IMHO is quite ambiguous.

gpiozero/boards.py
mtouzot: Sorry, if it's still unclear. I've changed the doc. Let's hope my English phrasing is more explicit.

In fact, yes all the 5 inner LEDs are seen as one, and the 25 outer LEDs can be controled individually.

gpiozero/boards.py
```diff
@@ -1324,6 +1324,105 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
 
+class ModMyPiStar(LEDBoard):
+    """
+    Extends :class:`LEDBoard` for the `ModMyPi Programmable Christmas Tree Star`_:
+    a Christmas tree star with 30 white LEDs.
+
+    The 30 white LEDs can be accessed through the attributes inner and outer.
+    The inner attribute controls the five inner LEDs as a single group, while
+    the outer attribute controls the 25 outer LEDs. Those 25 LEDs can be controlled
+    individually.
```
lurch: Thanks, this looks much better now! Only a small tweak, but perhaps "Those 25 LEDs can be controlled individually." would read better as "The 25 outer LEDs can also be controlled individually." ?

gpiozero/boards.py
```diff
@@ -1324,6 +1324,105 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+
+    Alternatively, as with all descendents of :class:`LEDBoard`, you can treat
+    the instance as a sequence of LEDs (the first element is the inner LEDs).
```
lurch:
```suggestion
    the instance as a sequence of LEDs (the first element is the 5 inner LEDs).
```

gpiozero/boards.py
```diff
@@ -1324,6 +1324,105 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+    The following example turns on all the inner and outer LEDs::
+
+        from gpiozero import ModMyPiStar
+        from time import sleep
+
+        star = ModMyPiStar()
+
+        for light in star:
+            light.on()
+            sleep(1)
```
lurch: Might be nice to expand this example to:
```python
        while True:
            for light in star:
                light.on()
                sleep(1)
            for light in star:
                light.off()
                sleep(1)
```
?

gpiozero/boards.py
```diff
@@ -1324,6 +1324,105 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+    The following example turns on the 25 outer LEDs individually in sequence,
+    at an increasingly rapid pace, before flashing the five inner LEDs, which
+    act as a single group::
+
+        from gpiozero import ModMyPiStar
+        from time import sleep
+
+        step = 0.5
+        count = 0
+        star = ModMyPiStar(pwm=True)
```
lurch: No need for `pwm=True` here as you're only turning the LEDs on and off?

gpiozero/boards.py
```diff
@@ -1324,6 +1324,105 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+        leds = star.leds
```
lurch: IMHO this might make the code slightly harder to read for some people?

gpiozero/boards.py
```diff
@@ -1324,6 +1324,105 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+
+        while True:
+            if(count%26!=0):
+                leds[count%26].on()
+                sleep(step)
+                leds[count%26].off()
```
lurch: I guess this could be "simplified" to:
```suggestion
            which_led = count % 26
            if which_led != 0:
                star.leds[which_led].on()
                sleep(step)
                star.leds[which_led].off()
```

Or you could even do:
```suggestion
            which_led = count % 25
            star.outer[which_led].on()
            sleep(step)
            star.outer[which_led].off()
```
? (note that I don't have a ModMyPiStar, so this is entirely untested!)

gpiozero/boards.py
```diff
@@ -1324,6 +1324,105 @@ def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):
...
+
+    .. attribute:: inner
+
+        Returns the :class:`LED` or :class:`PWMLED` representing the five inner
+        white LEDs in the center of the star.
+
+    .. attribute:: outer
+
+        Returns the :class:`LED` or :class:`PWMLED` representing the twenty-five
+        outer white LEDs.
```
lurch:
```suggestion
        Returns the :class:`LEDBoard` representing the twenty-five outer white
        LEDs.
```
?

gpiozero/boards.py
mtouzot: Ok, so with your example:
 * all 5 inners LEDs will turn on, 
 * sleep a second, 
 * all 25 outer LEDs will turn on,
 * sleep a second
 * all 5 inners LEDs will turn off
 * sleep a second
 * all 25 outer LEDs will turn off
 
 I can change the example and the description

lurch: I just thought it might be nicer if the demo was continually "doing something", rather than getting to an end-state and then stopping? It's entirely up to you though. :shrug: 
But you're right that I hadn't spotted that you were iterating over `star` and not `star.leds`, so perhaps a simpler and more illustrative example would be:
```python3
        while True:
            for group in star:
                group.on()
                sleep(1)
            star.off()
            sleep(1)
```

lurch:
This LGTM now, thank you for acting on all my suggestions! :+1: 
So I'll now leave it for the project maintainers to merge, when they get a chance.

mtouzot:
Thank you too for your reviews and for challenging me to improve the doc !

bennuttall:
## Review

### Overview
Adds `ModMyPiStar`, a new `LEDBoard` subclass for the ModMyPi Programmable Christmas Star (30 white LEDs: 5 inner as one group, 25 outer individually addressable), modeled on the vendor's own reference repo. Small, self-contained PR (+108/−0, 3 files: new class, `__init__.py` export, docs page). I checked it out and ran it live against `MockFactory` — construction, attribute access, `.on()`, and the `.value` namedtuple all work exactly as documented.

### Issues found

**1. Real bug, not just style: constructor isn't keyword-only, and fails in a confusing way when misused**
Every comparable class in this file (`PiHutXmasTree`, `SnowPi`, etc.) declares `def __init__(self, *, pwm=False, initial_value=False, pin_factory=None):` — keyword-only, per the project's own documented 2.x convention (`docs/compat.rst`, "Keyword arguments" section: *"In GPIO Zero 2.0, all such arguments are now actually keyword arguments"*). `ModMyPiStar.__init__` is missing the `*,`. I confirmed the consequence isn't just theoretical:
```
>>> PiHutXmasTree(True)
TypeError: PiHutXmasTree.__init__() takes 1 positional argument but 2 were given
>>> ModMyPiStar(True)
... 8 frames deep ...
gpiozero.exc.PinPWMUnsupported
```
`PiHutXmasTree(True)` fails immediately with a clear message. `ModMyPiStar(True)` silently binds `True` to `pwm` positionally, then blows up much later with an unrelated `PinPWMUnsupported` from the mock pin — a genuinely confusing failure mode for anyone who makes this mistake. Easy fix: add `*,` after `self`.

**2. Missing test coverage**
Every sibling class (`PiHutXmasTree`, `SnowPi`) has dedicated tests in `tests/test_boards.py` — construction, `repr()`, `pwm`/`initial_value` variants. `ModMyPiStar` has none. Given issue #1 above, a basic construction/attribute test would have caught it.

**3. Duplicated pin list (DRY / transcription-risk)**
`PiHutXmasTree` builds its pin mapping programmatically:
```python
pins_dict = OrderedDict(star=2)
pins = (4, 15, 13, ...)
for i, pin in enumerate(pins):
    pins_dict[f'led{i + 1:d}'] = pin
...
_order=pins_dict.keys(),
**pins_dict
```
one source of truth. `ModMyPiStar` instead hand-writes all 25 pin assignments (`A=8,B=7,C=12,...`) *and* separately repeats all 25 names again in `_order=('A','B','C',...)`. With 25 single-letter names, a transposition typo in either list would silently create a mismatch between a device and its position in `.value`, and nothing here would catch it. Worth following the existing loop-based pattern.

**4. Attribute naming inconsistent with sibling classes**
The 25 outer LEDs are named `A` through `Y`. Every other multi-LED board in this file uses descriptive names (`PiHutXmasTree`: `led0`, `led1`, ...; `SnowPi`: `top`/`middle`/`bottom`, `left`/`right`). Single uppercase letters are non-descriptive and a bit visually odd next to Python class names. `led0`..`led24` (matching `PiHutXmasTree` exactly) would be more consistent with the rest of the library's public API.

### Minor / style
- `super(ModMyPiStar, self).__init__(...)` — old-style explicit-args `super()`; every other class in this file (including `PiHutXmasTree` right above it) uses bare `super().__init__(...)`. This codebase has been actively cleaning up exactly this kind of Python-2-era pattern elsewhere.
- `A=8,B=7,C=12,D=21,...` — no spaces after commas; inconsistent with the rest of the file's formatting and would likely trip a linter (PEP8 E231).

### Test coverage / CI
None added, as noted above.

### Security
None relevant — static pin numbers, no external input.

### Recommendation
The board definition itself is correct and works (verified live), and it's exactly the kind of small, well-scoped board-support PR this project accepts often. But #1 is a real, demonstrated bug (not a nitpick) that should block merge as-is — worth adding the `*,` and, ideally, a quick construction test before merging. #3/#4 are more about matching established project conventions than correctness, but given this file already has a clean reference pattern one screen above, it'd be a shame not to follow it.

---

bennuttall:
@lurch does this look ok to you?

PR #1218, opened 2026-04-29
Add HumidityTemperatureSensor class for DHT22/DHT11 sensors

### LINKED ISSUE #254: DHT
We should look at adding support for DHTs like the DHT11 and DHT22.

There's an Adafruit library which is pretty poor, and requires sudo. There was talk of a kernel module providing a file interface, but I'm not sure where that went. Perhaps, as it's one-wire, this could work similarly to the Temperature Sensor (#93).

PatrickSchaffner:
## Summary

- Adds `HumidityTemperatureSensor` to `input_devices.py`, supporting DHT22 and DHT11 sensors over a single-wire protocol
- Exposes `.humidity` and `.temperature` properties alongside the standard `ValuesMixin` `.value` tuple
- Adds `MockHumidityTemperatureSensor` to the mock pin factory so the class can be tested without hardware
- Includes docs (`api_input.rst`, `api_exc.rst`, usage example) and 11 new test functions

Closes #254

## Test plan

- [ ] `pytest tests/test_inputs.py -k HumidityTemperature` — all new unit tests pass with `MockFactory`
- [ ] Verify `humidity` and `temperature` properties return expected values from mock sensor
- [ ] Check docs build: `make doc` produces no warnings for new entries
- [ ] (Optional hardware) Run `docs/examples/humidity_temperature_sensor.py` against a real DHT22 on a Pi

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/api_exc.rst
+++ b/docs/api_exc.rst
@@ -196,6 +196,12 @@ Warnings
 .. autoexception:: DistanceSensorNoEcho
     :show-inheritance:
 
+.. autoexception:: HumidityTemperatureSensorNoResponse
+    :show-inheritance:
+
+.. autoexception:: HumidityTemperatureSensorBadChecksum
+    :show-inheritance:
+
 .. autoexception:: SPIWarning
     :show-inheritance:
 
--- a/docs/api_input.rst
+++ b/docs/api_input.rst
@@ -70,6 +70,13 @@ RotaryEncoder
     :members: wait_for_rotate, wait_for_rotate_clockwise, wait_for_rotate_counter_clockwise, when_rotated, when_rotated_clockwise, when_rotated_counter_clockwise, steps, value, max_steps, threshold_steps, wrap
 
 
+HumidityTemperatureSensor
+-------------------------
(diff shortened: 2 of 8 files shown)


PatrickSchaffner:
Closing this PR — withdrawing it after tests on the lgpio backend.

`HumidityTemperatureSensor` decodes the DHT protocol by timestamping GPIO edge callbacks: the sensor packs 40 bits into pulses as short as ~26 µs across a ~4 ms frame. That works under `PiGPIOFactory` — the pigpio daemon samples every GPIO in a dedicated tight loop and hands back reliably-timestamped edges. It does **not** work under `LGPIOFactory`: there, edge events are delivered to a userspace callback that can't keep pace with a DHT frame's bursts of edges, so edges get dropped and the frame can't be reconstructed.

As written, then, this device only works on a non-default backend — merging it would ship a sensor that silently fails for most users out of the box. The edge-callback approach just isn't the right foundation for DHT support in current gpiozero.

PR #1228, opened 2026-07-25
Make TimeOfDay timezone aware

bennuttall:
Continuation from #1116


bennuttall:
@lurch do you think this looks ok?

